PR #2, opened 2020-12-03
Fix UnsignedByteArray.of builder method

nhartner:
This method was not assembly the bytes in the correct order which was causing large blobs that
have multibyte length headers to fail.


sappenin (approved):
LGTM!

PR #21, opened 2020-12-08
Type LedgerHeader.transactions as a List of TransactionResults

### LINKED ISSUE #7: Fix issue with LedgerHeader transaction deserialization
`XrplClient.ledger(params)` method throws an exception if transactions are returned because the `LedgerHeader.transactions()` declaration tries to deserialize them as `String` instead of JSON.

The transactions that are returned can either be full JSON transaction objects, or the identifying hashes of the transactions.  We could add a custom deserializer, but this kind of polymorphism isn't easy to do in Java, so I think we should just be opinionated and force the request to set `expand=true` when asking for transactions

nkramer44:
Fixes #7 

- Adds a `TransactionDeserializer` to deserialize transaction JSON to the correct transaction type
- Adds a `TransactionResultDeserializer` to deserialize `TransactionResults` with the correct transaction type, which needs to be wrapped


xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/transactions/TransactionResultJsonTests.java
```diff
@@ -0,0 +1,32 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.junit.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.ImmutablePayment;
+
+public class TransactionResultJsonTests extends AbstractJsonTest {
+
+  @Test
+  public void deserializePaymentTransactionResult() throws JsonProcessingException {
```
sappenin, line 14: Can we add test coverage around serialize -> deserialize? Or is that perhaps covered elsewhere?

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/ledger/LedgerHeaderJsonTests.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.TransactionType;
+
+public class LedgerHeaderJsonTests extends AbstractJsonTest {
+
+  @Test
+  public void deserializeLedgerHeaderWithTransactions() throws JsonProcessingException {
```
sappenin, line 16: Can we add test coverage around serialize -> deserialize? Or is that perhaps covered elsewhere?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/transactions/TransactionResult.java
```diff
@@ -1,22 +1,23 @@
...
+
 /**
  * The result of a tx rippled API method call.
  *
  * @param <TxnType> The type of {@link Transaction} that was requested.
  */
 @Value.Immutable
-@JsonSerialize(as = ImmutableTransactionResult.class)
-@JsonDeserialize(as = ImmutableTransactionResult.class)
+@JsonDeserialize(using = TransactionResultDeserializer.class)
```
sappenin, line 20: Is this class not used for serialization?

sappenin (changes requested):

sappenin:
Build seems to be broken, FYI.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/transactions/TransactionResultJsonTests.java
nkramer44: Originally, we decided this class and `TransactionResult` didn't need to be serialized, since they are just response objects and we'd need a custom serializer to serialize correctly.  The only reason I can think of to serialize would be if you were writing your own rippled in java.

Now that I've slept on it though, I think i'll add that custom serializer, just so the response objects are consistent

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/Xrpl4jModule.java
```diff
@@ -38,5 +39,7 @@ public Xrpl4jModule() {
 
     addSerializer(LedgerIndex.class, new LedgerIndexSerializer());
     addDeserializer(LedgerIndex.class, new LedgerIndexDeserializer());
+
+    addDeserializer(Transaction.class, new TransactionDeserializer());
```
sappenin, line 43: I think we're missing `TransactionResultDeserializer`?

nkramer44: I have it in an annotation on `TransactionResult`. Tried adding it here, but it wasn't getting registered properly, then I checked the javadoc for `addDeserializer`
```
/**
     * Method for adding deserializer to handle specified type.
     *<p>
     * WARNING! Type matching only uses type-erased {@code Class} and should NOT
     * be used when registering serializers for generic types like
     * {@link java.util.Collection} and {@link java.util.Map}.
     */
```

Not sure if there is a good way to add a deserializer with a generic type, but using the annotation seems to work fine, so I think I'll just leave it

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerRequestParams.java
```diff
@@ -78,9 +78,9 @@ default boolean transactions() {
    * Provide full JSON-formatted information for transaction/account information instead of only hashes.
    * Defaults to false. Ignored unless you request {@link #transactions()}, {@link #accounts()}, or both.
    */
-  @Value.Default
+  @Value.Derived
```
sappenin, line 81: Minor nit but if these are derived they should probably be marked `@Overide`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerRequestParams.java
```diff
@@ -98,7 +98,7 @@ default boolean ownerFunds() {
    * If true, and {@link #transactions()} and {@link #expand()} are both also true, return transaction information
    * in binary format (hexadecimal string) instead of JSON format.
    */
-  @Value.Default
+  @Value.Derived
```
sappenin, line 101: Minor nit but if these are derived they should probably be marked `@Overrde`

sappenin (approved):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerRequestParams.java
nkramer44: They aren't overriding anything though?

sappenin: Oh, I'm confused then - these "Derived" values will only ever have a single value, but the function makes it look like these might have two values (true or false). The Javadoc also makes this seems like this field is a shared concept.

PR #41, opened 2021-01-04
Add XAddress wrapper type

nkramer44:
Adds an `XAddress` immutable wrapper, which wraps a `String`.  Also converts method signatures in `AddressCodec` and elsewhere to use `Address` and `XAddress` instead of just `String`s


xrpl4j-address-codec/src/main/java/org/xrpl/xrpl4j/codec/addresses/AddressCodec.java
```diff
@@ -319,10 +332,10 @@ private boolean isTestAddress(final byte[] decoded) {
   /**
    * Tests if the given X-Address is a valid X-Address.
    *
-   * @param xAddress A potentially valid X-Address.
+   * @param xAddress A potentially valid {@link XAddress}.
    * @return {@code true} if the given address is a valid X-Address, {@code false} if not.
    */
-  public boolean isValidXAddress(final String xAddress) {
+  public boolean isValidXAddress(final XAddress xAddress) {
```
sappenin, line 338: More of a question for a different feature, but would it make sense to combine `Address` and `XAddress` into a single class just called `Address`?

sappenin: Talked with Noah, this is all good.

sappenin (approved):

PR #42, opened 2021-01-05
add codecov fixes

nkramer44:


codecov.yml
```diff
@@ -0,0 +1,2 @@
+fixes:
+  - "/home/runner/work/xrpl4j/xrpl4j/::"
```
sappenin, line 2: It's been awhile since I setup codecov, but in Quilt we did this instead: https://github.com/hyperledger/quilt/blob/master/.circleci/config.yml#L96

Does what you have here actually work though? The "2 colons" look odd, but what do i know?

sappenin (approved):

codecov.yml
nkramer44: Yeah there's a codecov specific github action that I'm using.  It gets configured here: https://github.com/XRPLF/xrpl4j/blob/4410c94e10085118512803bf744fb97fb104dbc7/.github/workflows/workflow.yml#L25.  I saw this fix in some tutorial somewhere. For some reason, without it codecov can't attribute coverage to specific files.

PR #47, opened 2021-01-22
Introduce BOM

sappenin:
A Bill of Materials (BOM) allows developers to control XRPL-4j versions from a single place.

### DIFF
--- a/xrpl4j-bom/README.md
+++ b/xrpl4j-bom/README.md
@@ -1,0 +1,33 @@
+# XRPL-4j Bill of Materials (BOM)
+A BOM is a special kind of Maven POM file that allows downstream projects using xrpl-4j to control dependency versions in one place.
+
+## Usage
+To use this bom in your project, add the following to your the `<dependencyManamgement/>` section of your project's primary POM file. For example:
+
+```xml
+<project ...>
+    
+    <groupId>...</groupId>
+    <artifactId>...</artifactId>
+    <version>...</version>
+    <packaging>...</packaging>
+
+    ...
+    
+    <dependencyManagement>
+        <dependencies>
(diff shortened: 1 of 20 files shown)


xrpl4j-bom/README.md
```diff
@@ -0,0 +1,33 @@
...
+                <version>1.0.1</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+</project>
+```
+
+With this in place, whenever you want add a `<dependency/>` you won't need to worr about specifying the version. Instead, version numbers are controlled by the BOM you import, as in the example above, which will use only version `1.0.1` of all XRPL-4j dependencies.
```
nkramer44, line 31:
```suggestion
With this in place, whenever you want add a `<dependency/>` you won't need to worry about specifying the version. Instead, version numbers are controlled by the BOM you import, as in the example above, which will use only version `1.0.1` of all XRPL-4j dependencies.
```

xrpl4j-bom/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+For more information on how BOM files work, consult this [tutorial](https://www.baeldung.com/spring-maven-bom) or other on Google.
```
nkramer44, line 33:
```suggestion
For more information on how BOM files work, consult this [tutorial](https://www.baeldung.com/spring-maven-bom) or others on Google.
```

nkramer44 (approved):
Just a couple typos, but this looks good

xrpl4j-bom/README.md
sappenin: This is fixed.

PR #54, opened 2021-02-10
Add account_tx support + fix marker ser/deser

### LINKED ISSUE #52: Re-type marker fields in request params and result objects
A lot of the request/response objects have a field called `marker`, which is a server defined field that can either be a `String` or a JSON object.  Currently, our objects are typed as `Optional<String>`, which will break Jackson when the server sends back a JSON object in that field.

We should create a `Marker` type with a custom serializer/deserializer to handle this.

nkramer44:
Adds support for the `account_tx` method to `XrplClient`, including the request/response objects.

Also fixes #52 .

Lastly, adds an `Optional<Hash256> hash` field to `Transaction` instead of including it in `TransactionResult` because `AccountTransactionsTransactionResult` needs the hash in the `"tx"` field value.


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -0,0 +1,116 @@
...
+  Optional<Hash256> ledgerHash();
+
+  /**
+   * Return transactions from the ledger with this index only.
+   *
+   * @return A {@link LedgerIndex} containing the ledger index, defaults to "current".
+   */
+  @JsonProperty("ledger_index")
+  @Value.Default
+  default LedgerIndex ledgerIndex() {
```
nhartner, line 74: This should be Optional according to the docs

nhartner (approved):
noice

PR #56, opened 2021-02-11
Introduce SignatureService

sappenin:
This PR introduces a SignatureService interface and other crypto prmitives to allow signing and verification to be used with key-storage mechanisms that do not hold the keys in memory (e.g., an HSM).


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -135,13 +141,49 @@ public XrplClient(HttpUrl rippledUrl) {
...
+  public <T extends Transaction> SubmitResult<T> submit(
+    final org.xrpl.xrpl4j.crypto.signing.SignedTransaction signedTransaction
+  ) throws JsonRpcClientErrorException, JsonProcessingException {
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("About to submit signedTransaction: {}", signedTransaction);
+    }
+
+    String signedJson = objectMapper.writeValueAsString(signedTransaction.signedTransaction());
+    String signedBlob = binaryCodec.encode(signedJson); // <-- txBlob must be binary-encoded.
```
nkramer44, line 165: Move this into `SignatureService` and add `signedBlob` to `SignedTransaction`. Also populate `Transaction.hash()` inside of `SignatureService` inside the `SignedTransaction`

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/SingleKeySignatureService.java
```diff
@@ -0,0 +1,182 @@
...
+  ) {
+    super(KEY_STORE_TYPE, signatureUtils);
+    this.ed25519Signer = Objects.requireNonNull(ed25519Signer);
+    this.ecdsaSigner = Objects.requireNonNull(ecdsaSigner);
+    this.privateKey = Objects.requireNonNull(privateKey);
+  }
+
+  @Override
+  public PublicKey getPublicKey(final KeyMetadata privateKeyMetadata) {
+    Objects.requireNonNull(privateKeyMetadata);
```
nkramer44, line 79: Get rid of this so we can call with null

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/SingleKeySignatureService.java
```diff
@@ -0,0 +1,182 @@
...
+    return BcKeyUtils.toPublicKey(this.privateKey);
+  }
+
+  @Override
+  protected synchronized Signature edDsaSign(
+    final KeyMetadata privateKeyMetadata, final UnsignedByteArray signableTransactionBytes
+  ) {
+    Objects.requireNonNull(privateKeyMetadata);
```
nkramer44, line 87: same here + everywhere

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/DerivedKeysSignatureService.java
```diff
@@ -0,0 +1,210 @@
...
+  public DerivedKeysSignatureService(
+    final ServerSecretSupplier serverSecretSupplier,
+    final VersionType versionType,
+    final KeyPairService keyPairService
+  ) {
+    this.serverSecretSupplier = Objects.requireNonNull(serverSecretSupplier);
+    this.versionType = Objects.requireNonNull(versionType);
+    this.keyPairService = Objects.requireNonNull(keyPairService);
+
+    this.keyMetadataLoadingCache = Caffeine.newBuilder()
```
nkramer44, line 75: Can we make this configurable?

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/test/java/org/xrpl/xrpl4j/crypto/signing/SoftHsmSignatureSigner.java
```diff
@@ -0,0 +1,326 @@
...
+ * consider something like PKCS#11 with JCA in order to support external signing operations). For now, in order to
+ * simulate keys that live on an external device, this implementation uses BouncyCastle keys for testing purposes. The
+ * main intent here is merely to validate the the signature interfaces work properly for both in-memory and externally
+ * held private keys, and this implementation simulates a sealed enclave where the private-key material is not
+ * accessible from the outside.</p>
+ *
+ * <p>This implementation has two keys, one generated using the secp256k1 curve (for ECDSA), and one generated using
+ * the Ed25519 curve (for EdDSA).</p>
+ */
+public class SoftHsmSignatureSigner implements SignatureService {
```
nkramer44, line 54:
```suggestion
public class SoftHsmSignatureService implements SignatureService {
```

nkramer44 (approved):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
sappenin: @nkramer44 This is moved into the `SigningService`, although I decided to maintain the original pattern that you introduced in `main` a few days ago because after I experimented with the design we discussed during PR review, I concluded that your original design was superior -- so, if a developer wants to get a "pre-handle" `hash` before submitting to the XRPL, the developer can use the `hash` value on `SignedTransaction`. This _should_ match the hash value on any transaction returned from XRPL via `account_tx` call.

Take a look here to see if you agree: https://github.com/XRPLF/xrpl4j/commit/2ee265f072626706d5920ac8dc52569b7c484c3b

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/DerivedKeysSignatureService.java
sappenin: Captured in https://github.com/XRPLF/xrpl4j/issues/58.

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/SingleKeySignatureService.java
sappenin: @nkramer44 I experimented a bit with this change as well. No surprise to you, but I consider `null` to be a bug, so I propose this change instead (https://github.com/XRPLF/xrpl4j/commit/0fec0f30ffe871374a027190dc7af98ddacf0c16). 

However, we could alternatively introduce default method in `TransactionSigner` and `TransactionVerifier` that only takes a `Transaction` and then delegates to the method that takes both metadata+transaction. Only reason I didn't go that route for now is that it seemed to complicate the interface for a use-case that maybe we want to prod developers not to use (i.e., a single-key, in-memory implementation). But open to other ideas here if you don't like my proposal above.

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/SingleKeySignatureService.java
sappenin: See above.

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/test/java/org/xrpl/xrpl4j/crypto/signing/SoftHsmSignatureSigner.java
sappenin: Fixed via https://github.com/XRPLF/xrpl4j/commit/59e9d59f15c35d0ebc82e1a07dd8b94346076fda.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
nkramer44: Yep this looks good to me

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/signing/SingleKeySignatureService.java
nkramer44: I think what you have is good for now.  We could always make a non-breaking change down the line to add that to `TransactionSigner` and `TransactionVerifier`, but it doesn't seem worth it now.

PR #65, opened 2021-02-19
make engineResult not Optional since it isn't

nhartner:
`SubmitResult#engineResult` has an Optional type which is not correct. Per the docs (https://xrpl.org/submit.html) it is not optional.


sappenin (approved):

sappenin:
Ship it!

nkramer44 (approved):

PR #68, opened 2021-02-23
Fixes Issue #49 (Make it easier to get the amount of an XRP payment as BigDecimal)

### LINKED ISSUE #49: Make it easier to get the amount of an XRP payment as BigDecimal
The way amounts are mapped could be more convenient to use. In most cases, people just want to get the amount in XRP as a decimal, not drops. But on things like payments, that requires doing an` instanceof` then a cast to `XrpCurrencyAmount` then no obvious way to get the decimal value (instead of drops) from there

sappenin:
* Make it easier to get the amout of an XRP Payment as a BigDecimal.
* Improve the boundary checking of the XrpCurrencyAmount.
* Add unit test coverage.

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
@@ -6,6 +6,8 @@ import org.immutables.value.Value;
 
 /**
  * A {@link CurrencyAmount} for Issued Currencies on the XRP Ledger.
+ *
+ * @see "https://xrpl.org/rippleapi-reference.html#value"
  */
 @Value.Immutable
 @JsonSerialize(as = ImmutableIssuedCurrencyAmount.class)
@@ -17,8 +19,11 @@ public interface IssuedCurrencyAmount extends CurrencyAmount {
   }
 
   /**
-   * Quoted decimal representation of the amount of currency. This can include scientific notation, such as
-   * 1.23e11 meaning 123,000,000,000. Both e and E may be used.
+   * Quoted decimal representation of the amount of currency. This can include scientific notation, such as 1.23e11
+   * meaning 123,000,000,000. Both e and E may be used. Note that while this implementation merely holds a {@link
+   * String} with no value restrictions, the XRP Ledger does not tolerate unlimited precision values. Instead, non-XRP
+   * values (i.e., values held in this object) can have up to 16 decimal digits of precision, with a maximum value of
+   * 9999999999999999e80. The smallest positive non-XRP value is 1e-81.
    *
    * @return A {@link String} containing the amount of this issued currency.
    */
@@ -32,8 +37,8 @@ public interface IssuedCurrencyAmount extends CurrencyAmount {
   String currency();
 
   /**
-   * Unique account {@link Address} of the entity issuing the currency. In other words, the person or business where
-   * the currency can be redeemed.
+   * Unique account {@link Address} of the entity issuing the currency. In other words, the person or business where the
+   * currency can be redeemed.
    *
    * @return The {@link Address} of the account of the issuer of this currency.
    */
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Payment.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Payment.java
@@ -1,13 +1,10 @@
 package org.xrpl.xrpl4j.model.transactions;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.primitives.UnsignedInteger;
 import org.immutables.value.Value;
-import org.immutables.value.Value.Derived;
-import org.immutables.value.Value.Lazy;
 import org.xrpl.xrpl4j.model.flags.Flags;
 import org.xrpl.xrpl4j.model.flags.Flags.PaymentFlags;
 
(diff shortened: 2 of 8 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Payment.java
```diff
@@ -41,14 +44,32 @@ default PaymentFlags flags() {
...
+   * A typed variant of {@link #amount()} which produces a value in XRP Drops if that result is an instance of {@link
+   * XrpCurrencyAmount}.
+   *
+   * @return The amount of this payment if this payment is denominated in XRP (i.e., if the result of calling {@link
+   *   #amount()} is an instance of {@link XrpCurrencyAmount}); otherwise, returns {@link Optional#empty()} (e.g., if
+   *   this payment is denominated in an Issued Currency).
+   */
+  @Derived
+  @JsonIgnore
+  default Optional<XrpCurrencyAmount> amountAsXrp() {
```
nkramer44, line 65: I liked your idea of providing a handler on `CurrencyAmount`.  The handler could work in tandem with an `amountAsXrp` method like this:
```java
payment.amount().handle(
  xrpCurrencyAmount -> xrpCurrencyAmount.toXrp(),
  issuedCurrencyAmount -> issuedCurrencyAmount
)
```

There are a few transactions that have an `amount` field that could be an `XrpCurrencyAmount` or an `IssuedCurrencyAmount`, so I would prefer having this pulled up into `CurrencyAmount` or `XrpCurrencyAmount`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
```diff
@@ -32,8 +37,8 @@
   String currency();
```
nkramer44, line 37: Maybe this is worth a separate issue, but maybe we could deprecate this in favor of a `BigDecimal` typed field.  I'm assuming there is a way to configure Jackson to deserialize scientific notation correctly, which would make this field much more useful.

nkramer44 (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -118,12 +128,23 @@ public static XrpCurrencyAmount ofDrops(UnsignedLong drops) {
...
-      return this.value().toString();
+    /**
+     * Convert the supplied amount of XRP drops into a decimal value representing a value denominated in whole XRP
+     * units.
+     *
+     * @param xrpCurrencyAmount A {@link XrpCurrencyAmount} representing an amount of XRP drops.
+     *
+     * @return A {@link BigDecimal} representing a value denominated in whole XRP units.
+     */
+    static BigDecimal toXrp(final XrpCurrencyAmount xrpCurrencyAmount) {
```
nhartner, line 145: This would be more obvious and convenient to use as a default method on XrpCurrency. So I could do something like `payment.amountAsXrp().toBigDecimal()` rather than `XrpCurrencyAmount.toXrp(payment.amountAsXrp())`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
sappenin: I like that in principle because using a number here seems more "correct" -- although the one downside of that is that a developer couldn't easily use a `String` here in their own code (they would have to translate to a BigDecimal). Is that a problem?

@nkramer44 I'm curious to hear more about how you use this field in your day-to-day development around issued currencies. Any thoughts there?

sappenin: In any event, let's create an issue for this and capture the conversation there?

sappenin: There's also this new [Discussions](https://github.com/XRPLF/xrpl4j/discussions) thing from GH -- might be an interesting first topic.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Payment.java
sappenin: Agreed. See latest commit here and let me know what you think?
https://github.com/XRPLF/xrpl4j/pull/68/commits/1dcd61f54b98b21a97053d05963a584efe7dc75d

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -118,12 +128,23 @@ public static XrpCurrencyAmount ofDrops(UnsignedLong drops) {
      * @return An {@link XrpCurrencyAmount} of the amount of drops in {@code amount}.
      */
     public static XrpCurrencyAmount ofXrp(BigDecimal amount) {
-      return ofDrops(amount.scaleByPowerOfTen(6).toBigIntegerExact().longValue());
+      if (FluentCompareTo.is(amount).notEqualTo(BigDecimal.ZERO)) {
```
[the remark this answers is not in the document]
sappenin: This is fully covered via [this test](https://github.com/XRPLF/xrpl4j/pull/68/files#diff-1bc50528c47cee0e21b5161efe5d957673e2eacfab4e4622e86d56bea8a1f479R41).

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
sappenin: Started a discussion thread here: https://github.com/XRPLF/xrpl4j/discussions/72

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: Agree, great suggestion! Changed [here](https://github.com/XRPLF/xrpl4j/pull/68/commits/195c7a87fd2bbb5c31e54353fb9ca56674c4415d).

nkramer44 (approved):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
nhartner: 👍

nhartner (approved):
bueno

PR #78, opened 2021-03-10
Return parameterized Transaction from SignatureService#sign

nhartner:
Accept parameterized Transaction in SignatureService#verify


nkramer44 (approved):
Noice, this will really clean up my warnings in Intellij

Issue #85, opened 2021-03-30
wrong type for load_factor in ServerInfo


tommq:
I started getting Exceptions from serverInfo() method when connected to s2.ripple.com. I realized the library expects load_factor  to be UnsignedInteger (and it usually is "1"), but rippled sometimes returns it as double.

It can be shown here: https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs1.ripple.com%2F&req=%7B%22id%22%3A1%2C%22command%22%3A%22server_info%22%7D

clicking send several times eventually get you response with some higher load like:

```
{
  "id": 1,
  "result": {
    "info": {
      "build_version": "1.7.0",
      "complete_ledgers": "61881385-62562429",
      "hostid": "LARD",
      "io_latency_ms": 2,
      "jq_trans_overflow": "0",
      "last_close": {
        "converge_time_s": 3.002,
        "proposers": 38
      },
      "load_factor": 511.83203125,
      "load_factor_server": 1,
      "peer_disconnects": "224008",
      "peer_disconnects_resources": "747",
      "peers": 261,
      "pubkey_node": "n9MozjnGB3tpULewtTsVtuudg5JqYFyV3QFdAtVLzJaxHcBaxuXD",
      "server_state": "full",
      "server_state_duration_us": "2274468435925",
      "state_accounting": {
        "connected": {
          "duration_us": "231197490",
          "transitions": 2
        },
        "disconnected": {
          "duration_us": "1156933",
          "transitions": 2
        },
        "full": {
          "duration_us": "2274468435925",
          "transitions": 1
        },
        "syncing": {
          "duration_us": "4065001",
          "transitions": 1
        },
        "tracking": {
          "duration_us": "63",
          "transitions": 1
        }
      },
      "time": "2021-Mar-30 15:37:51.486384 UTC",
      "uptime": 2274704,
      "validated_ledger": {
        "age": 4,
        "base_fee_xrp": 0.00001,
        "hash": "E5A958048D98D4EFEEDD2BC3F36D23893BBC1D9354CB3E739068D2DFDE3D1AA3",
        "reserve_base_xrp": 20,
        "reserve_inc_xrp": 5,
        "seq": 62562429
      },
      "validation_quorum": 31
    }
  },
  "status": "success",
  "type": "response"
}
``` 

The exception I am getting is clearly saying that jackson can't construct UnsignedInteger from 512.90234375
```
[...]

PR #87, opened 2021-04-06
Fix for types in ServerInfo to reflect rippled response

tommq:
Serveral 'load' properties now Double instead od UnsignedInteger

Peers, loadFactor, jqTransOverflow made Optional

I did not create a _Load wrapper as @nkramer44  suggested [here](https://github.com/XRPLF/xrpl4j/issues/85#issuecomment-811388126) as it seemed unnecessary.

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/server/ServerInfo.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/server/ServerInfo.java
@@ -14,6 +14,7 @@ import org.immutables.value.Value.Immutable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.math.BigDecimal;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/server/ServerInfo.java
```diff
@@ -192,18 +192,18 @@ default boolean isLedgerInCompleteLedgers(final UnsignedLong ledgerIndex) {
    * individual server's load factor, the cluster's load factor, the open ledger cost and the overall network's load
    * factor.
    *
-   * @return An {@link UnsignedInteger} representing the load factor.
+   * @return An optionally-present {@link Double} representing the load factor.
    */
   @JsonProperty("load_factor")
-  UnsignedInteger loadFactor();
+  Optional<Double> loadFactor();
```
sappenin, line 198: I'm coming late to this party, but I'd like us to reconsider the use of `Double`. My first inclination was BigDecimal on grounds of precision/resolution/etc, but haven't looked too deeply into the XRPL.org docs. Thoughts?

nkramer44: The xrpl.org docs are somewhat unclear, they just say this is a "Number".  To be safe I think `BigDecimal` is probably the way to go.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/server/ServerInfoResultTests.java
```diff
@@ -335,7 +335,7 @@ private ServerInfo serverInfo(final String completeLedgers) {
         )
         .threads(UnsignedLong.valueOf(6))
         .build())
-      .loadFactor(UnsignedInteger.ONE)
+      .loadFactor(1.0)
```
nkramer44, line 338: A little more testing here could be useful. I'd suggest a few tests with differing precisions in JSON, as well as a test without any decimal precision, just to make sure this change covers everything rippled will send back

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/server/ServerInfo.java
tommq: I agree that BigDecimal is better for this

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/server/ServerInfoResultTests.java
nkramer44: Hey @tommq, just wanted to make sure you saw this comment.

nkramer44 (approved):
LGTM! Thanks for doing this @tommq

sappenin (approved):
LGTM!

PR #91, opened 2021-04-13
(BREAKING CHANGE) Make fields in LedgerResult and LedgerHeader Optional 

### LINKED ISSUE #86: LedgerResult and LedgerHeader objects break when asking for current ledger
The `ledgerHash` and `ledgerIndex` fields are not `Optional` in `LedgerResult`, nor are `closeTimeHuman` and `parentCloseTime` in `LeaderHeader`.  These fields are always present in responses to requests for `validated` ledgers, but are absent when asking for the `current` ledger.

We should probably go ahead and make the breaking change of making these `Optional`

nkramer44:
Fixes #86.

This is a breaking change

### DIFF
--- a/xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/ledger/LedgerHeaderJsonTests.java
+++ b/xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/ledger/LedgerHeaderJsonTests.java
@@ -10,6 +10,7 @@ import org.junit.Test;
 import org.skyscreamer.jsonassert.JSONAssert;
 import org.skyscreamer.jsonassert.JSONCompareMode;
 import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
 import org.xrpl.xrpl4j.model.client.transactions.TransactionResult;
 import org.xrpl.xrpl4j.model.flags.Flags;
 import org.xrpl.xrpl4j.model.transactions.Address;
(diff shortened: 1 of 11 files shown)


xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/ledger/LedgerHeaderJsonTests.java
```diff
@@ -77,7 +78,7 @@ public void deserializeLedgerHeaderWithTransactions() throws JsonProcessingExcep
         .build();
 
     String json = "{\n" +
-        "  \"ledger_index\" : \"54300940\",\n" +
```
sappenin, line 80: I think this needs to serialize as a String, because can't a `LedgerIndex` occasionally take a [string form](https://xrpl.org/xrp-api-data-types-ledgerindex.html#ledgerindex)?

e.g., couldn't this be `validated`?

nhartner: This is just the JSON that the client sends right? In which case Jackson doesn't require (but also allows) quotes around the number.

nhartner: We should add a test though that verifies serialization using `"ledger_index": "validated"`

nkramer44: 2 things here:
1) `LedgerHeader`s always come from rippled responses, and `LedgerIndex`es in responses are always numbers, not `String`s, which deserialize fine because of our custom deserializer.
2) For some reason `LedgerHeader.ledgerIndex` comes back from rippled as a numerical `String`, not just a number. This is fine because of point (1), but these JSON tests were failing because `LedgerIndex` gets serialized to a number, so I just fixed the JSON

sappenin: I sync'd with @nkramer44 and he told me why this is all correct, and I concur. He's going to write a unit test for `"ledger_index": "validated"` as a final verification.

nkramer44: @nhartner I think we should unit test `LedgerIndexSerializer` and `LedgerIndexDeserializer`, but I don't think unit testing `"ledger_index": "validated"` makes sense for result objects, since `ledger_index` will always be a number (or a numerical `String`) in response objects. How's that sound?

nkramer44: @nhartner Just added tests. Lmk what you think

sappenin (approved):
LGTM - thanks for adding those final unit tests.

Issue #93, opened 2021-04-14
Add validation to LedgerIndex


nkramer44:
`LedgerIndex` is just a wrapper around `String`. There are three defined character `String`s that are recognized by rippled as shortcuts: `"validated"`, `"current"`, and `"closed"`.  All other ledger indexes must be a numerical `String` like `"12345"`. 

Currently, we don't do any client side validation that the `String` wrapped by `LedgerIndex` is either one of the shortcut Strings or a number, so a user could technically construct a `LedgerIndex` wrapping `"foo"`.  While this will ultimately be caught by rippled, we should add validation in the `LedgerIndex` class to short circuit any errors.

tekbar11:
Hi. This doesn't seem very difficult. I'd like to work on this issue if it's okay.

sappenin:
@tekbar11 thanks, and that sounds great! I'll assign this task to you. I think the only outstanding question in my mind is, "what should happen if someone does: `LedgerIndex.of("foo")`? If an exception should be thrown, which kinds?

@nkramer44 @nhartner for input here too.

tekbar11:
Thanks @sappenin for assigning me to this. In my opinion, the `LedgerIndex` factory should throw an IllegalArgumentException, but tests seem to expect a NumberFormatException in all cases.

I would also like to create a nested *enum* to represent shortcut values (for modularity's sake and easy iteration on possible values), but maybe this is overkill? I have no idea if you plan to add more shortcuts.

Also, shouldn't the constructor be private, since there is a public factory method?

sappenin:
> Also, shouldn't the constructor be private, since there is a public factory method?

This seems fined to me, but it would be a potentially breaking change, so maybe let's just mark the public constructor as `@Deprecated` with a Javadoc comment that it will be removed in a future version?

tekbar11:
Hi, since my pull request has been merged, shouldn't this issue be considered as fixed? Thanks.

sappenin:
Yes, thanks for pointing this out. This is fixed via #98

PR #98, opened 2021-04-22
add validation to LedgerIndex

tekbar11:
Proposed fix for #93 .

Added a method `isValid` to LedgerIndex to check the validity of the wrapped index.

The `LedgerIndex.of` factory method now throws a `NumberFormatException` if the given value is not a valid index.

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
@@ -47,14 +47,12 @@ public class LedgerIndex {
    */
   public static LedgerIndex of(String value)
   throws NumberFormatException {
-    if (value == null) {
-      throw new NullPointerException();
-    }
-    LedgerIndex li = new LedgerIndex(value);
-    if (li.isValid()) {
-      return li;
+    Objects.requireNonNull(value);
+    if (isValidShortcut(value)) {
+      return new LedgerIndex(value);
     } else {
-      throw new NumberFormatException(value);
+      UnsignedLong.valueOf(value);
+      return new LedgerIndex(value);
     }
   }
 
@@ -68,9 +66,7 @@ public class LedgerIndex {
    * @throws NullPointerException if value is null
    */
   public static LedgerIndex of(UnsignedLong value) {
-    if (value == null) {
-      throw new NullPointerException();
-    }
+    Objects.requireNonNull(value);
     return new LedgerIndex(value.toString());
   }
 
@@ -109,22 +105,11 @@ public class LedgerIndex {
     return plus(other.unsignedLongValue());
   }
 
-  /**
-   * Check whether or not the index wrapped by this {@link LedgerIndex} is valid.
-   * A valid index is a shortcut or an unsigned long.
-   *
-   * @return true if the wrapped index is valid, false otherwise.
-   */
-  public boolean isValid() {
+  public static boolean isValidShortcut(String value) {
     if (value.equals("current")) return true;
     if (value.equals("validated")) return true;
     if (value.equals("closed")) return true;
-    try {
-      unsignedLongValue();
-      return true;
-    } catch (NumberFormatException e) {
-      return false;
-    }
+    return false;
   }
 
   @Override
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
@@ -41,16 +41,18 @@ public class LedgerIndex {
    *
    * @return A {@link LedgerIndex} with the given value.
    *
+   * @throws NullPointerException if value is null
+   *
    * @throws NumberFormatException if value is an invalid index
    */
   public static LedgerIndex of(String value)
   throws NumberFormatException {
     if (value == null) {
-      throw new NumberFormatException("Cannot build LedgerIndex from null value.");
+      throw new NullPointerException();
     }
     LedgerIndex li = new LedgerIndex(value);
     if (li.isValid()) {
-      return new LedgerIndex(value);
+      return li;
     } else {
       throw new NumberFormatException(value);
     }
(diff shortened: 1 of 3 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -34,9 +34,17 @@ public LedgerIndex(String value) {
    * @param value A {@link String} containing either an integer or a shortcut.
    *
    * @return A {@link LedgerIndex} with the given value.
+   *
+   * @throws NumberFormatException if value is an invalid index
    */
-  public static LedgerIndex of(String value) {
-    return new LedgerIndex(value);
+  public static LedgerIndex of(String value)
```
sappenin, line 40: Let's add some unit test coverage for the three text-values, and probably for a `null` value too?

sappenin (changes requested):

tekbar11:
Marked constructor in `LedgerIndex` as deprecated, with a javadoc comment explaining the reason of the deprecation.

Added unit tests to `LedgerIndexTest` for `LedgerIndex`'s factory methods.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
tekbar11: I added some unit tests as requested.

I'd also like to point out that the `LedgeIndexTest.createInvalidNumericalLedgerIndex` could be removed because it does not make sense to test the `LedgerIndex.plus` method after bad calls to `LedgerIndex.of`.

nkramer44 (approved):
Looks good to me! Thanks @tekbar11 for doing this!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -45,8 +64,13 @@ public static LedgerIndex of(String value) {
    * @param value An {@link UnsignedLong} specifying a ledger index.
    *
    * @return A {@link LedgerIndex} with the given value as a {@link String}.
+   *
+   * @throws NullPointerException if value is null
    */
   public static LedgerIndex of(UnsignedLong value) {
+    if (value == null) {
```
nkramer44, line 71: You could replace this with `Object.requireNonNull(value)`, which may make Codecov happy

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -34,9 +40,22 @@ public LedgerIndex(String value) {
...
+   *
+   * @throws NullPointerException if value is null
+   *
+   * @throws NumberFormatException if value is an invalid index
    */
-  public static LedgerIndex of(String value) {
-    return new LedgerIndex(value);
+  public static LedgerIndex of(String value)
+  throws NumberFormatException {
+    if (value == null) {
```
sappenin, line 50: As a style, this should be: `Objects.requireNonNull(value);`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
sappenin: Yes, also see comment above. As a style, this should be: `Objects.requireNonNull(value);`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -85,6 +109,24 @@ public LedgerIndex plus(LedgerIndex other) {
...
+   */
+  public boolean isValid() {
+    if (value.equals("current")) return true;
+    if (value.equals("validated")) return true;
+    if (value.equals("closed")) return true;
+    try {
+      unsignedLongValue();
+      return true;
+    } catch (NumberFormatException e) {
+      return false;
```
sappenin, line 126: Is this a phantom comment, or is this actually not covered? Seems like line 31 of the unit test should trigger this. Am I missing something? Might be worth "run with coverage" in Intellij just to double-check.

sappenin (changes requested):
This is really close - some very minor changes and I think we're good to go. Thanks for doing this!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
sappenin: Yes, I agree -- it looks like your new test (see [here](https://github.com/XRPLF/xrpl4j/pull/98/files#diff-ba78bf512011908df564830de4f68373ea102c5f6ec9b3dea7a1febef3936615R31)) covers everything that was covered by `createInvalidNumericalLedgerIndex` -- feel free to remove it and verify that test coverage is the same?

tekbar11:
I fixed the code coverage loss I introduced, jacoco now only highlights the code that was already not considered as covered before I made changes to `LexerIndex`.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
nkramer44: I was confused about this as well. I can pull this down and run with coverage. Sometimes Codecov isn't 100% accurate

tekbar11: Yes this should have been considered covered, but jacoco seems to not report `throw` statements as covered.
From [the jacoco FAQ](https://www.eclemma.org/jacoco/trunk/doc/faq.html):

**Source code lines with exceptions show no coverage. Why?**

*JaCoCo determines code execution with so called probes. Probes are inserted into the control flow at certain positions. Code is considered as executed when a subsequent probe has been executed. In case of exceptions such a sequence of instructions is aborted somewhere in the middle and the corresponding line of source code is not marked as covered.*

I fixed this in my latest commit, now the code coverage should be OK.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
tekbar11: done

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
tekbar11: done

sappenin (approved):

PR #106, opened 2021-05-18
Un-deprecate Signer class

### LINKED ISSUE #102: Transaction.signers is @Deprecated when it shouldn't be
I think `Transaction.signers()` inadvertently got deprecated when we added `SignatureService`. This is a standard transaction field and should not be deprecated.

nkramer44:
Fixes #102


sappenin (changes requested):
While we probably want to maintain this class, doesn't it make sense to move it to the `xrpl4j-crypto` module, and thus keep this one deprecated?

sappenin:
Oh, my bad - I made the same mistake as when I deprecated this class. Nm my comment above.

sappenin (approved):

PR #120, opened 2021-06-28
Add option to retrieve domain value in ledger transactions

francisrosario:
Transaction.java needed to have an option to retrieve transaction domain value... I've been creating an NFT Indexer but it seems I'm only able to retrieve sourceTag, account address, fee, transaction type...

With the changes below, I'm able to retrieve the domain value of a specific ledger transaction.


nkramer44:
Hey @francisrosario did you want to keep this PR open or can we close it?

PR #121, opened 2021-07-13
adds account_currencies and account_offers request models

ledhed2222:
Title says it all :)

### DIFF
--- a/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
+++ b/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
@@ -15,12 +15,16 @@ import org.xrpl.xrpl4j.keypairs.KeyPairService;
 import org.xrpl.xrpl4j.model.client.XrplMethods;
 import org.xrpl.xrpl4j.model.client.accounts.AccountChannelsRequestParams;
 import org.xrpl.xrpl4j.model.client.accounts.AccountChannelsResult;
+import org.xrpl.xrpl4j.model.client.accounts.AccountCurrenciesRequestParams;
+import org.xrpl.xrpl4j.model.client.accounts.AccountCurrenciesResult;
 import org.xrpl.xrpl4j.model.client.accounts.AccountInfoRequestParams;
 import org.xrpl.xrpl4j.model.client.accounts.AccountInfoResult;
 import org.xrpl.xrpl4j.model.client.accounts.AccountLinesRequestParams;
 import org.xrpl.xrpl4j.model.client.accounts.AccountLinesResult;
 import org.xrpl.xrpl4j.model.client.accounts.AccountObjectsRequestParams;
 import org.xrpl.xrpl4j.model.client.accounts.AccountObjectsResult;
+import org.xrpl.xrpl4j.model.client.accounts.AccountOffersRequestParams;
+import org.xrpl.xrpl4j.model.client.accounts.AccountOffersResult;
 import org.xrpl.xrpl4j.model.client.accounts.AccountTransactionsRequestParams;
 import org.xrpl.xrpl4j.model.client.accounts.AccountTransactionsResult;
 import org.xrpl.xrpl4j.model.client.channels.ChannelVerifyRequestParams;
@@ -251,6 +255,25 @@ public class XrplClient {
   }
 
   /**
+   * Get the {@link AccountCurrenciesResult} for the account specified in {@code params} by making an account_currencies
+   * method call.
+   *
+   * @param params The {@link AccountCurrenciesRequestParams} to send in the request.
+   *
+   * @return The {@link AccountCurrenciesResult} returned by the account_currencies method call.
+   *
+   * @throws JsonRpcClientErrorException If {@code jsonRpcClient} throws an error.
+   */
+  public AccountCurrenciesResult accountCurrencies(AccountCurrenciesRequestParams params) throws JsonRpcClientErrorException {
+    JsonRpcRequest request = JsonRpcRequest.builder()
+            .method(XrplMethods.ACCOUNT_CURRENCIES)
+            .addParams(params)
+            .build();
+
+    return jsonRpcClient.send(request, AccountCurrenciesResult.class);
+  }
+
+  /**
    * Get the {@link AccountInfoResult} for the account specified in {@code params} by making an account_info method
    * call.
    *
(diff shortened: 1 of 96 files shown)


ledhed2222:
AFAICT those failed coverage reports don't relate this this..? could use help

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesResult.java
```diff
@@ -0,0 +1,65 @@
...
+        return ImmutableAccountCurrenciesResult.builder();
+    }
+
+    /**
+     * The identifying Hash of the ledger version used to generate this response.
+     *
+     * @return A {@link Hash256} containing the ledger hash.
+     */
+    @JsonProperty("ledger_hash")
+    Hash256 ledgerHash();
```
nkramer44, line 31: I think this may need to be an `Optional<Hash256>`. The docs say "May Be Omitted" and I think all the other result objects with a `ledgerHash` are typed as `Optional<Hash256>`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/CurrencyAmount.java
```diff
@@ -12,7 +12,7 @@
 public interface CurrencyAmount {
 
   /**
-   * Handle this {@link CurrencyAmount} depending on its actualy polymorphic sub-type.
+   * Handle this {@link CurrencyAmount} depending on its actually polymorphic sub-type.
```
nkramer44, line 15:
```suggestion
   * Handle this {@link CurrencyAmount} depending on its actual polymorphic sub-type.
```

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesRequestParamsJsonTests.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.common.primitives.UnsignedInteger;
+import org.json.JSONException;
+import org.junit.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.transactions.Address;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.Marker;
+
+public class AccountCurrenciesRequestParamsJsonTests extends AbstractJsonTest {
```
nkramer44, line 13: I think if you create a test and don't specify `ledgerIndex` when constructing `AccountCurrenciesRequestParams`, codecov will stop yelling at you

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesResultJsonTests.java
```diff
@@ -0,0 +1,45 @@
...
+import org.json.JSONException;
+import org.junit.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.transactions.Address;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.Marker;
+import org.xrpl.xrpl4j.model.transactions.XrpCurrencyAmount;
+
+public class AccountCurrenciesResultJsonTests extends AbstractJsonTest {
```
nkramer44, line 15: Can we add a test here to make sure the JSON works when requesting with `"ledgerIndex": "current"`? We've run into problems with other result objects with differences between what rippled sends back for the current ledger.

I would just use PostMan or Curl to send a request to rippled and copy/paste the response into a test.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/AccountOffersRequestParamsJsonTests.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.common.primitives.UnsignedInteger;
+import org.json.JSONException;
+import org.junit.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.transactions.Address;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.Marker;
+
+public class AccountOffersRequestParamsJsonTests extends AbstractJsonTest {
```
nkramer44, line 13: Same thing here about the `"ledgerIndex": "current"` test

nkramer44 (changes requested):

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesResultJsonTests.java
ledhed2222: tested it with `current` using xrpl-py - works fine for a response. not quite sure if that's what you were asking?

nkramer44: yeah that's probably good enough

nkramer44 (approved):
LGTM. Thanks so much for doing this!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesRequestParams.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Request parameters for the account_currencies rippled method.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableAccountCurrenciesRequestParams.class)
+@JsonDeserialize(as = ImmutableAccountCurrenciesRequestParams.class)
+public interface AccountCurrenciesRequestParams extends XrplRequestParams {
+
+    static ImmutableAccountCurrenciesRequestParams.Builder builder() {
```
sappenin, line 22: Minor nit (so minor I hate to even bring it up) but we're trying to make the project to use two-spaces for all indentations so we can keep commit noise lower in the general case (looks like you're using 4 spaces). [Checkstyle](https://github.com/XRPLF/xrpl4j/blob/main/checkstyle.xml#L162) will catch this during the build, but the build plugin is only set to `WARN` because a good chunk of the code hasn't been brought into line.

(I'm literally being a meme from a Silicon Valley episode with this comment...bucket-list checked).

Anyway, mind fixing? Otherwise, the PR LGTM!

sappenin (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountCurrenciesRequestParams.java
ledhed2222: no - i totally understand! i didn't know how to import the .idea project settings into intellij and assumed that that happened automatically. while i was here, I reformatted the `XrplClient` as well, since I touched that file, which accounts for the large number of line changes in the most recent commmit

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -100,19 +104,17 @@ public XrplClient(HttpUrl rippledUrl) {
...
-   *
    * @return The {@link SubmitResult} resulting from the submission request.
-   *
    * @throws JsonRpcClientErrorException If {@code jsonRpcClient} throws an error.
    * @see "https://xrpl.org/submit.html"
    */
   public <T extends Transaction> SubmitResult<T> submit(
-    Wallet wallet,
-    T unsignedTransaction
+      Wallet wallet,
```
nkramer44, line 112: I think these should all also be 2 spaces, not 4?

PR #125, opened 2021-07-20
Add a LedgerSpecifier class to replace ledgerIndex and ledgerHash in request objects

nkramer44:
Adds a new immutable class called `LedgerSpecifier`, which will allow developers to specify ledger versions in `XrplRequestParams` more clearly. 

This PR does not actually change any of the `XrplRequestParams` and is strictly additive. Another PR will alter all `XrplRequestParams` extensions by removing `ledgerIndex` and `ledgerHash` fields and replacing them with a `LedgerSpecifier` field.

PR #126, opened 2021-07-22
Make use of new LedgerSpecifier

nkramer44:
Builds on #125 by using the new classes throughout the codebase. All `XrplRequestParams` extensions that previously had `ledgerHash` and `ledgerIndex` fields now have a `LedgerSpecifier ledgerSpecifier()` field instead. All uses of the old `LedgerIndex` class are gone as well and replaced by the new implementation.

Also, I realized that `AccountTransactionsRequestParams` could use some enhancements. `ledgerIndexMin` and `ledgerIndexMax` were typed as the old `LedgerIndex` which wrapped a `String`. This worked well because the default values for those fields is `-1`. The new implementation of `LedgerIndex` wraps an an `UnsignedLong`, so negative values are clearly unsupported by `LedgerIndex`. To fix this, I added a `LedgerIndexBound` that wraps a `Long` and bounds numbers to `(-1, 0], (1, Long.MAX_VALUE)`. The `account_tx` method also only supports the `"validated"` shortcut word, so that is enforced in the object as well.


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountChannelsRequestParams.java
```diff
@@ -47,22 +44,15 @@
...
-   * The ledger index of the ledger to use, or a shortcut string to choose a ledger automatically.
-   *
-   * @return A {@link LedgerIndex}.  Defaults to {@link LedgerIndex#CURRENT}.
-   */
-  @JsonProperty("ledger_index")
   @Value.Default
-  default LedgerIndex ledgerIndex() {
-    return LedgerIndex.CURRENT;
+  @JsonUnwrapped
+  default LedgerSpecifier ledgerSpecifier() {
```
theotherian, line 54: Is there a risk of breaking clients making this change as it currently stands? Seems like this new method is replacing the old one?

nkramer44: Yes this will require a 3.0 release. This is fine because we also have some other breaking changes coming down the pipeline

nkramer44:
> Is the intention to have a release of what's in #125 and then immediately merge this in and cut another release so that there's a bridge point from normal -> deprecated -> removed

No, I just split up the PRs so that they didn't get too big to review. Yes, normally it's best to deprecate, then release, then remove and re-release. 

However, I feel like at this stage of development and use of the library, we're not really risking much by just making breaking changes like this without a deprecation period.  These changes are strictly enhancements to the API to improve developer experience. There is really nothing functionally changing about how this stuff works. My feeling is, if you aren't ready to update to the new stuff, then don't. Your older version will still work, it might just not be as nice to use as the newest version.

Obviously feel free to push back on me here :)

theotherian:
🤔  I'm of two minds on this...

On one hand, there's a trigger discipline aspect to this of making backwards incompatible changes in a pattern of having at least one "bridge" version and then one that forces migration. That's not a complicated thing to do of course, it just takes a little more time.

To your point through, our user base isn't so considerable that making backwards incompatible changes is likely to have a significant impact. 

I do feel a bit that since you went through the effort of making the deprecation changes at all that the sensible thing would be to follow with a deprecated release to actually leverage them. Otherwise it wasn't really worth doing in the first place if you're immediately breaking backwards compatibility. 

If the rationale was to deprecate for the sake of splitting a PR, then we should have a discussion about that as a general practice.

nkramer44:
I think you're right on this. We should probably be more disciplined about deprecating things and releasing rather than making breaking changes and releasing major versions left and right.

You've convinced me to add more deprecation notices in #125 and release a v2.1.0. There are a few features that we probably want to release without introducing breaking changes while we work on things like this PR and a new signing library (among other breaking changes). Plus, this will give consumers a heads up that the current `LedgerIndex` and `ledgerHash` fields are going away soon.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountChannelsRequestParams.java
sappenin: Resolving this comment now that this PR is non-breaking.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/AccountChannelsRequestParamsJsonTests.java
```diff
@@ -2,59 +2,187 @@
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.primitives.UnsignedInteger;
+import com.google.common.primitives.UnsignedLong;
 import org.json.JSONException;
 import org.junit.jupiter.api.Test;
 import org.xrpl.xrpl4j.model.AbstractJsonTest;
 import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndexShortcut;
```
sappenin, line 10: Unused?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/Wrappers.java
```diff
@@ -0,0 +1,169 @@
...
+   * This class is similar to {@link LedgerIndex} in that it represents a numerical ledger version on the XRP Ledger.
+   * However, {@link LedgerIndexBound} can be used to specify ranges of ledgers in some rippled API request objects,
+   * such as {@link org.xrpl.xrpl4j.model.client.accounts.AccountTransactionsRequestParams}. Unlike {@link LedgerIndex}
+   * which wraps an unsigned long value, this class supports {@code -1} and does not allow {@code 0}.
+   */
+  @Value.Immutable
+  @Wrapped
+  @JsonSerialize(as = LedgerIndexBound.class)
+  @JsonDeserialize(as = LedgerIndexBound.class)
+  abstract static class _LedgerIndexBound extends Wrapper<Long> implements Serializable {
```
sappenin, line 27: Should this be typed as an `UnsignedInt` instead of `Long`?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerRequestParams.java
```diff
@@ -81,7 +104,8 @@ default boolean accounts() {
 
   /**
    * If true, return information on transactions in the specified ledger version. Defaults to false.
-   * Ignored if you did not specify {@link #ledgerHash()}.
+   * Ignored if you did not specify a {@code} ledgerHash in
```
sappenin, line 107: Minor nit: Formatting seems off.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/common/LedgerIndexBoundTests.java
```diff
@@ -0,0 +1,209 @@
+package org.xrpl.xrpl4j.model.client.common;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
sappenin, line 4: Unused.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/common/LedgerIndexShortcutTest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+import org.json.JSONException;
+import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.skyscreamer.jsonassert.JSONCompareMode;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndexShortcut;
```
sappenin, line 14: Unused.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/common/LedgerIndexTest.java
```diff
@@ -4,8 +4,18 @@
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.primitives.UnsignedInteger;
 import com.google.common.primitives.UnsignedLong;
```
sappenin, line 12: Unused.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/common/LedgerSpecifierTest.java
```diff
@@ -0,0 +1,284 @@
...
+
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
+import com.google.common.primitives.UnsignedInteger;
+import com.google.common.primitives.UnsignedLong;
```
sappenin, line 15: Unused imports.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -15,22 +18,28 @@
...
+   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexShortcut}.
    */
+  @Deprecated
   public static final LedgerIndex CURRENT = LedgerIndex.of("current");
 
   /**
    * Constant shortcut value to request the most recent ledger that has been validated by consensus.
    *
    * @see "https://xrpl.org/basic-data-types.html#specifying-ledgers"
+   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexShortcut}.
```
sappenin, line 30:
```suggestion
   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexSpecifier}.
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -15,22 +18,28 @@
    * Constant shortcut value to request a rippled server's current working version of the ledger.
    *
    * @see "https://xrpl.org/basic-data-types.html#specifying-ledgers"
+   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexShortcut}.
```
sappenin, line 21:
```suggestion
   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexSpecifier}.
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerIndex.java
```diff
@@ -15,22 +18,28 @@
...
    */
+  @Deprecated
   public static final LedgerIndex VALIDATED = LedgerIndex.of("validated");
 
   /**
    * Constant shortcut value to request a the most recent ledger that has been closed for modifications
    * and proposed for validation.
    *
    * @see "https://xrpl.org/basic-data-types.html#specifying-ledgers"
+   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexShortcut}.
```
sappenin, line 40:
```suggestion
   * @deprecated Ledger index shortcut values should now be specified using {@link LedgerIndexSpecifier}.
```

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/ledger/LedgerResultJsonTests.java
```diff
@@ -20,28 +20,28 @@
...
-          .transactionHash(Hash256.of("FC6FFCB71B2527DDD630EE5409D38913B4D4C026AA6C3B14A3E9D4ED45CFE30D"))
-          .build()
-      )
-      .build();
+        .status("success")
+        .ledgerHash(Hash256.of("3652D7FD0576BC452C0D2E9B747BDD733075971D1A9A1D98125055DEF428721A"))
+        .ledgerIndex(LedgerIndex.of(UnsignedInteger.valueOf(54300940)))
+        .validated(true)
+        .ledger(
+            LedgerHeader.builder()
```
sappenin, line 28: Formatting?

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/LocalRippledEnvironment.java
```diff
@@ -49,10 +48,7 @@ public void fundAccount(Address classicAddress) {
 
   protected AccountInfoResult getCurrentAccountInfo(Address classicAddress) {
     try {
-      AccountInfoRequestParams params = AccountInfoRequestParams.builder()
-        .account(classicAddress)
-        .ledgerIndex(LedgerIndex.CURRENT)
-        .build();
+      AccountInfoRequestParams params = AccountInfoRequestParams.of(classicAddress);
```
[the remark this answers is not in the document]
nkramer44: Before I decided to unbreak the breaking change, all the request param immutables had `ledgerSpecifier` as a `Value.Default` set to `LedgerSpecifier.CURRENT`, which I assume will be the case once `#ledgerIndex` and `#ledgerHash` go away.

Happy to set it explicitly here though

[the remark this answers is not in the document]
sappenin: Oh yes, you're right - forgot about that. This is fine as-is then.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/Wrappers.java
nkramer44: The whole point of this class is to allow `-1` and values greater than `0`, which won't be supported by `LedgerIndex` once it just wraps an `UnsignedInteger`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/LedgerSpecifier.java
```diff
@@ -0,0 +1,197 @@
...
+    if (ledgerIndex().isPresent()) {
+      numSpecified += 1;
+    }
+    if (ledgerIndexShortcut().isPresent()) {
+      numSpecified += 1;
+    }
+
+    Preconditions.checkArgument(
+      numSpecified == 1,
+      String.format("Exactly one Ledger specifier must be provided. %s were specified", numSpecified)
```
sappenin, line 194: This error message isn't going to make sense to a developer without them going and inspecting our source code to see what it means.

Could we make it a bit more descriptive, maybe like this:

```suggestion
      String.format("Only one Ledger specifier may be specified using one of ledgerHash, ledgerIndex or ledgerIndexShortcut. %s were specified", numSpecified)
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AccountTransactionsRequestParamsDeserializer.java
```diff
@@ -0,0 +1,104 @@
...
+      if (ledgerIndex.isNumber()) {
+        ledgerSpecifier = LedgerSpecifier.ledgerIndex(LedgerIndex.of(UnsignedInteger.valueOf(ledgerIndex.asInt())));
+      } else {
+        switch (ledgerIndex.asText()) {
+          case "validated":
+            ledgerSpecifier = LedgerSpecifier.VALIDATED;
+            break;
+          case "current":
+            ledgerSpecifier =  LedgerSpecifier.CURRENT;
+            break;
```
sappenin, line 90: We should increase test coverage here.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/LedgerSpecifierDeserializer.java
```diff
@@ -0,0 +1,63 @@
...
+      if (ledgerIndex.isNumber()) {
+        return LedgerSpecifier.ledgerIndex(LedgerIndex.of(UnsignedInteger.valueOf(ledgerIndex.asInt())));
+      } else {
+        switch (ledgerIndex.asText()) {
+          case "validated":
+            return LedgerSpecifier.VALIDATED;
+          case "current":
+            return LedgerSpecifier.CURRENT;
+          case "closed":
+            return LedgerSpecifier.CLOSED;
```
sappenin, line 56: We should increase test coverage here.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/LegacyLedgerSpecifierUtils.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public static LedgerSpecifier computeLedgerSpecifier(Optional<Hash256> ledgerHash, LedgerIndex ledgerIndex) {
+    return ledgerHash
+      .map(LedgerSpecifier::ledgerHash)
+      .orElseGet(() -> {
+        if (ledgerIndex != null) {
+          if (ledgerIndex.equals(LedgerIndex.VALIDATED)) {
+            return LedgerSpecifier.VALIDATED;
+          } else if (ledgerIndex.equals(LedgerIndex.CURRENT)) {
+            return LedgerSpecifier.CURRENT;
```
sappenin, line 33: We should increase test coverage here.

nkramer44: see https://github.com/XRPLF/xrpl4j/pull/126/files#diff-eb3ba4fa4fb59d2592f632aeffdd45be98ed34e9cb79c3e658831745e29ca24b

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AccountTransactionsRequestParamsDeserializer.java
nkramer44: see https://github.com/XRPLF/xrpl4j/pull/126/files#diff-55535dcfa4ab18a1d568c8bacab84af10fd097187fed816aa684d0fef5472dc5

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/LedgerSpecifierDeserializer.java
nkramer44: see https://github.com/XRPLF/xrpl4j/pull/126/files#diff-4b74734b8435896b1897caef211f527615f6db8bae74e562c408c0fb0662e95c

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/LegacyLedgerSpecifierUtils.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Optional;
+
+/**
+ * Simple utility class to help deprecate ledger index and ledger hash fields in request parameter objects.
+ *
+ * @deprecated This will be removed once all {@code ledgerIndex} and {@code ledgerHash} fields are removed.
+ */
+@Deprecated
+public class LegacyLedgerSpecifierUtils {
```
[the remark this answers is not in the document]
sappenin: LGTM!

sappenin (approved):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -42,35 +56,106 @@
...
+   * {@link org.xrpl.xrpl4j.model.client.common.LedgerIndexShortcut#VALIDATED}.</p>
+   *
+   * <p>Setting this value will nullify and take precedence over {@link #ledgerIndexMinimum()}
+   * and {@link #ledgerIndexMaximum()}</p>
+   *
+   * @return A {@link LedgerSpecifier} specifying the ledger version to request.
+   */
+  @JsonUnwrapped
+  @Value.Default // TODO: Make non-default once ledgerIndex and ledgerHash are gone
+  default Optional<LedgerSpecifier> ledgerSpecifier() {
```
sappenin, line 150: This gives a `warn` during compile because it's unusual for an Optional method to have a default value -- can this be made non-optional?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -1,24 +1,38 @@
 package org.xrpl.xrpl4j.model.client.accounts;
 
+import static org.xrpl.xrpl4j.model.client.LegacyLedgerSpecifierUtils.computeLedgerSpecifierFromLedgerIndex;
```
sappenin, line 3: Minor nit: this is unused.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
nkramer44: It has to be `Optional`, since in these request params you don't have to specify a `ledger_index` or `ledger_hash` (you can specify `ledgerIndexMinimum` or `ledgerIndexMaximum`). It's not ideal to have a `@Value.Default` for an `Optional`, but I think it's the only way to get the deprecation to work correctly. Once we get rid of the deprecated fields, we can get rid of the `Default`

sappenin (changes requested):
Let's remove `OldLedgerIndexDeserializerTest.java` and `OldLedgerIndexSerializerTest.java`?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
sappenin: I thought from yesterday we said the default (now and in the future) would just be `current`. So, can't we just get rid of the `Optional`, but keep the `@Default`?

sappenin (approved):

PR #160, opened 2021-09-30
fix #159

### LINKED ISSUE #159: Correct IllegalArgumentException message in `TransactionType`
If you try to construct a `TransactionType` that doesn't exist using `TransactionType.forValue(String)`, the method throws an `IllegalArgumentException`. 

This exception is correct, but the message that we attach to it is incorrect (due to copy pasting poorly). Currently, the exception is:
```
throw new IllegalArgumentException("No matching AccountSetFlag enum value for int value " + value);
```

We should fix the message by changing it to 
```
throw new IllegalArgumentException("No matching TransactionType enum value for String value " + value);
```

awydini:
fix #159

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
@@ -106,8 +106,8 @@ public enum TransactionType {
   /**
    * Gets an instance of {@link TransactionType} for the given string value.
    *
-   * @param value The {@link String} value corresponding to a {@link TransactionType}.
-   *
+   * @param value The {@link String} value corresponding to a
+   *              {@link TransactionType}.
    * @return The {@link TransactionType} with the corresponding value.
    */
   public static TransactionType forValue(String value) {
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
@@ -117,7 +117,7 @@ public enum TransactionType {
       }
     }
 
-    throw new IllegalArgumentException("No matching AccountSetFlag enum value for int value " + value);
+    throw new IllegalArgumentException("No matching TransactionType enum value for String value " + value);
   }
 
   /**
--- a/xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransactionTypeTests.java
+++ b/xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransactionTypeTests.java
@@ -4,8 +4,6 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import java.util.stream.Stream;
-
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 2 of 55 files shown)


xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransactionTypeTests.java
```diff
@@ -0,0 +1,60 @@
...
+		assertTrue(transactionType instanceof TransactionType);
+		System.out.println("=========================================OK==============================================");
+	}
+
+	@EmptySource
+	@NullSource
+	@ParameterizedTest
+	@ArgumentsSource(value = TransactionTypeInvalidArgumentProvider.class)
+	public void shouldThrowIllegalArgumentExceptionForInvalidValues(String value) {
+		System.out.println("running test shouldThrowIllegalArgumentExceptionForInvalidValues with argument " + value);
```
sappenin, line 53: Let's either use a `Logger` here or instead just consider removing the emissions. For example, are these emissions adding any value to the actual build? Seems like not because if a test fails, the assertion will indicate what went wrong down on line 55.

sappenin (changes requested):
One minor nit, but otherwise LGTM!

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransactionTypeTests.java
```diff
@@ -0,0 +1,60 @@
...
+		public Stream<? extends Arguments> provideArguments(ExtensionContext context) throws Exception {
+			return Stream.of("bla", "blaaa","123").map(Arguments::of);
+		}
+
+	}
+
+	@ParameterizedTest
+	@ArgumentsSource(value = TransactionTypeValidArgumentProvider.class)
+	public void shouldReturnTransactionTypeForValidValues(String value) {
+		System.out.println("running test shouldReturnTransactionTypeForValidValues with argument " + value);
```
sappenin, line 40: Let's remove these emissions per my comment below.

sappenin:
Also, the build is failing due to Checkstyle violations. If you run `mvn clean package` from the terminal you should be able to replicate the failures locally. You'll want to make sure formatting is correct (e.g., you're using too many spaces/tabs).

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransactionTypeTests.java
awydini: You are absolutely right.I will remove it.

nkramer44 (changes requested):
Looks good to me other than the checkstyle violations that are causing the build to fail.

It looks to me like your spaces are off -- If you are using IntelliJ, under `Preferences` -> `Editor` -> `Code Style` -> `Java` -> `Tabs and Indents`, set all three options to `2` and reformat the file. 

Thanks for making this fix!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
```diff
@@ -106,8 +106,8 @@
   /**
    * Gets an instance of {@link TransactionType} for the given string value.
    *
-   * @param value The {@link String} value corresponding to a {@link TransactionType}.
-   *
+   * @param value The {@link String} value corresponding to a
+   *              {@link TransactionType}.
```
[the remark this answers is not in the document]
sappenin: @awydini Bump...

nkramer44:
Hey @awydini Do you mind making the above changes so we can merge? Thanks!

sappenin:
This has been sitting for too long, so I addressed the comments in https://github.com/XRPLF/xrpl4j/pull/203.

Thanks for doing the heavy lifting though @awydini !

PR #178, opened 2021-11-11
New Feature: Add Nft support (DO NOT MERGE Until NFT Amendment is endorsed on mainnet)

mukulljangid:
* Add XLS-20d (NFT) support to xrpl4j.
* Specs:
    * `NfTokenMint`, `NfTokenBurn`, `NfTokenCreateOffer`, `NfTokenCancelOffer`, `NfTokenAcceptOffer` transactions.
    * `account_nfts`, `nft_buy_offers`, `nft_sell_offers` requests/methods.

Passes all CI tests.


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  Address account();
+
+  /**
+   * Limit the number of NFTs to retrieve from an account. The server is not required to honor this value.
+   * Must be within the inclusive range 10 to 400.
+   *
+   * @return An optionally-present {@link UnsignedInteger} representing the response limit.
+   */
+  Optional<UnsignedInteger> limit();
```
sappenin, line 42: What's the size of the number-type for `limit` in the actual XRP Ledger?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsResult.java
```diff
@@ -0,0 +1,81 @@
...
+  List<NfTokenObject> accountNfts();
+
+  /**
+   * The ledger index of the current open ledger, which was used when retrieving this information. Only present
+   * in responses to requests with ledger_index = "current".
+   *
+   * @return An optionally-present {@link LedgerIndex} representing the current ledger index.
+   */
+  @JsonProperty("ledger_current_index")
+  Optional<LedgerIndex> ledgerCurrentIndex();
```
sappenin, line 53: @nkramer44 is this the right pattern for LedgerIndex?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
```diff
@@ -0,0 +1,33 @@
...
+//@JsonSerialize(as = ImmutableNfToken.class)
+//@JsonDeserialize(as = ImmutableNfToken.class)
+public interface NfTokenObject {
+
+  /**
+   * The unique TokenID of the token.
+   *
+   * @return The unique TokenID of the token.
+   */
+  String tokenId();
```
sappenin, line 25: We should create a `TokenId` type here -- see `Ids.java`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NfTokenOfferObject.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * A hint indicating which page of the sender's owner directory links to this object, in case the directory
+   * consists of multiple pages.
+   * Note: The object does not contain a direct link to the owner directory containing it,
+   * since that value can be derived from the Account.
+   *
+   * @return A {@link String} containing the hint.
+   */
+  @JsonProperty("OwnerNode")
+  Optional<String> ownerNode();
```
sappenin, line 38: Is there an object for `OwnerNode` or is it just a `string` in XRPL?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NfTokenOfferObject.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * A time after which this offer is considered unfunded, as the number of seconds since
+   * the Ripple Epoch.
+   *
+   * @return The offer's expiration.
+   */
+  Optional<UnsignedInteger> expiration();
```
sappenin, line 46: Just double-checking this number-type is the correct size (and that we don't want a typed date of some kind here)

mukulljangid: No, It's the number of seconds since Ripple Epoch!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NfTokenOfferObject.java
mukulljangid: It's string and is for internal bookkeeping purposes.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
nkramer44: `* Must be within the inclusive range 10 to 400.`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Optional;
+
+/**
+ * Request parameters for the "account_ntfs" rippled API method.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableAccountNftsRequestParams.class)
+@JsonDeserialize(as = ImmutableAccountNftsRequestParams.class)
+public interface AccountNftsRequestParams {
```
nkramer44, line 18: Where do the rippled method params/result definitions live? I don't see any in the xls-20d spec

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsResult.java
nkramer44: I don't think so, but I need to see what the rippled API method is supposed to return in JSON.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
nkramer44: xls-20d states the following about `TokenID`:
> This composite field uniquely identifiers a token; it contains a set of 16 bits that identify flags or settings specific to the NFT, 16 bits that encode the transfer fee associated with this token, if
any, the 160-bit account identifier of the issuer, a 32-bit issuer-specified taxon, and an
(automatically generated) monotonically increasing 32-bit sequence number.

Do we want to try to parse this ID in our `TokenId` wrapper class or are those bits of info available by other means?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
sappenin: Cool, thanks!

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
sappenin: I'm not opposed to parsing meaning if we think it's useful. We _could_ do this in the Id wrapper, but given there's a lot of meaning, it might make sense to create a dedicated class.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
mukulljangid: https://github.com/XRPLF/XRPL-Standards/discussions/46

mukulljangid: There is no concrete documentation as of now. Greg put up his binary-codec and NFT spec PR as per his understanding from the  xls-20d spec and his conversations with Nik about it. And this might change, hence they should be left in PR.

CrazedCountryBoy (commented):
Commitment

CrazedCountryBoy:
[]()

CrazedCountryBoy:
Yes

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -610,6 +631,10 @@ private Transaction addSignature(
       return EscrowFinish.builder().from((EscrowFinish) unsignedTransaction)
         .transactionSignature(signature)
         .build();
+    } else if (NfTokenMint.class.isAssignableFrom(unsignedTransaction.getClass())) {
+      return NfTokenMint.builder().from((NfTokenMint) unsignedTransaction)
+          .transactionSignature(signature)
+          .build();
```
ledhed2222, line 637: don't we need to add the other 4 nft transactions to this?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NftIT.java
```diff
@@ -0,0 +1,78 @@
...
+  Wallet wallet1 = walletFactory.fromSeed("ssRKx55ibwH4WTmww7AH5daez15K5", true);
+
+  AccountInfoResult accountInfoResult1 = this.scanForResult(
+      () -> this.getValidatedAccountInfo(Address.of(
+          "rMp1yPjPWJdiiGvdFFbeFTeodJv17uFwgx"
+      ))
+  );
+//  AccountInfoResult accountInfoResult2 = this.scanForResult(
+//      () -> this.getValidatedAccountInfo(wallet2.classicAddress())
+//  );
```
ledhed2222, line 37: remove

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NftIT.java
```diff
@@ -0,0 +1,78 @@
...
+import org.xrpl.xrpl4j.wallet.DefaultWalletFactory;
+import org.xrpl.xrpl4j.wallet.Wallet;
+import org.xrpl.xrpl4j.wallet.WalletFactory;
+
+public class NftIT extends AbstractIT {
+  private final XrplClient xrplClient = new XrplClient(HttpUrl.parse("http://xls20-sandbox.rippletest.net:51234"));
+
+  private final WalletFactory walletFactory = DefaultWalletFactory.getInstance();
+
+//  Wallet wallet1 = walletFactory.fromSeed("shPVCWYEbwa7jnkiQ28HV6fm5uLfs", true);
```
ledhed2222, line 27: remove

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/XrplMethods.java
```diff
@@ -27,6 +27,11 @@
    */
   public static final String ACCOUNT_LINES = "account_lines";
 
+  /**
+   * Constant for the account_nfts rippled API method.
+   */
+  public static final String ACCOUNT_NFTS = "account_nfts";
```
ledhed2222, line 33: missing `nft_buy_offers` and `nft_sell_offers`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
```diff
@@ -265,4 +265,13 @@ public int getValue() {
...
   }
+
+  /**
+   * Sets an alternate account that is allowed to mint NFTokens on this
+   * account's behalf using NFTokenMint's `Issuer` field.
+   *
+   * @return An {@link Optional} field MintAccount of type {@link Address}.
+   */
+  @JsonProperty("MintAccount")
+  Optional<Address> mintAccount();
```
ledhed2222, line 276: not sure where this lives, but you will also need to add the ASF_AUTHORIZED_MINTER account set flag

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
```diff
@@ -52,6 +52,11 @@
    */
   ESCROW_FINISH("EscrowFinish"),
 
+  /**
+   * The {@link TransactionType} for the {@link NfTokenMint} transaction.
+   */
+  NFTOKEN_MINT("NFTokenMint"),
+
```
ledhed2222, line 59: don't we need to add the other 4 transaction types here?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Transaction.java
```diff
@@ -49,6 +49,7 @@
       .put(ImmutableEscrowCancel.class, TransactionType.ESCROW_CANCEL)
       .put(ImmutableEscrowCreate.class, TransactionType.ESCROW_CREATE)
       .put(ImmutableEscrowFinish.class, TransactionType.ESCROW_FINISH)
+      .put(ImmutableNfTokenMint.class, TransactionType.NFTOKEN_MINT)
```
ledhed2222, line 52: don't we need to add the other 4 transaction types here?

CrazedCountryBoy (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -1294,4 +1294,239 @@ public PaymentChannelClaimFlags build() {
...
+      public NfTokenMintFlags build() {
+        return NfTokenMintFlags.of(tfFullyCanonicalSig, tfBurnable, tfOnlyXRP, tfTrustLine, tfTransferable);
+      }
+    }
+  }
+
+  /**
+   * A set of static {@link Flags} which can be set on {@link org.xrpl.xrpl4j.model.transactions.NfTokenCreateOffer}s.
+   */
+  public static class NfTokenCreateOfferFlags extends Flags {
```
nkramer44, line 1495: We need a builder on this flags class as well.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
nkramer44: @mukulljangid The `ASF_AUTHORIZED_MINTER` flag should go in `AccountSetFlags` in this class.

Also, can you move this field declaration up with the other fields?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenBurn.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>If this operation succeeds, the corresponding NfToken is removed. If this
+ * operation empties the NfTokenPage holding the NfToken or results in the
+ * consolidation, thus removing an NfTokenPage, the owner’s reserve requirement
+ * is reduced by one.</p>
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableNfTokenBurn.class)
+@JsonDeserialize(as = ImmutableNfTokenBurn.class)
+public interface NfTokenBurn extends Transaction {
```
nkramer44, line 23: Just noticed this, but all of the `Transaction`s in this PR need to have a `flags` field. For transactions with no specific flags, just set it to 

```java
@JsonProperty("Flags")
  @Derived
  default TransactionFlags flags() {
    return new TransactionFlags.Builder().tfFullyCanonicalSig(true).build();
  }
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenBurn.java
```diff
@@ -0,0 +1,64 @@
...
+  @JsonProperty("Account")
+  Address account();
+
+  /**
+   * Identifies the NfToken object to be removed by the transaction.
+   *
+   * @return The TokenID of the NfToken to be burned.
+   */
+  @JsonProperty("TokenID")
+  String tokenId();
```
nkramer44, line 51: Type with `TokenId`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCancelOffer.java
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * <p>It is an error if an entry in this list points to an
+   * object that is not an NfTokenOffer object. It is not an
+   * error if an entry in this list points to an object that
+   * does not exist.
+   *
+   * @return Array of TokenIDs, each identifying a unique NfToken object, to cancel the offers for.
+   */
+  @JsonProperty("TokenIDs")
+  String[] tokenIds();
```
nkramer44, line 44: This should be a `List<TokenId>`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCreateOffer.java
```diff
@@ -0,0 +1,112 @@
...
+  Address account();
+
+  /**
+   * Identifies the TokenID of the NfToken object that the
+   * offer references.
+   *
+   * @return Offer is created for the token with TokenID.
+   */
+  @JsonProperty("TokenId")
+  String tokenId();
```
nkramer44, line 47: type as TokenId

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCreateOffer.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Set of {@link Flags.NfTokenCreateOfferFlags}s for this {@link NfTokenCreateOffer}.
+   *
+   * @return The {@link org.xrpl.xrpl4j.model.flags.Flags.NfTokenCreateOfferFlags} for this transaction.
+   */
+  @JsonProperty("Flags")
+  @Value.Derived
+  default Flags.TransactionFlags flags() {
+    return new Flags.TransactionFlags.Builder().tfFullyCanonicalSig(true).build();
+  }
```
nkramer44, line 111: Dont these need to be `NfTokenCreateOfferFlags`? If so, change `@Value.Derived` to `@Value.Default`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenMint.java
```diff
@@ -0,0 +1,89 @@
...
+   * Specifies the fee charged by the issuer for secondary sales of the Token,
+   * if such sales are allowed. Valid values for this field are between 0 and
+   * 50000 inclusive, allowing transfer rates between 0.000% and 50.000% in
+   * increments of 0.001%. This field must NOT be present if the
+   * `tfTransferable` flag is not set.
+   *
+   * @return An {@link Optional} field TransferFee of type {@link CurrencyAmount}.
+   */
+  @JsonProperty("TransferFee")
+  Optional<CurrencyAmount> transferFee();
```
nkramer44, line 65: I think `CurrencyAmount` is the wrong type for this. It seems like it is not an amount but rather a percentage. If so, we should create a class called `TransferFee` that wraps a `BigDecimal`. In that class, we should either round to the correct number of decimals or enforce no more than 3 decimal places and also enforce bounds (between 0% and 50%)

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.Optional;
+
+/**
+ * Structure of an NFToken stored on the ledger.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableNfTokenObject.class)
+@JsonDeserialize(as = ImmutableNfTokenObject.class)
+public interface NfTokenObject {
```
nkramer44, line 16: This class needs to go in the `@JsonSubTypes` annotation on `LedgerObject`.

nkramer44, line 16: Can you also add unit tests for this class and make sure that when you deserialize JSON to a `LedgerObject`, the resulting object is a `NfTokenObject`?

nkramer44 (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NfTokenOfferObject.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * A hint indicating which page of the sender's owner directory links to this object, in case the directory
+   * consists of multiple pages.
+   * Note: The object does not contain a direct link to the owner directory containing it,
+   * since that value can be derived from the Account.
+   *
+   * @return A {@link String} containing the hint.
+   */
+  @JsonProperty("OwnerNode")
+  Optional<String> ownerNode();
```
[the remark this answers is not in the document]
mukulljangid: OwnerNode is an Optional value. see here: https://xrpl.org/nftokenoffer.html
It is for internal bookkeeping purposes and irrelevant for users.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenAcceptOffer.java
```diff
@@ -0,0 +1,118 @@
...
+   * Identifies the NfTOffer that offers to buy the NfT.
+   *
+   * <p>In direct mode this field is optional, but either SellOffer or
+   * BuyOffer must be specified. In brokered mode, both SellOffer
+   * and BuyOffer MUST be specified.
+   *
+   * @return An {@link Optional} NfTOffer of type {@link String} that offers to buy the NfT.
+   */
+  @JsonProperty("BuyOffer")
+  Optional<String> buyOffer();
```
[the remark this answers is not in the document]
mukulljangid: Yes, these are hashes. I'll change those.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCreateOffer.java
mukulljangid: Sorry.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenMint.java
mukulljangid: It is an unsigned integer. ref: https://xrpl.org/nftokenmint.html#nftokenmint-fields

nkramer44: Ah right. In that case, we should still create a class called `TransferFee` that wraps an `UnsignedInteger`. It should do the following:
1) Check the bounds of the integer (ie <= 9999) (also are the docs correct here? Greg's code says max is 50% in increments of 0.001% but docs say max is 99.99% in increments of 0.01%)
2) Have a static constructor like `public static TransferFee ofPercent(BigDecimal percent)` so that developers can construct using percentages.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
mukulljangid: This is  just a helper object. I think NfTokenOfferObject is the one that would go on this list. ref: https://xrpl.org/nftokenoffer.html

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftBuyOffersRequestParams.java
```diff
@@ -0,0 +1,32 @@
...
+  static ImmutableNftBuyOffersRequestParams.Builder builder() {
+    return ImmutableNftBuyOffersRequestParams.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   *  @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("TokenID")
+  String tokenId();
```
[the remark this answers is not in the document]
nkramer44: I think this still needs to be re-typed?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftBuyOffersResult.java
```diff
@@ -0,0 +1,40 @@
...
+  static ImmutableNftBuyOffersResult.Builder builder() {
+    return ImmutableNftBuyOffersResult.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   * @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("TokenID")
+  String tokenId();
```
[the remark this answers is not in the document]
nkramer44: This needs to be re-typed

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftBuyOffersResult.java
```diff
@@ -0,0 +1,40 @@
...
+   * @return An {@link ImmutableNftBuyOffersResult.Builder}.
+   */
+  static ImmutableNftBuyOffersResult.Builder builder() {
+    return ImmutableNftBuyOffersResult.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   * @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
```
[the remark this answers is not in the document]
nkramer44: Can you commit this suggestion?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftSellOffersRequestParams.java
```diff
@@ -0,0 +1,32 @@
...
+  static ImmutableNftSellOffersRequestParams.Builder builder() {
+    return ImmutableNftSellOffersRequestParams.builder();
+  }
+
+  /**
+   * The TokenID of the NFToken object.
+   * @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("TokenID")
+  String tokenId();
```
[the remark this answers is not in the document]
nkramer44: needs to be re-typed

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -1294,4 +1295,353 @@ public PaymentChannelClaimFlags build() {
...
```
nkramer44, line 1496: Since these flags are for the `NfTokenCreateOffer` transaction, this should extend `TransactionFlags`.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -1294,4 +1295,353 @@ public PaymentChannelClaimFlags build() {
...
+    }
+
+    /**
+     * Require a fully canonical transaction signature.
+     *
+     * @return {@code true} if {@code tfFullyCanonicalSig} is set, otherwise {@code false}.
+     */
+    public boolean tfFullyCanonicalSig() {
+      return this.isSet(TransactionFlags.FULLY_CANONICAL_SIG);
+    }
```
nkramer44, line 1370: You can get rid of this, since it is defined in `TransactionFlags`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -1294,4 +1295,353 @@ public PaymentChannelClaimFlags build() {
...
```
nkramer44, line 1543: Once `NfTokenCreateOfferFlags` extends `TransactionFlags`, you can get rid of this method, since it is already defined in `TransactionFlags`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -1294,4 +1295,353 @@ public PaymentChannelClaimFlags build() {
...
+       */
+      public NfTokenCreateOfferFlags build() {
+        return NfTokenCreateOfferFlags.of(tfFullyCanonicalSig, tfSellToken);
+      }
+    }
+  }
+
+  /**
+   * A set of static {@link Flags} which can be set on {@link org.xrpl.xrpl4j.model.ledger.RippleStateObject}s.
+   */
```
nkramer44, line 1600:
```suggestion
  /**
   * A set of static {@link Flags} which can be set on {@link org.xrpl.xrpl4j.model.ledger.NftTokenOfferObject}s.
   */
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NfTokenOfferObject.java
```diff
@@ -0,0 +1,130 @@
+package org.xrpl.xrpl4j.model.client.nft;
```
nkramer44, line 1: this class should go with the other `LedgerObject`s in `org.xrpl.xrpl4j.model.ledger`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/TransferFeeDeserializer.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.google.common.primitives.UnsignedInteger;
+import org.xrpl.xrpl4j.model.transactions.TransferFee;
+
+import java.io.IOException;
+
+/**
+ * Custom Jackson deserializer for {@link UnsignedInteger}s.
```
nkramer44, line 12:
```suggestion
 * Custom Jackson deserializer for {@link TransferFee}s.
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCreateOffer.java
```diff
@@ -0,0 +1,112 @@
...
+  @JsonProperty("Destination")
+  Optional<Address> destination();
+
+  /**
+   * Set of {@link Flags.NfTokenCreateOfferFlags}s for this {@link NfTokenCreateOffer}.
+   *
+   * @return The {@link org.xrpl.xrpl4j.model.flags.Flags.NfTokenCreateOfferFlags} for this transaction.
+   */
+  @JsonProperty("Flags")
+  @Value.Derived
```
nkramer44, line 108: This should be `@Value.Default`, otherwise you won't be able to set this field in the builder.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransferFeeTest.java
```diff
@@ -0,0 +1,21 @@
...
+public class TransferFeeTest {
+
+  @Test
+  public void transferFeeEquality() {
+
+    assertThat(TransferFee.of(UnsignedInteger.ONE).equals(TransferFee.of(UnsignedInteger.ONE)));
+    assertThat(TransferFee.of(UnsignedInteger.valueOf(10)).equals(TransferFee.of(UnsignedInteger.valueOf(10))));
+
+    assertThat(TransferFee.ofPercent(BigDecimal.valueOf(99.99))
+      .equals(TransferFee.ofPercent(BigDecimal.valueOf(99.99))));
```
nkramer44, line 19: these should all use the `assertThat(expression).isEqualTo(expression)` syntax

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -231,4 +232,76 @@ public String toString() {
...
+    public void validateBounds() {
+      Preconditions.checkArgument(
+        FluentCompareTo.is(value()).lessThanOrEqualTo(UnsignedInteger.valueOf(9999)) &&
+          FluentCompareTo.is(value()).greaterThanEqualTo(UnsignedInteger.valueOf(0)),
+        "TransferFee should be in the range 0 to 9999.");
+    }
+
+    public boolean equals(UnsignedInteger obj) {
+      return FluentCompareTo.is(value()).equalTo(obj);
+    }
```
nkramer44, line 303: I think this is unnecessary. Developers should not be comparing `TransferFee`s with `UnsignedInteger`s. If you want to compare a `TransferFee` to something, it has to be another `TransferFee`, otherwise you lose the benefits of using `TransferFee` in the first place. If you want to compare a `TransferFee` to an `UnsignedInteger`, just do `transferFee.value().equals(unsignedInt)`.

nkramer44 (changes requested):

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransferFeeTest.java
```diff
@@ -0,0 +1,21 @@
+package org.xrpl.xrpl4j.model.transactions;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.common.primitives.UnsignedInteger;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+
+public class TransferFeeTest {
```
nkramer44, line 10: We should add a test for the `ofPercent` constructor that tries to construct using a `BigDecimal` with more than 2 decimal places (like 50.001) and ensure that an exception is thrown.

nkramer44 (changes requested):

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransferFeeTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+  @Test
+  public void transferFeeEquality() {
+
+    assertThat(TransferFee.of(UnsignedInteger.ONE).equals(TransferFee.of(UnsignedInteger.ONE)));
+    assertThat(TransferFee.of(UnsignedInteger.valueOf(10)).equals(TransferFee.of(UnsignedInteger.valueOf(10))));
```
nkramer44, line 16: Please use `assertThat(expression).isEqualTo(expression)` syntax

nkramer44 (changes requested):

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/TransferFeeTest.java
mukulljangid: Sorry to have missed those.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -283,6 +283,10 @@ public boolean equals(Object obj) {
      * @return {@link TransferFee}
      */
     static TransferFee ofPercent(BigDecimal percent) {
+      Preconditions.checkArgument(
+        Math.max(0, percent.stripTrailingZeros().scale()) <= 2,
+        "Percent value should have a maximum of 2 decimal places."
+        );
```
nkramer44: This is fine and actually probably better, but we don't technically need this. `toBigIntegerExact()` will throw an `ArithmeticException` if you have too many decimal places. An `IllegalArgumentException` with our own message is probably better for developers to see though.

nkramer44:
I am going to _not_ approve this so that we don't accidentally merge the PR, however I believe this PR looks great and we should go ahead with a dev-preview release from this branch

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenCancelOffer.java
```diff
@@ -42,7 +42,7 @@
    *
    * @return Array of TokenIDs, each identifying a unique NfToken object, to cancel the offers for.
    */
-  @JsonProperty("TokenIDs")
+  @JsonProperty("TokenOffers")
```
sappenin, line 45: @mukulljangid Can we add unit tests for this case? E.g., we should have some unit tests that validate JSON.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsResult.java
sappenin: This is resolved.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountNftsRequestParams.java
sappenin: The current code matches the binary-codec as [specified in the JS SDK here](https://github.com/XRPLF/xrpl.js/pull/1829/files#diff-3bfe85079db79c97868c6e30ff908327bb3cdb70be62cc01a7c46279a169d0a2).

francisrosario:
This is regarding xls20d, when will this class lib support xrpl4j? Any timeline for this.

sappenin:
> This is regarding xls20d, when will this class lib support xrpl4j? Any timeline for this. @sappenin

We aren't going to merge xls20d support into `main` until the spec has been finalized and shipped in rippled. For now, you can rely on maven coordinates via jitpack.io (example [here](https://jitpack.io/#xrplf/xrpl4j/nft-support-SNAPSHOT)).

Ping here if that's not obvious and I can supply an example.

mukulljangid:
Name change ref:
https://github.com/XRPLF/XRPL-Standards/discussions/46#discussioncomment-2473748

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
```diff
@@ -0,0 +1,43 @@
...
+  static ImmutableNfTokenObject.Builder builder() {
+    return ImmutableNfTokenObject.builder();
+  }
+
+  /**
+   * The unique TokenID of the token.
+   *
+   * @return The unique TokenID of the token.
+   */
+  @JsonProperty("TokenID")
```
alexandar96, line 33: Shouldn't this property be named `NFTokenID` as in other places? I've tried listing account NFTs and it threw an exception because `TokenID` was not part of JSON response.

mukulljangid: Thank you for pointing that out!

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -777,6 +802,26 @@ protected Transaction addSignature(
...
+        .transactionSignature(signature)
+        .build();
+    } else if (NfTokenCreateOffer.class.isAssignableFrom(unsignedTransaction.getClass())) {
+      return NfTokenCreateOffer.builder().from((NfTokenCreateOffer) unsignedTransaction)
+        .transactionSignature(signature)
+        .build();
+    } else if (NfTokenMint.class.isAssignableFrom(unsignedTransaction.getClass())) {
+      return NfTokenMint.builder().from((NfTokenMint) unsignedTransaction)
+          .transactionSignature(signature)
+          .build();
```
ledhed2222, line 824: outside of the scope of this PR, but this seems like it should be using a switch

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
```diff
@@ -0,0 +1,43 @@
...
+    return ImmutableNfTokenObject.builder();
+  }
+
+  /**
+   * The unique TokenID of the token.
+   *
+   * @return The unique TokenID of the token.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 34: does it make sense to call this `nftokenid`? and perhaps alias that as `id`?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftBuyOffersRequestParams.java
```diff
@@ -0,0 +1,53 @@
...
+    return ImmutableNftBuyOffersRequestParams.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   *
+   *  @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 37: ditto

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftBuyOffersResult.java
```diff
@@ -0,0 +1,43 @@
...
+    return ImmutableNftBuyOffersResult.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   *
+   * @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 36: ditto

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftSellOffersRequestParams.java
```diff
@@ -0,0 +1,53 @@
...
+    return ImmutableNftSellOffersRequestParams.builder();
+  }
+
+  /**
+   * The TokenID of the NFToken object.
+   *
+   * @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 37: this one i really think should be named after the field itself, so `NfTokenID`

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftSellOffersResult.java
```diff
@@ -0,0 +1,44 @@
...
+    return ImmutableNftSellOffersResult.builder();
+  }
+
+  /**
+   *  The TokenID of the NFToken object.
+   *
+   *  @return the TokenID of the {@link org.xrpl.xrpl4j.model.client.accounts.NfTokenObject} object.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 36: ditto

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/NfTokenBurn.java
```diff
@@ -0,0 +1,78 @@
...
+  @JsonProperty("Account")
+  Address account();
+
+  /**
+   * Identifies the NfToken object to be removed by the transaction.
+   *
+   * @return The TokenID of the NfToken to be burned.
+   */
+  @JsonProperty("NFTokenID")
+  NfTokenId tokenId();
```
ledhed2222, line 52: feels like this should be named after the field again. perhaps aliased for convenience too

ledhed2222 (commented):
left a few comments

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
mukulljangid: The method is deprecated, so I'll let it be.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
mukulljangid: nfTokenId is more descriptive and makes more sense. I change it to that

ledhed2222: the method itself should be called `nfTokenID()` instead of `tokenId()` too, right?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/NfTokenObject.java
mukulljangid: NfTokenObject is not a LedgerObject, but that applies to NfTokenOfferObject, for which the JSON tests for serializing and deserializing are in there!

nkramer44 (approved):

sappenin:
@mukulljangid Can you cleanup this branch and get it building so we can do a proper review?

nkramer44 (approved):

Issue #179, opened 2021-11-12
Remove all interning


theotherian:
We're currently using Immutables interning for at least 3 types: `Address`, `XAddress` and `XrpCurrencyAmount` (all of which are in `Wrappers` in the model package).

Doing this causes a few issues:
- Increased memory usage due to data that is both sparse and temporary in nature being long lived
- Serialization issues in Jackson due to https://github.com/FasterXML/jackson-databind/issues/2158 and https://github.com/FasterXML/jackson-databind/issues/3143

We need to stop doing this. That said, this will cause backwards compatibility issues unless you have `@Value.Immutable(builder = true)` in your annotation (despite `true` being the default) as, at least in `Wrappers`, removing `intern = true` results in the `builder()` method disappearing.

theotherian:
From looking at the code, it appears to only be these three types.

PR #180, opened 2021-11-13
Add support for gateway_balances method

theotherian:
Adds support for the XRPL `gateway_balances` method as documented here:
https://xrpl.org/gateway_balances.html

Adds the following types:
- `GatewayBalancesRequestParams`, a wrapper for the JSON RPC request as outlined in the document.
- `GatewayBalancesResult`, a top level type to deserialize the ledger response into.
- `GatewayBalancesAssets` to wrap the `assets` field of the response, which uses keys that are the address associated with the data.
- `GatewayBalancesHotWallets` to wrap the `balances` field of the response, which also uses keys that are the address associated with the data. Admittedly I took some creative license here which I will explain below.
- `GatewayBalancesObligations` to wrap the `obligations` field of the response, which uses keys that are the currency code we need to represent the data.

There are some issues with the documentation that need to be addressed, namely that the `hotwallets` field in the request is incorrectly documented as being an exclusionary field rather than inclusive (thanks @florent-uzio for catching that). Additionally, the response will wrap the addresses/balances the client seeks to include in the `hotwallets` field in a field called `balances` which is, in my opinion, too decoupled from the intent of the request. I tried to bridge this gap by naming the type `GatewayBalancesHotWallets` to try to depict a better description of the data, but I'm open to criticism or change suggestions here if there's strong opposition.

Lastly, I made no attempt to add a serializer for `GatewayBalancesResult` because, to be frank, it's just a painful response to deserialize due to using keys that should be field, and I can't imagine there would be huge demand to be able to serialize that type, at least at this moment in time.

And finally, which comes after lastly I guess, this also addresses #179.


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResult.java
```diff
@@ -0,0 +1,135 @@
...
+      .orElseThrow(() -> new IllegalStateException("Result did not contain a ledgerIndex."));
+  }
+
+  /**
+   * The ledger index of the current open ledger, which was used when retrieving this information.
+   *
+   * @return An optionally-present {@link LedgerIndex} representing the current ledger index.
+   */
+  @JsonProperty("ledger_current_index")
+  Optional<LedgerIndex> ledgerCurrentIndex();
```
nkramer44, line 123: Can we add a `ledgerCurrentIndexSafe()` method as well?

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResultsJsonTest.java
```diff
@@ -0,0 +1,483 @@
...
+            "BTC": "5908.324927635318",
+            "EUR": "992471.7419793958",
+            "GBP": "4991.38706013193",
+            "USD": "1997134.20229482"
+        },
+        "status": "success",
+        "validated": true
+    }
+
+   */
```
nkramer44, line 481: Is this still needed?

nkramer44 (changes requested):
Nice job, thanks for doing this!

Just had a few small comments. Can you also try to improve code coverage where it's missing? Looks like it's mostly on default/derived fields such as `ledgerIndexSafe`

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResultsJsonTest.java
theotherian: Whooooooops! Nope; just missed it.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResult.java
theotherian: Sure.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesRequestParams.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * The unique {@link Address} for the account, which should be an issuer. The request assets and
+   * balances associated with the issuer account.
+   *
+   * @return The {@link Address} for the account.
+   */
+  Address account();
+
+  /**
+   * An optional set of addresses of operational accounts that should not be included in the
```
florent-uzio, line 41: Should we change this description for the opposite as we know the documentation on xrpl.org is incorrect?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesRequestParams.java
```diff
@@ -0,0 +1,71 @@
...
+   * balances field of the response. Defaults to an empty set.
+   *
+   * @return An optionally specified set of operational address balances to exclude.
+   */
+  @Value.Default
+  default Set<Address> hotwallet() {
```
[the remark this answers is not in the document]
theotherian: Aye

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesRequestParams.java
theotherian: I feel comfortable saying it's incorrect based on the test actually. I created an issuer and standby and passed the standby in `hotwallet` and got it back in the results rather than having it excluded.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -234,4 +240,111 @@ protected UnsignedLong instantToXrpTimestamp(Instant instant) {
...
+   *
+   * @param currency           The currency code to send.
+   * @param value              The amount of currency to send.
+   * @param issuerWallet       The {@link Wallet} of the issuer account.
+   * @param counterpartyWallet The {@link Wallet} of the counterparty account.
+   * @param fee                The current network fee, as an {@link XrpCurrencyAmount}.
+   *
+   * @throws JsonRpcClientErrorException If anything goes wrong while communicating with rippled.
+   */
+  public void issueBalance(
```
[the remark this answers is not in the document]
theotherian: Sure, `sendIssuedCurrency` is fine by me. I don't think it needs more than that personally.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesRequestParams.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * If true, only accept an address or public key for the account parameter. Defaults to false.
+   *
+   * @return true if only accepting an address or public key for the account parameter, false otherwise.
+   */
+  @Value.Default
+  default boolean strict() {
+    return false;
+  }
```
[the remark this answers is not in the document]
theotherian: This is addressed now.

theotherian:
@nkramer44 @sappenin made some updates per the feedback. Let me know if your concerns have been addressed to your satisfaction.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResult.java
```diff
@@ -0,0 +1,149 @@
...
+   * {@link #ledgerCurrentIndex()} is empty.
+   *
+   * @return The value of {@link #ledgerIndex()}.
+   * @throws IllegalStateException If {@link #ledgerIndex()} is empty.
+   */
+  @JsonIgnore
+  @Value.Auxiliary
+  default LedgerIndex ledgerCurrentIndexSafe() {
+    return ledgerCurrentIndex()
+      .orElseThrow(() -> new IllegalStateException("Result did not contain a ledgerIndex."));
```
nkramer44, line 136:
```suggestion
      .orElseThrow(() -> new IllegalStateException("Result did not contain a ledgerCurrentIndex."));
```

nkramer44 (approved):
LGTM other than a minor nit (plz forgive me). Approving so as not to block.

sappenin (approved):
LGTM!

PR #183, opened 2021-11-17
Add PublicKey Jackson serializers

sappenin:


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/PublicKey.java
```diff
@@ -13,16 +18,17 @@
...
+import java.io.IOException;
 import java.util.Objects;
 
 /**
  * A typed instance of an XRPL Seed, which can be decoded into an instance of {@link Decoded}.
  */
-@JsonSerialize(as = ImmutableDefaultPublicKey.class)
-@JsonDeserialize(as = ImmutableDefaultPublicKey.class)
+@JsonSerialize(using = PublicKeySerializer.class)
+@JsonDeserialize(using = PublicKeyDeserializer.class)
```
nkramer44, line 31: Remind me -- why add the serializer in this annotation rather than in the Xrpl4j jackson module?

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/PublicKey.java
```diff
@@ -107,8 +111,8 @@ default String hexValue() {
    * Abstract implementation for immutables.
    */
   @Value.Immutable
-  @JsonSerialize(as = ImmutableDefaultPublicKey.class)
-  @JsonDeserialize(as = ImmutableDefaultPublicKey.class)
+  @JsonSerialize(using = PublicKeySerializer.class)
+  @JsonDeserialize(using = PublicKeyDeserializer.class)
   abstract class DefaultPublicKey implements PublicKey {
```
nkramer44, line 116: Why do we need a default implementation here, rather than just making `PublicKey` an immutable?

nkramer44 (approved):

PR #184, opened 2021-11-18
Add static builder methods to Signature

sappenin:

### DIFF
--- a/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
+++ b/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
@@ -17,6 +17,39 @@ import org.xrpl.xrpl4j.codec.addresses.UnsignedByteArray;
 public interface Signature {
 
   /**
+   * Static builder.
+   *
+   * @param unsignedByteArray A {@link UnsignedByteArray}.
+   *
+   * @return A {@link Signature}.
+   */
+  static Signature of(final UnsignedByteArray unsignedByteArray) {
+    return Signature.builder().value(unsignedByteArray).build();
+  }
+
+  /**
+   * Static builder.
+   *
+   * @param signatureBytesBase16 A base64-encoded {@link String} containing the bytes of a signature.
+   *
+   * @return A {@link Signature}.
+   */
+  static Signature fromBase16(final String signatureBytesBase16) {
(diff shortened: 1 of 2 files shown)


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
```diff
@@ -16,6 +16,39 @@
...
+  }
+
+  /**
+   * Static builder.
+   *
+   * @param signatureBytesBase16 A base64-encoded {@link String} containing the bytes of a signature.
+   *
+   * @return A {@link Signature}.
+   */
+  static Signature fromHex(final String signatureBytesBase16) {
```
nkramer44, line 48: I think having `fromHex` and `fromBase16`, as well as `hexValue` and `base16Value` is slightly confusing/unnecessary. As a developer looking at the available methods, I would probably do a double take and question things a bit if I saw this in my autocomplete ("Wait, aren't hex and base16 the same thing? Now I have to go look at xrpl4j code to make sure they're the same")

Open to pushback here...

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
```diff
@@ -16,6 +16,39 @@
...
+   * @return A {@link Signature}.
+   */
+  static Signature of(final UnsignedByteArray unsignedByteArray) {
+    return Signature.builder().value(unsignedByteArray).build();
+  }
+
+  /**
+   * Static builder.
+   *
+   * @param signatureBytesBase16 A base64-encoded {@link String} containing the bytes of a signature.
```
nkramer44, line 33:
```suggestion
   * @param signatureBytesBase16 A base16-encoded {@link String} containing the bytes of a signature.
```
```suggestion
   * @param signatureBytesBase16 A base64-encoded {@link String} containing the bytes of a signature.
```

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
sappenin: Good catch - I'll remove `fromHex` and just keep `base16`

nkramer44 (approved):

Issue #134, opened 2021-08-06
Add static builders to AccountTransactionRequestParams


sappenin:
There will be four: 

1. ledgerIndex
2. ledgerMin
3. LedgerMax
4. LedgerMin + LedgerMax

Also, deprecate the empty builder().

PR #190, opened 2021-11-23
feat: add static constructors for builders with ledger input(account_tx req)

mukulljangid:
The `account_tx` rippled request method lets you specify the ledger(s) that you want the transactions from:
    * By specifying a specific `ledger_index`
    * By specifying a `ledger_index` shortcut like "validated", "current", "closed"
    * By specifying a ledger hash in `ledger_hash`
    * By specifying a lower bound in `ledger_index_min`
    * By specifying an upper bound in `ledger_index_max`
    * By specifying a `ledger range` with both ledger_index_min and ledger_index_max

This PR adds static builder constructors to support these types and make `account_tx` request less confusing.

Adds tests to check new ones and modify previous tests to use new formats. Passes all CI tests. 

Solves Issue #134

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountInfoRequestParams.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountInfoRequestParams.java
@@ -43,8 +43,8 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    */
   static AccountInfoRequestParams of(Address account) {
     return builder()
-        .account(account)
-        .build();
+      .account(account)
+      .build();
   }
 
   /**
@@ -94,7 +94,7 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    * Always true, as {@link #account()} is always an {@link Address}.
    *
    * @return {@code true} if the account field only accepts a public key or XRP Ledger address, otherwise {@code false}.
-   *     Defaults to {@code true}.
+   *   Defaults to {@code true}.
    */
   @Value.Derived
   default boolean strict() {
@@ -107,7 +107,7 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    * querying for the data from the current open ledger.
    *
    * @return {@code true} if queue transactions should be returned in the response, {@code false} if not.
-   *     Defaults to {@code false}.
+   *   Defaults to {@code false}.
    */
   @Value.Default
   default boolean queue() {
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountInfoRequestParams.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountInfoRequestParams.java
@@ -9,7 +9,6 @@ import org.immutables.value.Value;
 import org.xrpl.xrpl4j.model.client.LegacyLedgerSpecifierUtils;
 import org.xrpl.xrpl4j.model.client.XrplRequestParams;
 import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
-import org.xrpl.xrpl4j.model.client.common.LedgerIndexShortcut;
 import org.xrpl.xrpl4j.model.client.common.LedgerSpecifier;
 import org.xrpl.xrpl4j.model.transactions.Address;
 import org.xrpl.xrpl4j.model.transactions.Hash256;
@@ -44,8 +43,8 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    */
   static AccountInfoRequestParams of(Address account) {
     return builder()
-      .account(account)
-      .build();
+        .account(account)
+        .build();
   }
 
   /**
@@ -95,7 +94,7 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    * Always true, as {@link #account()} is always an {@link Address}.
    *
    * @return {@code true} if the account field only accepts a public key or XRP Ledger address, otherwise {@code false}.
-   *   Defaults to {@code true}.
+   *     Defaults to {@code true}.
    */
   @Value.Derived
   default boolean strict() {
@@ -108,7 +107,7 @@ public interface AccountInfoRequestParams extends XrplRequestParams {
    * querying for the data from the current open ledger.
    *
    * @return {@code true} if queue transactions should be returned in the response, {@code false} if not.
-   *   Defaults to {@code false}.
+   *     Defaults to {@code false}.
    */
   @Value.Default
   default boolean queue() {
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
@@ -236,17 +236,16 @@ public interface AccountTransactionsRequestParams extends XrplRequestParams {
   @Value.Check
   default void validateSpecifierNotCurrentOrClosed() {
     ledgerSpecifier().ifPresent(
-        ledgerSpecifier -> ledgerSpecifier.handle(
-            ledgerHash -> {
-            },
-            ledgerIndex -> {
-            },
-            ledgerIndexShortcut -> Preconditions.checkArgument(
-                ledgerIndexShortcut.equals(LedgerIndexShortcut.VALIDATED),
-                "Invalid LedgerIndexShortcut. The account_tx API method only accepts 'validated' when specifying " +
-                    "a shortcut."
-            )
+      ledgerSpecifier -> ledgerSpecifier.handle(
+        ledgerHash -> {
+        },
+        ledgerIndex -> {
+        },
+        ledgerIndexShortcut -> Preconditions.checkArgument(
+          ledgerIndexShortcut.equals(LedgerIndexShortcut.VALIDATED),
+          "Invalid LedgerIndexShortcut. The account_tx API method only accepts 'validated' when specifying a shortcut."
         )
+      )
     );
   }
 
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
@@ -39,7 +39,7 @@ public interface AccountTransactionsRequestParams extends XrplRequestParams {
    *
    * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
    * @deprecated Prefer one of the other defined builders so that you are overtly
-   * specifying which ledger index or type you want.
+   *     specifying which ledger index or type you want.
    */
   @Deprecated
   static ImmutableAccountTransactionsRequestParams.Builder builder() {
@@ -207,7 +207,7 @@ public interface AccountTransactionsRequestParams extends XrplRequestParams {
    * ordered.)
    *
    * @return {@code true} if values should be indexed with the oldest ledger first, otherwise {@code false}. Defaults
-   * to {@code false}.
+   *     to {@code false}.
    */
   @Value.Default
   default boolean forward() {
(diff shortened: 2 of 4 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,70 @@
    *
    * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
    */
+  @Deprecated
```
sappenin, line 41: Let's add a message that clarifies why this is being deprecated, maybe something like this suggestion.
```suggestion
  @Deprecated Prefer one of the other defined builders so that you are overtly specifying which ledger index or type you want.
```

@nkramer44 Any thoughts here around better wording?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,70 @@
...
+    return ImmutableAccountTransactionsRequestParams.builder().ledgerSpecifier(Optional.ofNullable(ledgerSpecifier));
+  }
+
+  /**
+   * Construct a builder for this class with ledgerHash field set.
+   *
+   * @param ledgerHash Hash of type {@link Hash256} of a specific ledger.
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(Hash256 ledgerHash) {
```
sappenin, line 75: I don't think we want this builder available here. The contract we want to prod developers to be using is only `LedgerSpecifier`. If someone wants to specify a hash, then they should build a LedgerSpecifier and supply it above above on line 65.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,70 @@
...
 
+  /**
+   * Construct a builder for this class with {@link LedgerIndex} field set.
+   *
+   * @param ledgerIndex An index of type {@link LedgerIndex} to specify a particular ledger.
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(LedgerIndex ledgerIndex) {
+    return ImmutableAccountTransactionsRequestParams.builder()
+        .ledgerSpecifier(
```
sappenin, line 54: Formatting seems off -- consider pulling line 54 up to line 53.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,70 @@
...
+   * @param ledgerIndexMaximum An upper bound of ledgerIndex of type {@link LedgerIndexBound}.
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(
+      LedgerIndexBound ledgerIndexMinimum,
+      LedgerIndexBound ledgerIndexMaximum) {
+    LedgerIndexBound ledgerIndexMin = ledgerIndexMinimum;
+    LedgerIndexBound ledgerIndexMax = ledgerIndexMaximum;
+
+    if (ledgerIndexMin == null) {
```
sappenin, line 93: In this library, `null` is a bug, so we should have overt `Objects.requireNonNull(ledgerIndexMinimum);` and `Objects.requireNonNull(ledgerIndexMaximum);` so that null is not allowed.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,70 @@
...
+      ledgerIndexMin = LedgerIndexBound.of(-1);
+    }
+    if (ledgerIndexMax == null) {
+      ledgerIndexMax = LedgerIndexBound.of(-1);
```
sappenin, line 97: @nkramer44 is `-1` an allowed value for `LedgerIndexBound`?

sappenin (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
mukulljangid: Do we want to have the similar thing for LedgerIndex? Even that seems to be passed via LedgerSpecifier.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
nkramer44: I think that wording is fine.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
nkramer44: Yes, `LedgerIndexBound` is essentially an unsigned int except for `-1`, because `-1` means "give me the max of this range". For example, if you set the `ledgerIndexMinimum` to `-1`, you are asking for all ledgers that rippled has record of. I believe `-1` is different than not specifying a minimum at all...

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,69 @@
...
+   * Construct a builder for this class with {@link LedgerIndex} field set.
+   *
+   * @param ledgerIndex An index of type {@link LedgerIndex} to specify a particular ledger.
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(LedgerIndex ledgerIndex) {
+    return ImmutableAccountTransactionsRequestParams.builder().ledgerSpecifier(
+        Optional.of(LedgerSpecifier.of(ledgerIndex))
+    );
+  }
```
nkramer44, line 56: We can get rid of this one as well, for the same reason as @sappenin stated below

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -38,10 +38,69 @@
...
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(LedgerSpecifier ledgerSpecifier) {
+    return ImmutableAccountTransactionsRequestParams.builder().ledgerSpecifier(Optional.ofNullable(ledgerSpecifier));
+  }
+
+  /**
+   * Construct a builder for this class with ledgerHash field set.
+   *
+   * @param ledgerHash Hash of type {@link Hash256} of a specific ledger.
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
```
nkramer44, line 72: There should be an empty javadoc line between `@param`s and `@return`. This isn't a hard and fast rule, but we should try to be consistent across the library to reduce PR noise.

You should be able to use IntelliJ's "reformat file" tool (`option + command + L` on mac).

```suggestion
   * @param ledgerHash Hash of type {@link Hash256} of a specific ledger.
   *
   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
```

nkramer44 (changes requested):
Looking good so far. Just a few small things to fix.

Can you also see what happens if you make send an `AccountTransactionsRequestParams` with no `ledgerSpecifier` and no bounds? Does rippled throw an error, or does it return all results from all ledgers?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountInfoRequestParams.java
```diff
@@ -44,8 +43,8 @@
    */
   static AccountInfoRequestParams of(Address account) {
     return builder()
-      .account(account)
-      .build();
+        .account(account)
+        .build();
```
nkramer44, line 47: Can you reformat this file?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -28,20 +29,56 @@
 @Value.Immutable
 @JsonSerialize(as = ImmutableAccountTransactionsRequestParams.class)
 @JsonDeserialize(
-  as = ImmutableAccountTransactionsRequestParams.class,
-  using = AccountTransactionsRequestParamsDeserializer.class
+    as = ImmutableAccountTransactionsRequestParams.class,
+    using = AccountTransactionsRequestParamsDeserializer.class
```
nkramer44, line 33: Same here.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -28,20 +29,56 @@
...
+   * Construct a builder for this class with ledgerIndexRange field set.
+   *
+   * @param ledgerIndexMinimum A lower bound of ledgerIndex of type {@link LedgerIndexBound}.
+   * @param ledgerIndexMaximum An upper bound of ledgerIndex of type {@link LedgerIndexBound}.
+   *
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(
+      LedgerIndexBound ledgerIndexMinimum,
+      LedgerIndexBound ledgerIndexMaximum) {
```
nkramer44, line 70:
```suggestion
  static ImmutableAccountTransactionsRequestParams.Builder builder(
      LedgerIndexBound ledgerIndexMinimum,
      LedgerIndexBound ledgerIndexMaximum
  ) {
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -28,20 +29,56 @@
...
+    LedgerIndexBound ledgerIndexMin = ledgerIndexMinimum;
+    LedgerIndexBound ledgerIndexMax = ledgerIndexMaximum;
```
nkramer44, line 72: Why are these needed? Not sure if I'm missing something...

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
```diff
@@ -28,20 +29,56 @@
...
+   * @param ledgerSpecifier A specifier of type {@link LedgerSpecifier} to specify a ledger.
+   *
+   * @return An {@link ImmutableAccountTransactionsRequestParams.Builder}.
+   */
+  static ImmutableAccountTransactionsRequestParams.Builder builder(LedgerSpecifier ledgerSpecifier) {
+    return ImmutableAccountTransactionsRequestParams.builder().ledgerSpecifier(Optional.ofNullable(ledgerSpecifier));
+  }
+
+  /**
+   * Construct a builder for this class with ledgerIndexRange field set.
```
nkramer44, line 61:
```suggestion
   * Construct a builder for this class with {@link LedgerIndexBound}s specified for the {@link #ledgerIndexMinimum} and {@link #ledgerIndexMaximum} fields.
```

nkramer44 (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountTransactionsRequestParams.java
mukulljangid: no, these were for the old == null logic. Taking these off.

sappenin (approved):

nkramer44 (approved):
LGTM. Nice job!

Issue #170, opened 2021-11-02
Things to fix in XRPL4j-v3-dev-preview


sappenin:
**Tasks**
- [x] Signature.java is missing JSON serialization annotations.
- [x] Port over Keymetadata codec (and maybe CLI) from interledger4j.
- [x] Create new branches for v2.x backport and v3-dev-preview.
- [x] Add Jackson (de)serializer for `PublicKey` and update unit test for `SignatureWithPublicKeyTest` to check hex values of serialized objects (as opposed to just null-checks)
- [x] Can we add two static constructors to `Signature`: `Signature.of(UnsignedByteArray value)` and `Signature.ofHex(String hexValue)`? Otherwise, you need a `Signature.builder()` just to create a `Signature` which only wraps one value.
- [X] Move all BC code out of `crypto-core` and into `crypto-bc` module (See PR #204).
- [x] Finish test coverage for `UnsignedClaim` signing with KeyMetadata
- [x] #205 
- [x] #209
- [x] `Seed` constructor is `public` but should be package private or `protected`
- [x] Add a way to construct a `Seed` from a Base58 encoded `String` (the thing on the faucet called "secret")
- [x] #211
- [x] #275 
- [x] Move `EcDsaSignature` to be an implementation detail of `BcSignatureService` and then use `BcSignatureService` inside of `SingleKeyDelegatedTransactionSigner` (the signing/verifying logic is duplicated in there). See [PR here](https://github.com/XRPLF/xrpl4j/pull/270).
- [x] Normalize hex `String`s to upper case when using Guava `BaseEncoding.base16()`, such as in `PublicKey`
**WalletFactory/KeyPair Resolution**
- [x] See suggestion [here](https://github.com/XRPLF/xrpl4j/pull/204#discussion_r768197522) and consider making `Ed25519KeyPairService` and `Secp256k1KeyPairService` just implementation details of `BcWalletFactory`.
- [x] Deprecate `AddressUtils` and `WalletFactory`
- [x] Create a typed wrapper object called `Base58EncodedSecret` that is used in `Seed.java` on line 117 (instead of allowing a normal string). We can also assert that the `String` value of `Base58EncodedSecret` is base58 by doing an @Check on the character set. This way it won't be possible to pass-in the wrong value.
- [X] Update to the latest Guava dependency (e.g., [31.1](https://search.maven.org/artifact/com.google.guava/guava/31.1-jre/bundle)) -- see #297 
[...]

PR #192, opened 2021-11-23
fix #150

### LINKED ISSUE #150: Deprecate `SubmitResult.result` and `SubmitMultisignedResult.result` in favor of a new `engineResult` field
In v2.1.0, `SubmitResult` and `SubmitMultisignedResult` have `Optional<String> engineResult` and `String result`. `engineResult` is deprecated in favor of `result`, since `result` is non-optional. However, we want to keep the name `engineResult`, but make it non-optional. Therefore, we should deprecate `result`, and in v3.x.x, we can change `engineResult` to a non-optional.

mukulljangid:
fix issue #150

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/transactions/SubmitMultiSignedResult.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/transactions/SubmitMultiSignedResult.java
@@ -38,8 +38,8 @@ public interface SubmitMultiSignedResult<TxnType extends Transaction> extends Xr
    */
   @Deprecated
   @Value.Auxiliary
-  default String engineResult() {
-    return result();
+  default Optional<String> engineResult() {
+    return Optional.of(result());
   }
 
   /**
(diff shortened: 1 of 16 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/transactions/SubmitMultiSignedResult.java
```diff
@@ -34,32 +34,33 @@
    * Text result code indicating the preliminary result of the transaction, for example "tesSUCCESS".
    *
    * @return An optionally-present {@link String} containing the result of the submission.
-   * @deprecated Use {@link #result()} instead.
    */
-  @Deprecated
```
sappenin, line 39: So, first -- apologies for the communications mixup here. However, the `main` branch is still on v2, so I don't think we can actually change the contracts here until we publish a more-final `v3`, and at the moment we're still keeping that branch v2-compatible until the very last moment (i.e., until most of #170 is complete in the `v3-dev` branch). See Discussion #187 for more details around the transition.

So, for now I think the only thing to be done here is to deprecate `result` with a message that it will transition to `engineResult` in v3.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/transactions/SubmitMultiSignedResult.java
```diff
@@ -34,32 +34,33 @@
...
+  String engineResult();
 
   /**
    * Text result code indicating the preliminary result of the transaction, for example "tesSUCCESS".
    *
    * @return A {@link String} containing the result of the submission.
+   * @deprecated Use {@link #engineResult()} instead.
    */
-  @JsonProperty("engine_result")
-  String result();
```
sappenin, line 51: Need to unwind this one. See above.

sappenin (changes requested):

sappenin (changes requested):

sappenin (approved):

nkramer44 (approved):

PR #193, opened 2021-11-23
fix #172

### LINKED ISSUE #172: Add additional IT to AccountSetIT
See TODO near line 28:

```
 // TODO: Make an IT that sets all flags, and unsets only 1, and validate that only that 1 single flag was cleared.
```
Basically, this IT will be very close to what's going on in `disableAndEnableAllFlags()` except that it will only unset 1 flag instead of unsetting all flags.

mukulljangid:
enableAll and disableOne IT and verify(accountSet)
Solves #172


sappenin (approved):
LGTM, thanks for fixing this one!

nkramer44 (approved):

PR #195, opened 2021-12-02
Return `void` from `DelegatedKeyPairService.createKeyPair()`

nkramer44:
Changes the signature of `DelegatedKeyPairService.createKeyPair()` to return `void` instead of a `PublicKey`. 

Some context:
When implementing this interface on GCP KMS, I noticed that the GCP API to create a key will return successfully before the private key is actually generated. In order to return the `PublicKey` from the `createKeyPair` method, you need to make a separate call to GCP KMS. However, if you make this call directly after the GCP createKey method returns, the call will throw an exception because the key is still being generated. I believe returning `void` from `createKeyPair` and documenting this phenomenon will make it clear that developers should understand the latency issues described above and account for them in their code.


sappenin (approved):
Lgtm!

PR #204, opened 2021-12-11
Removes BouncyCastle from crypto-core module

sappenin:


xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/bc/wallet/DefaultWalletFactory.java
```diff
@@ -1,18 +1,23 @@
...
+import org.xrpl.xrpl4j.crypto.core.wallet.SeedWalletGenerationResult;
+import org.xrpl.xrpl4j.crypto.core.wallet.Wallet;
+import org.xrpl.xrpl4j.crypto.core.wallet.WalletFactory;
 
 import java.util.Objects;
 
 /**
  * Default implementation of {@link WalletFactory}.
  */
+// TODO: Call this BcWalletFactory
```
nkramer44, line 20: +1

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/AddressUtils.java
```diff
@@ -39,27 +15,6 @@ private AddressUtils() {
...
-   * Compute the RIPEMD160 of the SHA256 of the given public key, which can be encoded to an XRPL address.
-   *
-   * @param publicKey The public key that should be hashed.
-   *
-   * @return An {@link UnsignedByteArray} containing the non-encoded XRPL address derived from the public key.
-   */
-  @SuppressWarnings("UnstableApiUsage")
-  private UnsignedByteArray computePublicKeyHash(final UnsignedByteArray publicKey) {
-    Objects.requireNonNull(publicKey);
+  Address deriveAddress(final PublicKey publicKey);
```
nkramer44, line 18: What would you think about moving this method to `KeyPairService` and getting rid of this interface entirely. One could argue that deriving an address should be functionality of a `KeyPairService` because you're doing an operation/derivation from a key.

The real thing I am guarding against here is having 20 different interfaces that have a small functionality footprint. IMO, this is one of the big problems with xrpl4j v2 -- it can be really difficult to remember which service classes do what.

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/bc/signing/DerivedKeyDelegatedSignatureService.java
```diff
@@ -60,11 +61,16 @@
  * requirements, consider a different implementation of {@link DelegatedSignatureService}.</p>
  */
 @SuppressWarnings("UnstableApiUsage")
+// TODO: Make Ed25519KeyPairService and Secp256k1KeyPairService and interface, and move back to core. Then this class
+//  can
+// be moved to core also. Plus, Ed25519KeyPairService should be BcEd25519KeyPairService so that it can coexist with
+// some other type in the future.
```
nkramer44, line 67: Assuming we want to keep `KeyPairService` instead of just using `WalletFactory`, I don't think having `Ed25519KeyPairService` and `Secp256k1KeyPairService` interfaces would make sense. The key algorithm that a given implementation chooses is an implementation detail. Plus, there would be nothing differentiating the ED interface from the secp256k1 interface, as the contracts currently stand.

+1 though to renaming to `BcEd25519KeyPairService`

nkramer44: To add to our conversation from this afternoon, I firmly believe that we should get rid of `KeyPairService` and use `WalletFactory` for everything. In v2, `KeyPairService` also did things like signing and verifying transactions, so it made sense to split `WalletFactory` and `KeyPairService`. Now that the new `KeyPairService` doesn't sign or verify, it basically just does what `WalletFactory` does with different return types and method names...

Although we wanted to have in-memory <--> delegated sibling objects, `KeyPairService` and `DelegatedKeyPairService`, I think this is a fools errand. Contracts with delegated services like cloud KMS providers are so different than in-memory contracts that the only similarity in a lot of these sibling objects is naming. IMO, it is totally fine to have a set of classes/services that are used for in-memory keys -- and this will be the default for 99% of developers -- and a totally different set of classes/services that are for delegated keys.

nkramer44 (changes requested):

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/bc/signing/DerivedKeyDelegatedSignatureService.java
sappenin: Captured this feedback in #170

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/AddressUtils.java
sappenin: I agree, although I think this should all be done when we condense the KeyPairService into the WalletFactory. I've added this to #170

xrpl4j-crypto-parent/xrpl4j-crypto-bouncycastle/src/main/java/org/xrpl/xrpl4j/crypto/bc/wallet/DefaultWalletFactory.java
sappenin: This is fixed.

nkramer44 (approved):

PR #206, opened 2021-12-15
V3 Fixes #205 (Make Delegated Transaction Signer return only a Signature)

### LINKED ISSUE #205: V3: Make Delegated Transaction Signer return only a Signature.
`DelegatedTransactionSigner.multiSign()` returns a `SignatureWithKeyMetadata`. We can just return a `Signature`, since we already know the `KeyMetadata` of the key that created the signature (we pass in `KeyMetadata` to the method)

sappenin:
Changes the signature of `DelegatedSignatureService` to only return a `Signature`


sappenin:
Replaced by https://github.com/XRPLF/xrpl4j/pull/207

PR #212, opened 2021-12-15
Fixes #211 Improve the way memos are constructed

### LINKED ISSUE #211: Improve the way memos are constructed
Improve the way memos are handled. E.g., add a helper method to just add a plain-text memo as a String. See [here](https://github.com/sappenin/xrp-ping-pong/blob/main/src/main/java/com/sappenin/xrpl/pingpong/PingPong.java#L158) for an example.

sappenin:
* Add checks to make sure that memos Strings are hex-encoded
* Add a utility method to construct a memo with text/plain content-type (format)

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
@@ -27,10 +27,17 @@ import java.util.Optional;
 @JsonDeserialize(as = ImmutableMemo.class)
 public abstract class Memo {
 
-  static ImmutableMemo.Builder withPlainText(final String plainText) {
+  /**
+   * Construct a {@link ImmutableMemo.Builder} by properly hex-encoding {@code plaintext}.
+   *
+   * @param plaintext A UTF-8 {@link String}.
+   *
+   * @return A {@link ImmutableMemo.Builder}.
+   */
+  public static ImmutableMemo.Builder withPlaintext(final String plaintext) {
     return builder()
       .memoFormat(BaseEncoding.base16().encode("text/plain".getBytes(StandardCharsets.UTF_8)))
-      .memoData(BaseEncoding.base16().encode(plainText.getBytes(StandardCharsets.UTF_8)));
+      .memoData(BaseEncoding.base16().encode(plaintext.getBytes(StandardCharsets.UTF_8)));
   }
 
   /**
@@ -38,7 +45,7 @@ public abstract class Memo {
    *
    * @return An {@link ImmutableMemo.Builder}.
    */
-  static ImmutableMemo.Builder builder() {
+  public static ImmutableMemo.Builder builder() {
     return ImmutableMemo.builder();
   }
 
@@ -106,7 +113,6 @@ public abstract class Memo {
    *
    * @return {@code true} if {@code input} is hex-encoded; {@code false} otherwise.
    */
-
   private static boolean isHex(final char input) {
     return (input >= '0' && input <= '9') || (input >= 'a' && input <= 'f') || (input >= 'A' && input <= 'F');
   }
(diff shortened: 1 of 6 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
```diff
@@ -37,24 +48,66 @@
...
+   * Determines if an input char is hex-encoded.
+   *
+   * @param input A {@link char}.
+   *
+   * @return {@code true} if {@code input} is hex-encoded; {@code false} otherwise.
+   */
+
+  private static boolean isHex(final char input) {
+    return (input >= '0' && input <= '9') || (input >= 'a' && input <= 'f') || (input >= 'A' && input <= 'F');
+  }
```
nkramer44, line 112: What happens here when you pass in a `String` with an odd number of characters, which is invalid hex?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
```diff
@@ -20,7 +25,13 @@
...
 @JsonSerialize(as = ImmutableMemo.class)
 @JsonDeserialize(as = ImmutableMemo.class)
-public interface Memo {
+public abstract class Memo {
+
+  static ImmutableMemo.Builder withPlainText(final String plainText) {
+    return builder()
+      .memoFormat(BaseEncoding.base16().encode("text/plain".getBytes(StandardCharsets.UTF_8)))
+      .memoData(BaseEncoding.base16().encode(plainText.getBytes(StandardCharsets.UTF_8)));
+  }
```
nkramer44, line 34: Should we also include a static constructor that lets you pass in a plaintext value for `MemoType`?

nkramer44 (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
sappenin: Not sure really - I didn't want to go overboard here except to support a plain-text memo. `MemoType` is meant to express a Link-Relation and I feel like anyone wanting to do that will need to encoded their URL as HEX manually -- this PR is really just meant to provide a simple utility to make plain-text easier, and to perhaps illustrate to devs how this should work.

nkramer44: Not opposed to this for now. We can always add another constructor later on.

> I feel like anyone wanting to do that will need to encoded their URL as HEX manually 

Shouldn't we provide a utility for them to encode URLs to hex automatically? Is there something about hex encoding URLs that is user specific?

sappenin: @nkramer44 and I decided to leave this out for now.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
```diff
@@ -37,24 +48,66 @@
...
+
+  /**
+   * Determines if an input string is hex-encoded.
+   *
+   * @param input A {@link String}.
+   *
+   * @return {@code true} if every character in {@code input} is hex-encoded; {@code false} otherwise.
+   */
+  private static boolean isHex(final String input) {
+    Objects.requireNonNull(input);
```
sappenin, line 92: Try Guava

sappenin: rippled supports nibbles in a memo, whereas Guava does not. So this seems correct as-is.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Memo.java
sappenin: Confirmed that this works with rippled in an IT.

nkramer44 (approved):

PR #213, opened 2021-12-18
Fix seed constructor

sappenin:
Construction of a `Seed` was found to be confusing while working on [xrp-ping-pong](https://github.com/sappenin/xrp-ping-pong). The primary reason was that the [XRPL Faucet](https://xrpl.org/xrp-testnet-faucet.html) generates a value called a "secret", but this doesn't align with the [xrpl.org docs describing Key Components](https://xrpl.org/cryptographic-keys.html#key-components) (which only mention a "passphrase" or "entropy").

This PR adds a new builder method to construct a `Seed` from this "secret", which is actually just a fully-encoded Seed (i.e., 21 bytes instead of only 16 bytes of entropy).


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
```diff
@@ -100,12 +100,32 @@ public static Seed secp256k1SeedFromEntropy(final Entropy entropy) {
     return new Seed(UnsignedByteArray.of(Base58.decode(base58EncodedSeed)));
   }
 
+  /**
+   * Construct a {@link Seed} from the supplied {@code base58EncodedSecret}. Values for this function are most commonly
+   * found from an XRP Faucet, for example for the XRP devnet or testnet. On the xrpl.org documentation page, this value
+   * is often referred to as an account "secret", but it is actually just a base58-encoded string that contains an
+   * encoded 16-bytes of entropy, in addition to other binary padding and identification data.
+   *
+   * @param base58EncodedSecret A {@link Entropy} to generate a {@link Seed} from.
```
nkramer44, line 109: I think this `@param` description is off

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
```diff
@@ -116,6 +136,7 @@ public Seed(final UnsignedByteArray value) {
    */
   @VisibleForTesting
   Seed(final Seed seed) {
+    Objects.requireNonNull(seed);
```
nkramer44, line 139: Can we add a unit test for this?

nkramer44 (changes requested):

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
sappenin: This was already covered, though I adjusted the null-checks because checking the internals of a typed-seed isn't necessary because they can never really be null.

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
sappenin: Fixed. See https://github.com/XRPLF/xrpl4j/pull/213/files#diff-a28cd18969c4ffb7aadfeef4ef676525ee30001ca9a1854cacef28b8da8d5499R109

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
```diff
@@ -116,7 +136,8 @@ public Seed(final UnsignedByteArray value) {
    */
   @VisibleForTesting
   Seed(final Seed seed) {
-    this.value = Objects.requireNonNull(seed.decodedSeed().bytes());
+    this.value = Objects.requireNonNull(seed)
+      .decodedSeed().bytes();
```
nkramer44, line 140: I think this line is still untested. Can we add some unit tests for the copy constructor?

nkramer44 (changes requested):

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
sappenin: Good catch - fixed!

nkramer44 (approved):

Issue #200, opened 2021-12-10
Add workaround for UNLModify pseudotransaction in BinarySerializer


nkramer44:
See https://github.com/XRPLF/xrpl.js/pull/1830

PR #228, opened 2022-03-02
Fix Issue #200

mukulljangid:
Do not encode `Account` when txtype = `UNLModify`
Fixes https://github.com/XRPLF/xrpl4j/issues/200

### DIFF
--- a/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
+++ b/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
@@ -74,7 +74,7 @@ public class STObjectType extends SerializedType<STObjectType> {
       /**
        * The Account field must not be a part of the UNLModify pseudotransaction encoding, due to a bug in rippled.
        */
-      if (isUNLModify && "Account".equals(fieldName)) {
+      if (isUNLModify && fieldName.equals("Account")) {
         continue;
       }
 
--- a/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
+++ b/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
@@ -60,9 +60,24 @@ public class STObjectType extends SerializedType<STObjectType> {
   public STObjectType fromJson(JsonNode node) {
     UnsignedByteArray byteList = UnsignedByteArray.empty();
     BinarySerializer serializer = new BinarySerializer(byteList);
+    boolean isUNLModify;
+    try {
+      isUNLModify = "UNLModify".equals(node.get("TransactionType").asText());
+    } catch (Exception e) {
+      isUNLModify = false;
+    }
+
 
     List<FieldWithValue<JsonNode>> fields = new ArrayList<>();
     for (String fieldName : Lists.newArrayList(node.fieldNames())) {
+
+      /**
+       * The Account field must not be a part of the UNLModify pseudotransaction encoding, due to a bug in rippled.
+       */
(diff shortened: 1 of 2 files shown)


xrpl4j-binary-codec/src/test/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodecTest.java
```diff
@@ -120,6 +120,23 @@ void encodeDecodeAmendments() throws JsonProcessingException {
     assertThat(encoder.decode(hex)).isEqualTo(json);
   }
 
+  @Test
+  void encodeUnlModify() throws JsonProcessingException {
+    String json = "{" +
```
sappenin, line 125: Can we add another test like this one that validates the `isUNLModify` having the opposite value so we cover both cases?

xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
```diff
@@ -60,9 +60,24 @@ public STObjectType fromParser(BinaryParser parser) {
...
+    }
+
 
     List<FieldWithValue<JsonNode>> fields = new ArrayList<>();
     for (String fieldName : Lists.newArrayList(node.fieldNames())) {
+
+      /**
+       * The Account field must not be a part of the UNLModify pseudotransaction encoding, due to a bug in rippled.
+       */
+      if (isUNLModify && "Account".equals(fieldName)) {
```
sappenin, line 77: Can we add a unit test for the case where `fieldName` is not `"Account"` but `isUNLModify` is true?

sappenin (changes requested):

xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
mukulljangid: This is actually happening for all other fields of the UNLModify transaction test that pass through here and get encodeed.

xrpl4j-binary-codec/src/test/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodecTest.java
mukulljangid: This is covered by tests like `encodeForSigning` and `encodeForMultiSigning` where the txType is Payment and isUNLModify is false.

xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
nkramer44, line 77: Small nit

```suggestion
      if (isUNLModify && fieldName.equals("Account")) {
```

nkramer44 (changes requested):

nkramer44 (approved):

xrpl4j-binary-codec/src/test/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodecTest.java
sappenin: I don't see `UNLModifyDisabling` in those other tests. Is that correct?

mukulljangid: oh, yes. I just added that! (Sorry, I didn't understand the previous comment correctly.)

sappenin (approved):

Issue #153, opened 2021-09-23
Add new isFinal() method to detect if a transaction is final on ledger


nkramer44:
We should add a method to `XrplClient` called `isFinal(Hash256 transactionHash)` that reaches out to rippled and determines if the transaction is in a final state on ledger.

This method should follow the logic found here: https://xrpl.org/reliable-transaction-submission.html#verification

PR #230, opened 2022-03-04
Add isFinal() to xrplClient

mukulljangid:
Address issue https://github.com/XRPLF/xrpl4j/issues/153

Also adds a basic test.


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +238,77 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+      .addParams(TransactionRequestParams.of(transactionHash))
+      .build();
+
+    TransactionResult response = jsonRpcClient.send(request, TransactionResult.class);
+    UnsignedInteger lastLedgerSequence = response.transaction().lastLedgerSequence()
+      .orElseThrow(() -> new IllegalStateException("Ledger result did not contain LastLedgerSequence"));
+
+    // condition 1: validated = true
+    if (response.validated()) {
+      if (response.status().get().equals("success")) {
```
nkramer44, line 274: I think you want to look at the transaction metadata `transactionResult` field. `success` on the `response.status` just means rippled returned a successful response, not that the transaction was a success.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +238,77 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+        return true;
+      } else {
+        return false;
+      }
+    } else if (FluentCompareTo.is(getMostRecentlyValidatedLedgerIndex()).lessThan(lastLedgerSequence)) {
+      // condition 2: last ledger seq hasn't passed yet
+      Thread.sleep(4000);
+      return isFinal(transactionHash);
```
nkramer44, line 282: I dont think we want to loop/recurse here. We should just return `false` and let the caller decide how they want to loop.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +238,77 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    } else {
+      // condition 3: last ledger seq has passed.
+
+      LedgerIndex submittedOn = response.ledgerIndexSafe();
```
nkramer44, line 286: This needs to be passed in to the method. `response.ledgerIndexSafe()` will return the ledger index of the data that rippled returned. So you could submit the transaction at ledger index 10, but you make this request at ledger index 12, `response.ledgerIndexSafe()` will return `12`, but we really want to know when the transaction was submitted, which should be `10`. 

Callers of this method will need to store this value somewhere when they submit the transaction and pass it in to this method.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +238,77 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+      Range<UnsignedLong> submittedToLast = Range.closed(UnsignedLong.valueOf(submittedOn.toString()),
+        UnsignedLong.valueOf(lastLedgerSequence.toString()));
+      boolean ledgerGapExists = this.serverInfo().completeLedgerRanges().stream()
+        .noneMatch(range -> range.encloses(submittedToLast));
+
+      if (ledgerGapExists) {
+        // wait to acquire the missing ledgers
+        Thread.sleep(4000);
+        return isFinal(transactionHash);
```
nkramer44, line 295: Same here, just return `false`

nkramer44 (changes requested):

.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
nkramer44, line 1: Can you add this file to `.gitignore` so it doesn't get committed?

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,80 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+   * @throws JsonRpcClientErrorException if {@code jsonRpcClient} throws an error.
+   */
+  public boolean isFinal(Hash256 transactionHash, LedgerIndex submittedOnLedgerIndex)
+    throws JsonRpcClientErrorException {
+    JsonRpcRequest request = JsonRpcRequest.builder()
+      .method(XrplMethods.TX)
+      .addParams(TransactionRequestParams.of(transactionHash))
+      .build();
+
+    TransactionResult response = jsonRpcClient.send(request, TransactionResult.class);
```
nkramer44, line 271: Could we just call `this.transaction(TransactionRequestPArams.of(transactionHash), Transaction.class)` to avoid duplicating code?

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,80 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    UnsignedInteger lastLedgerSequence = response.transaction().lastLedgerSequence()
```
nkramer44, line 272: What happens here if the transaction has already expired (ie lastLedgerSequence has already passed)? I believe your call to `this.transaction()` will throw an exception because you won't be able to find the transaction on ledger. 

This ends up being sort of paradoxical -- You can't check to see if a transaction is expired without checking the `lastLedgerIndex`, but you can't get `lastLedgerIndex` without getting the transaction from the ledger, but if the condition you are looking for is true, you will never get the transaction from ledger, and therefore can never get `lastLedgerSequence`.

The solution to this is that users have to record the `lastLedgerSequence` that they set on the transaction _before_ submitting it, and you need to allow them to pass that `lastLedgerSequence` into this method.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,80 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+      } else {
+        return false;
+      }
+    } else if (FluentCompareTo.is(getMostRecentlyValidatedLedgerIndex()).lessThan(lastLedgerSequence)) {
+      // condition 2: last ledger seq hasn't passed yet
+      // possible solution would be to check again after a while to get the updated validation status.
+      return false;
+    } else {
+      // condition 3: last ledger seq has passed.
+      Range<UnsignedLong> submittedToLast = Range.closed(UnsignedLong.valueOf(submittedOnLedgerIndex.toString()),
```
nkramer44, line 289: Why not make this a `Range<UnsignedInteger>` so you don't have to do all this `UnsignedInteger` -> `UnsignedLong` conversion?

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,80 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+  /**
+   * Check if the transaction is final on the ledger or not.
+   *
+   * @param transactionHash {@link Hash256} of the submitted transaction to check the status for.
+   * @param submittedOnLedgerIndex {@link LedgerIndex} on which the transaction with hash transactionHash
+   *        was submitted. This can be obtained from submit() response of the tx as validatedLedgerIndex.
+   * @return {@code true} if the {@link Transaction} is final/validated else {@code false}.
+   * @throws JsonRpcClientErrorException if {@code jsonRpcClient} throws an error.
+   */
+  public boolean isFinal(Hash256 transactionHash, LedgerIndex submittedOnLedgerIndex)
```
nkramer44, line 264: What do you think about changing the return type of this method from `boolean` to some `enum` called `Finality` or `FinalStatus` or something? We could have one enum value for each scenario, so something like

```java
public enum FinalStatus {

  NOT_FINAL,
  VALIDATED_SUCCESS,
  VALIDATED_FAILURE,
  EXPIRED

}
```

Otherwise, user's won't really get any good indicator of what happened and it will make it much harder to react to the outcome.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,80 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    UnsignedInteger lastLedgerSequence = response.transaction().lastLedgerSequence()
+      .orElseThrow(() -> new IllegalStateException("Ledger result did not contain LastLedgerSequence"));
+
+    // condition 1: validated = true
+    if (response.validated()) {
+      String txResult = ((TransactionMetadata) response.metadata().get()).transactionResult();
+      if (txResult.equals("tesSUCCESS")) {
+        return true;
+      } else {
+        return false;
```
nkramer44, line 281: If a transaction has been validated, it is final, even if it has a `tec` code instead of `tesSUCCESS`, so this should return `true`. However, if you take my `FinalStatus` suggestion above, you would return `FinalStatus.VALIDATED_FAILURE`.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
nkramer44, line 264: This method should also be unit tested somewhere.

nkramer44 (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
mukulljangid: Because, completeLedgerRanges is coming in as Range<UnsignedLong>. So, I have to convert either one of those.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,102 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+   * @return {@code true} if the {@link Transaction} is final/validated else {@code false}.
+   * @throws JsonRpcClientErrorException if {@code jsonRpcClient} throws an error.
+   */
+  public FinalStatus isFinal(
+    Hash256 transactionHash,
+    LedgerIndex submittedOnLedgerIndex,
+    UnsignedInteger lastLedgerSequence
+  ) throws JsonRpcClientErrorException {
+
+    TransactionResult response = this.transaction(TransactionRequestParams.of(transactionHash), Transaction.class);
```
nkramer44, line 295: If the transaction doesn't exist on ledger, this line will throw a `JsonRpcClientErrorException`. However, we need to know if the transaction isn't found so we can hit line 311.

I'd suggest adding a method like this:
```java
protected Optional<? extends TransactionResult<? extends Transaction>> getValidatedTransaction(
  final Hash256 transactionHash
) {
  Objects.requireNonNull(transactionHash);

  try {
    final TransactionResult<? extends Transaction> transactionResult = xrplClient.transaction(
      TransactionRequestParams.of(transactionHash), Transaction.class
    );
    return Optional.ofNullable(transactionResult).filter(TransactionResult::validated);
  } catch (JsonRpcClientErrorException e) {
    // The transaction was not found on ledger, warn on this, but otherwise return.
    logger.warn(e.getMessage(), e);
    return Optional.empty();
  }
}
```

Then you can call that method and then do `.map(...).orElseGet(() -> ...)` instead of your `if`/`else`s that you have here.

nkramer44 (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -232,6 +239,140 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+   */
+  public FinalStatus isFinal(
+    Hash256 transactionHash,
+    LedgerIndex submittedOnLedgerIndex,
+    UnsignedInteger lastLedgerSequence,
+    UnsignedInteger sequence,
+    Address account
+  ) {
+
+    return getValidatedTransaction(transactionHash)
```
sappenin, line 325: We need pretty extensive unit test coverage here. See Rsubs testing maybe?

sappenin (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
mukulljangid: Yes. I'll start writing those.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +259,140 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+              // there is a possibility that the transaction with hash transactionHash exists on the missing ledgers
+              // possible solution would be to wait and query the missing ledgers once acquired, resubmit with
+              // updated fee and lastLedgerSequence if not found.
+              return FinalStatus.NOT_FINAL;
+            } else {
+              AccountInfoResult accountInfoResult = this.accountInfo(
+                AccountInfoRequestParams.of(account)
+              );
+              UnsignedInteger accountSequence = accountInfoResult.accountData().sequence();
+              if (FluentCompareTo.is(sequence).lessThan(accountSequence)) {
```
sappenin, line 381: I'm confused by this check - it seems like regardless of the current _account_ sequence, if the transaction doesn't exist on a ledger less than `lastLedgerSequence` (checked above on line 359) then it shouldn't matter what the account sequence is right?

If I'm wrong here, then we need to change RSubs, but the comment in RSubs is just this: 

```
// If we get here, it means the transaction was not found in the range of acceptable validated
// ledgers and never will be found, so expire the ReliableSubmission.
```

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/IsFinalIT.java
```diff
@@ -0,0 +1,73 @@
...
+public class IsFinalIT extends AbstractIT {
+
+  Wallet wallet = createRandomAccount();
+
+  @Test
+  public void simpleIsFinalTest() throws JsonRpcClientErrorException, InterruptedException {
+
+    ///////////////////////
+    // Get validated account info and validate account state
+    AccountInfoResult accountInfo = this.scanForResult(() -> this.getValidatedAccountInfo(wallet.classicAddress()));
```
sappenin, line 28: I could be wrong about this, but I think there are more ITs in RSubs that we should consider capturing. If not, we should probably try to test a few more cases in our ITs, e.g., the other states if we keep them.

sappenin (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
nkramer44: This actually needs to be changed in RSubs I believe. If you look at [this flowchart](https://xrpl.org/img/reliable-tx-submission.svg), this check corresponds to the box that says (next to the yellow note):
```
Unknown transaction
has been validated.
Manual intervention
recommended
```

I'm not sure why we didn't implement this in RSubs, but I actually think we should wait until this gets merged and we should just delegate to this method in RSubs, instead of making a fix in the current RSubs method. This check is a super edge case and shouldn't really happen.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +259,140 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+  /**
+   * Enumeration of the types of responses from isFinal function.
+   */
+  public enum FinalStatus {
+    /**
+     * Transaction has not been validated until this moment, which means it could be validated in the
+     * following closed ledgers, or, the XRPL peer you are connected to could be missing the ledgers
+     * it would have been validated on.
+     */
+    NOT_FINAL,
```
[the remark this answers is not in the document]
sappenin: After discussing further, these states are sufficient.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +259,140 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    LedgerIndex submittedOnLedgerIndex,
+    UnsignedInteger lastLedgerSequence,
+    UnsignedInteger sequence,
+    Address account
+  ) {
+
+    return getValidatedTransaction(transactionHash)
+      .map(transactionResult -> {
+        //Do some stuff with transactionResult and return a FinalityStatus
+        String resultCode = transactionResult.metadata()
```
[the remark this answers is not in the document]
sappenin: Let's introduce a new `Finality` object so that we can capture the resultCode when making this call:

```

  public interface Finality {
    FinalityStatus finalityStatus();
    
    Optiona<String> resultCode();

    @Default
    default String resultCodeSafe(){
      return resultCode().orElseThrow(...);
    }
  }
```

xrpl4j-client/pom.xml
```diff
@@ -69,6 +69,23 @@
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
sappenin, line 74: Let's add the `test` scope here, just to be safe.

```suggestion
      <artifactId>assertj-core</artifactId>
      <scope>test</scope>
```

xrpl4j-client/pom.xml
```diff
@@ -69,6 +69,23 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
```
sappenin, line 78: Let's add the `test` scope here, just to be safe.

```suggestion
      <artifactId>mockito-core</artifactId>
      <scope>test</scope>
```

xrpl4j-client/pom.xml
```diff
@@ -69,6 +69,23 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
```
sappenin, line 82: Let's add the `test` scope here, just to be safe.

```suggestion
       <artifactId>junit-jupiter-api</artifactId>
      <scope>test</scope>
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,31 @@
...
+   */
+  static ImmutableFinality.Builder builder() {
+    return ImmutableFinality.builder();
+  }
+
+  XrplClient.FinalityStatus finalityStatus();
+
+  Optional<String> resultCode();
+
+  @Value.Auxiliary
```
sappenin, line 27: Any particular reason this is marked `@Auxilary`?

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Construct a builder for this class.
+   *
+   * @return An {@link ImmutableFinality.Builder}.
+   */
+  static ImmutableFinality.Builder builder() {
+    return ImmutableFinality.builder();
+  }
+
+  XrplClient.FinalityStatus finalityStatus();
```
sappenin, line 23: TODO: Javadoc

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,31 @@
+package org.xrpl.xrpl4j.client;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+import java.util.Optional;
+
+@Value.Immutable
```
sappenin, line 9:
```suggestion
/**
* Defines how an XRP Ledger finality decision is represented; includes both a finality status and an engine code.
*/
@Value.Immutable
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,31 @@
...
+  static ImmutableFinality.Builder builder() {
+    return ImmutableFinality.builder();
+  }
+
+  XrplClient.FinalityStatus finalityStatus();
+
+  Optional<String> resultCode();
+
+  @Value.Auxiliary
+  default String resultCodeSafe() {
```
sappenin, line 28: TODO: Javadoc

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,31 @@
...
+   * Construct a builder for this class.
+   *
+   * @return An {@link ImmutableFinality.Builder}.
+   */
+  static ImmutableFinality.Builder builder() {
+    return ImmutableFinality.builder();
+  }
+
+  XrplClient.FinalityStatus finalityStatus();
+
```
sappenin, line 24:
```suggestion
/**
* The rippled server summarizes transaction results with result codes, which appear in fields such as engine_result * and meta.TransactionResult. These codes are grouped into several categories of with different prefixes.
* 
* @see "https://xrpl.org/transaction-results.html#transaction-results"
*/
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    Objects.requireNonNull(transactionHash);
+
+    try {
+      final TransactionResult<? extends Transaction> transactionResult = this.transaction(
+        TransactionRequestParams.of(transactionHash), Transaction.class
+      );
+      return Optional.ofNullable(transactionResult).filter(TransactionResult::validated);
+    } catch (JsonRpcClientErrorException e) {
+      // The transaction was not found on ledger, warn on this, but otherwise return.
+      LOGGER.warn(e.getMessage(), e);
```
sappenin, line 294: This portion doesn't seem to be covered in unit tests. We can force coverage here by mocking `jsonRpcClient` to throw an exception of type `JsonRpcClientErrorException` and then asserting an `Optional.empty` response.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+      return Optional.empty();
+    }
+  }
+
+  /**
+   * Check if there missing ledgers in rippled in the given range.
```
sappenin, line 300: Your formatter is still quite off from the rest of the code base. E.g., this javadoc should look like this instead (formatting only):

```
 /**
   * Check if there missing ledgers in rippled in the given range.
   *
   * @param submittedLedgerSequence {@link LedgerIndex} at which the {@link Transaction} was submitted on.
   * @param lastLedgerSequence      he ledger index/sequence of type {@link UnsignedInteger} after which the transaction
   *                                will expire and won't be applied to the ledger.
   *
   * @return {@link Boolean} to indicate if there are gaps in the ledger range.
   */
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+   * @param submittedLedgerSequence {@link LedgerIndex} at which the {@link Transaction} was submitted on.
+   * @param lastLedgerSequence he ledger index/sequence of type {@link UnsignedInteger} after which the
+   *        transaction will expire and won't be applied to the ledger.
+   * @return {@link Boolean} to indicate if there are gaps in the ledger range.
+   */
+  protected boolean ledgerGapsExistBetween(UnsignedLong submittedLedgerSequence, UnsignedLong lastLedgerSequence) {
```
sappenin, line 306:
```suggestion
  protected boolean ledgerGapsExistBetween(
  final UnsignedLong submittedLedgerSequence, 
  final UnsignedLong lastLedgerSequence
  ) {
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+    final ServerInfo serverInfo;
+    try {
+      serverInfo = this.serverInfo();
+    } catch (JsonRpcClientErrorException e) {
+      LOGGER.error(e.getMessage(), e);
+      return true; // Assume ledger gaps exist so this can be retried.
```
sappenin, line 312: TODO: Unit test coverage for this scenario.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
sappenin, line 306: Let's add unit tests for this method directly (i.e., `ledgerGapsExistBetween`). We should test a variety of inputs here (even nulls) against a variety of `serverInfo` responses (likely means you'll need to mock `jsonRpcClient`).

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+          final boolean isTransactionExpired = FluentCompareTo.is(getMostRecentlyValidatedLedgerIndex())
+            .greaterThan(lastLedgerSequence);
+          if (!isTransactionExpired) {
+            LOGGER.debug("Transaction with hash: {} has not expired yet, check again", transactionHash);
+            return Finality.builder().finalityStatus(FinalityStatus.NOT_FINAL).build();
+          } else {
+            boolean isMissingLedgers = ledgerGapsExistBetween(UnsignedLong.valueOf(submittedOnLedgerIndex.toString()),
+              UnsignedLong.valueOf(lastLedgerSequence.toString()));
+            if (isMissingLedgers) {
+              LOGGER.debug("Transaction with hash: {} has expired and rippled is missing some to confirm if it" +
```
sappenin, line 411: This scenario is not covered by the unit tests covering this method (i.e., `isFinal`)

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,180 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+              } else {
+                LOGGER.debug("Transaction with hash: {} has expired, consider resubmitting with updated" +
+                  " lastledgersequence and fee", transactionHash);
+                return Finality.builder().finalityStatus(FinalityStatus.EXPIRED).build();
+              }
+            }
+          }
+        } catch (JsonRpcClientErrorException e) {
+          LOGGER.warn(e.getMessage(), e);
+          return Finality.builder().finalityStatus(FinalityStatus.NOT_FINAL).build();
```
sappenin, line 432: This scenario is not covered by the unit tests covering this method (i.e., `isFinal`)

sappenin (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
mukulljangid: It cannot be assigned a default value and auxiliary will remove the attribute from equals comparison. But now that I run without any annotation, I feel like there is no need for one. I removed it!

sappenin: Oh, my bad -- I didn't realize this was a `...Safe` variant, so you were correct to mark it as `Auxiliary`. The _actual_ value for equals/etc will be taken from `resultCode()`.

Please add this annotation back -- sorry for leading you in the wrong direction.

sappenin: I think `resultCode` is the proper name. If you look [here](https://xrpl.org/transaction-results.html#transaction-results), there are two types of result: (1) and `engine_result` and (2) a transaction result defined in metadata. To me calling it `resultCode` seems like it captures both meanings.

That said, @nkramer44 WDYT? Should we call this `engineResult` instead?

nkramer44: I wouldn't call it `engineResult` personally. That is the field you get back in a `SubmitResult`. While the values of that field are the same ones you see in the `TransactionMetadata` field, I think it's clearer to call it `resultCode`, since this is what you see in `TransactionMetadata`

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
nkramer44, line 23: Small nit, but can we move `XrplClient.FinalityStatus` into it's own enum, or maybe move it into a sub-class of `Finality`?

nkramer44 (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
mukulljangid: Okay!

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,52 @@
...
+  Optional<String> resultCode();
+
+  /**
+   * Text result code indicating the preliminary result of the transaction.
+   *
+   * @return The #result() value for the particular transaction.
+   */
+  @Value.Auxiliary
+  default String resultCodeSafe() {
+    return resultCode().orElseThrow(() -> new IllegalStateException("Finality does not contain resultCode."));
```
nkramer44, line 50: This doesn't seem to be unit tested. Can you add a `FinalityTest`?

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,160 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+          final boolean isTransactionExpired = FluentCompareTo.is(getMostRecentlyValidatedLedgerIndex())
+            .greaterThan(lastLedgerSequence);
+          System.out.println(isTransactionExpired);
+          if (!isTransactionExpired) {
+            LOGGER.debug("Transaction with hash: {} has not expired yet, check again", transactionHash);
+            return Finality.builder().finalityStatus(FinalityStatus.NOT_FINAL).build();
+          } else {
+            boolean isMissingLedgers = ledgerGapsExistBetween(UnsignedLong.valueOf(submittedOnLedgerIndex.toString()),
+              UnsignedLong.valueOf(lastLedgerSequence.toString()));
+            System.out.println("isMissingLedgers" + isMissingLedgers);
```
nkramer44, line 389: Let's replace this with a `logger.debug`

nkramer44 (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
mukulljangid: I don't that should be there. Was just a debug statement. I'll remove it!

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/FinalityStatus.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Transaction is final/validated on the ledger successfully.
+   */
+  VALIDATED_SUCCESS,
+  /**
+   * Transaction is final/validated on the ledger with some failure.
+   */
+  VALIDATED_FAILURE,
+  /**
+   * In the past, the XRP Ledger supported transactions without metadata. While uncommon, this is a
```
sappenin, line 22:
```suggestion
   * The transaction is validated, but it's unclear if the transaction was successful or not. For example, in the past, the XRP Ledger supported transactions without metadata. While uncommon, this is a
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/FinalityStatus.java
```diff
@@ -0,0 +1,31 @@
...
+   * potential state for older transactions.
```
sappenin, line 23:
```suggestion
   * potential state for older transactions that prevents this library from determining a success/failure state.
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/Finality.java
```diff
@@ -0,0 +1,52 @@
+package org.xrpl.xrpl4j.client;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
```
sappenin, line 3: Let's move this class into the `xrpl4j-model` module.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/FinalityStatus.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Enumeration of the types of responses from isFinal function.
+ */
+public enum FinalityStatus {
+  /**
+   * Transaction has not been validated until this moment, which means it could be validated in the
+   * following closed ledgers, or, the XRPL peer you are connected to could be missing the ledgers
+   * it would have been validated on.
+   */
+  NOT_FINAL,
```
sappenin, line 12: Let's move this into the `xrpl4j-model` module.

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,158 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+   *        from the same account.
+   * @param account The unique {@link Address} of the account that initiated this transaction.
+   *
+   * @return {@code true} if the {@link Transaction} is final/validated else {@code false}.
+   */
+  public Finality isFinal(
+    Hash256 transactionHash,
+    LedgerIndex submittedOnLedgerIndex,
+    UnsignedInteger lastLedgerSequence,
+    UnsignedInteger sequence,
```
sappenin, line 346:
```suggestion
    UnsignedInteger transactionAccountSequence,
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,158 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+              return Finality.builder().finalityStatus(FinalityStatus.NOT_FINAL).build();
+            } else {
+              AccountInfoResult accountInfoResult = this.accountInfo(
+                AccountInfoRequestParams.of(account)
+              );
+              UnsignedInteger accountSequence = accountInfoResult.accountData().sequence();
+              if (FluentCompareTo.is(sequence).lessThan(accountSequence)) {
+                // a different transaction with this sequence has a final outcome.
+                // this represents an unexpected case
+                throw new IllegalStateException("Something unexpected happened. Tx not final.");
```
sappenin, line 400: Should return `EXPIRED_WITH_SPENT_ACCOUNT_SEQUENCE`

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/FinalityStatus.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * In the past, the XRP Ledger supported transactions without metadata. While uncommon, this is a
+   * potential state for older transactions.
+   */
+  VALIDATED_UNKNOWN,
+  /**
+   * The lastLedgerSequence of the tx has passed, i.e., transaction was not included in any closed
+   * ledgers ranging from the one it was submitted on to the lastLedgerSequence.
+   */
+  EXPIRED
```
sappenin, line 30:
```suggestion
  EXPIRED,
  
    /**
   * An unknown transaction has been validated ahead of or instead of this transaction, so manual intervention and
   * investigation is recommended.
   *
   * Some reasons an unknown transaction may have been validated instead of this one: The transaction was malleable and
   * succeeded with a different hash; A different program or person with your secret key is also sending transactions
   * from the same account; You previously sent a transaction, but lost your record of it.
   */
  EXPIRED_WITH_SPENT_ACCOUNT_SEQUENCE
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -252,6 +260,158 @@ public FeeResult fee() throws JsonRpcClientErrorException {
...
+
+  /**
+   * Check if the transaction is final on the ledger or not.
+   *
+   * @param transactionHash {@link Hash256} of the submitted transaction to check the status for.
+   * @param submittedOnLedgerIndex {@link LedgerIndex} on which the transaction with hash transactionHash
+   *        was submitted. This can be obtained from submit() response of the tx as validatedLedgerIndex.
+   * @param lastLedgerSequence The ledger index/sequence of type {@link UnsignedInteger} after which the
+   *        transaction will expire and won't be applied to the ledger.
+   * @param sequence The sequence number of the account submitting the {@link Transaction}. A {@link Transaction}
```
sappenin, line 335:
```suggestion
   * @param transactionAccountSequence The sequence number of the account submitting the {@link Transaction}. A {@link Transaction}
```

xrpl4j-client/src/test/java/org/xrpl/xrpl4j/client/FinalityTest.java
```diff
@@ -0,0 +1,33 @@
...
+    Finality finality = Finality.builder()
+      .finalityStatus(FinalityStatus.NOT_FINAL)
+      .resultCode("tesSUCCESS")
+      .build();
+
+    assertThat(finality.finalityStatus()).isEqualTo(FinalityStatus.NOT_FINAL);
+    assertThat(finality.resultCode().get()).isEqualTo("tesSUCCESS");
+    assertThat(finality.resultCodeSafe()).isEqualTo("tesSUCCESS");
+  }
+
```
sappenin, line 22: Add new tests for `EXPIRED_WITH_SPENT_ACCOUNT_SEQUENCE`

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/IsFinalIT.java
```diff
@@ -0,0 +1,152 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.common.primitives.UnsignedInteger;
+import com.google.common.primitives.UnsignedLong;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.xrpl.xrpl4j.client.FinalityStatus;
+import org.xrpl.xrpl4j.client.JsonRpcClientErrorException;
+import org.xrpl.xrpl4j.client.XrplClient;
```
sappenin, line 11: Let's add coverage for the following methods:

1. submit()
2. addSignature()

xrpl4j-client/src/test/java/org/xrpl/xrpl4j/client/FinalityTest.java
```diff
@@ -0,0 +1,35 @@
+package org.xrpl.xrpl4j.client;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+import org.xrpl.xrpl4j.model.client.Finality;
+import org.xrpl.xrpl4j.model.client.FinalityStatus;
+
+public class FinalityTest {
```
sappenin, line 10: This test needs to be moved into the `xrpl4j-model` module.

sappenin, line 10:
```suggestion
/**
* Unit tests for {@link Finality}.
*/
public class FinalityTest {
```

xrpl4j-client/src/test/java/org/xrpl/xrpl4j/client/XrplClientTest.java
```diff
@@ -0,0 +1,1256 @@
...
+      .account(Address.of("rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn"))
+      .fee(XrpCurrencyAmount.ofDrops(UnsignedLong.ONE))
+      .sequence(UnsignedInteger.ONE)
+      .ticketCount(UnsignedInteger.ONE)
+      .build();
+    assertThat(xrplClient.addSignature(ticketCreate, "sign").transactionSignature().get()).isEqualTo("sign");
+  }
+
+  @Test
+  public void submitWithSignedTx() {
```
sappenin, line 842: This test looks good, but we're still missing coverage for the other variant, with this signature:

```
public <T extends Transaction> SubmitResult<T> submit(
    final org.xrpl.xrpl4j.crypto.signing.SignedTransaction signedTransaction
  )
```

xrpl4j-client/src/test/java/org/xrpl/xrpl4j/client/XrplClientTest.java
```diff
@@ -0,0 +1,1256 @@
...
+        UnsignedInteger.ONE,
+        Address.of("rDgZZ3wyprx4ZqrGQUkquE9Fs2Xs8XBcdw")
+      ).finalityStatus()
+    ).isEqualTo(FinalityStatus.EXPIRED);
+
+    assertThat(calledWithHash.get()).isEqualTo(transactionHash);
+  }
+
+  @Test
+  void isFinalForNoLedgerGapAndLastLedgerSequencePassedTransaction_ReturnsExpired() {
```
sappenin, line 501: This test is slightly mis-named (it's not returning `EXPIRED`). How about this:

```suggestion
  void isFinalForNoLedgerGapAndLastLedgerSequencePassed_ReturnsExpiredWithSpentSequence() {
```

sappenin (changes requested):
This is really close, very nice work @mukulljangid -- just a few minor nits and then I think we're good to go!

mukulljangid:
The above change is for AccountTransactionsIT

sappenin (approved):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AccountTransactionsIT.java
```diff
@@ -39,20 +39,20 @@
 import org.xrpl.xrpl4j.model.client.ledger.LedgerResult;
 import org.xrpl.xrpl4j.model.transactions.Address;
 import org.xrpl.xrpl4j.model.transactions.Hash256;
-import org.xrpl.xrpl4j.tests.environment.MainnetEnvironment;
+import org.xrpl.xrpl4j.tests.environment.TestnetEnvironment;
```
sappenin, line 42: @mukulljangid What instigated this change?

Unless there's a really good reason to prefer Testnet, I propose that we revert this change because if Testnet ever gets reset we'll need to go in and adjust this test (and I can't think of any reason _not_ to hit `Mainnet`.

WDYT?

Also, @nkramer44 WDYT?

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AccountTransactionsIT.java
mukulljangid: I didn't think of the reset possibility. Mainnet requests were usually pretty slow and led tests to fail.

sappenin: Interesting -- I would expect the move to clio in s1/s2 to make Mainnet requests much more responsive, FWIW.

nkramer44 (approved):

sappenin (approved):

PR #232, opened 2022-03-08
Add `AccountSetTransactionFlags` so that we can set multiple AccountRoot flags in one transaction

nkramer44:
Adds `AccountSetTransactionFlags` corresponding to the `tf` flags for the `AccountSet` transaction. The docs say that using these flags is discouraged, but we should give developers the option to use them, as using them can save them from having to submit multiple `AccountSet` transactions to set multiple flags.

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
@@ -1,6 +1,7 @@
 package org.xrpl.xrpl4j.model.flags;
 
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.google.common.base.Preconditions;
 import org.xrpl.xrpl4j.model.ledger.PayChannelObject;
 import org.xrpl.xrpl4j.model.transactions.AccountSet;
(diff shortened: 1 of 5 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
```diff
@@ -535,6 +536,289 @@ public boolean lsfOneOwnerCount() {
...
+     */
+    public boolean tfAllowXrp() {
+      return this.isSet(ALLOW_XRP);
+    }
+
+    /**
+     * A builder class for {@link AccountSetTransactionFlags}.
+     */
+    public static class Builder {
+      private boolean tfFullyCanonicalSig = true;
```
sappenin, line 723: Any reason for `private`, but only here and not the other fields?

sappenin (approved):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/flags/Flags.java
nkramer44: copy paste error :shame:

PR #234, opened 2022-03-10
Add pseudo-transactions

mukulljangid:
Add pseudo-transactions:

- SetFee
- EnableAmendment
- UNLModify

Add tests.

### DIFF
--- a/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
+++ b/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
@@ -38,6 +38,11 @@ public enum TransactionType {
   DEPOSIT_PRE_AUTH("DepositPreauth"),
 
   /**
+   * The {@link TransactionType} for the {@link EnableAmendment} transaction.
+   */
+  ENABLE_AMENDMENT("EnableAmendment"),
+
+  /**
    * The {@link TransactionType} for the {@link EscrowCancel} transaction.
    */
   ESCROW_CANCEL("EscrowCancel"),
@@ -83,6 +88,11 @@ public enum TransactionType {
   PAYMENT_CHANNEL_FUND("PaymentChannelFund"),
 
   /**
+   * The {@link TransactionType} for the {@link SetFee} transaction.
+   */
+  SET_FEE("SetFee"),
+
+  /**
    * The {@link TransactionType} for the {@link SetRegularKey} transaction.
    */
   SET_REGULAR_KEY("SetRegularKey"),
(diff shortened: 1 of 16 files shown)


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
```diff
@@ -37,6 +37,11 @@
    */
   DEPOSIT_PRE_AUTH("DepositPreauth"),
 
+  /**
+   * The {@link TransactionType} for the {@link EnableAmendment} transaction.
+   */
+  ENABLE_AMENDMENT("EnableAmendment"),
```
sappenin, line 43: Can you add Unit tests that assert the capitalization and contents of each value in this class (i.e., in `TransactionType`)?

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/EnableAmendmentTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.json.JSONException;
+import org.junit.jupiter.api.Test;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+
+import java.util.Optional;
+
+public class EnableAmendmentTest {
+
+  @Test
+  public void testJson() throws JsonProcessingException, JSONException {
```
sappenin, line 17: This test doesn't appear to be testing any JSON.  Let's rename the test to something clearer, like this:

```suggestion
  public void testBuilder() throws JsonProcessingException, JSONException {
```

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/transactions/SetFeeTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.json.JSONException;
+import org.junit.jupiter.api.Test;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+
+import java.util.Optional;
+
+public class SetFeeTest {
+
+  @Test
+  public void testJson() throws JsonProcessingException, JSONException {
```
sappenin, line 17: This test doesn't appear to be testing any JSON.  Let's rename the test to something clearer, like this:

```suggestion
  public void testBuilder() throws JsonProcessingException, JSONException {
```

sappenin (changes requested):
All looks good, just one request around testing.

nkramer44 (approved):
This looks great. Thanks for thinking through this one and nice job figuring out the serializer stuff!

mukulljangid:
@sappenin @nkramer44 Also, this would help us in future, when we introduce FeeSettings ledger object. FeeSettings also has the field, BaseFee, which is also expected to be in hex, so this serializer change will help us there in future.

sappenin (approved):
LGTM!

Issue #248, opened 2022-03-22
Add more unit test coverage for UnsignedByteArray


sappenin:
We need to add unit test coverage to the following methods in `UnsignedByteArray`:

1. `empty()`
2. `ofSize(..)`
3. `fill()`
4. `length()`
5. `get(..)`
6. `append(UnsignedByte)`
7.  `append(UnsignedByteArray)`
8. `set()`
9. `slice()`
10. `hashCode()`
11. `toString()`

PR #249, opened 2022-03-23
add tests for all the methods in class UnsignedByteArray

mukulljangid:
Addresses: https://github.com/XRPLF/xrpl4j/issues/248

### DIFF
--- a/xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
+++ b/xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
@@ -56,7 +56,7 @@ public class UnsignedByteArrayTest {
   public void simpleLength() {
     final int size = 2;
     assertThat(UnsignedByteArray.of(new byte[] {0, MAX_BYTE}).length()).isEqualTo(size);
-    assertThat(UnsignedByteArray.fill(size).size()).isEqualTo(UnsignedByteArray.ofSize(size).length());
+    assertThat(UnsignedByteArray.of(new byte[] {0, 1}).length()).isEqualTo(UnsignedByteArray.ofSize(size).length());
     assertThat(UnsignedByteArray.ofSize(size).length()).isEqualTo(size);
   }
 
--- a/xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
+++ b/xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
@@ -162,7 +162,7 @@ public class UnsignedByteArrayTest {
   }
 
   @Test
-  public void UnsignedByteArrayToString() {
+  public void unsignedByteArrayToString() {
     UnsignedByteArray array1 = UnsignedByteArray.of(new byte[] {0, 1});
     UnsignedByteArray array2 = UnsignedByteArray.of(new byte[] {8, 9});
     assertThat(array1.toString()).isEqualTo(array1.toString());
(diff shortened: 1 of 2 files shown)


xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
```diff
@@ -41,4 +42,96 @@ public void lowerCaseOrUpperCase() {
     assertThat(UnsignedByteArray.fromHex("abcdef0123").hexValue()).isEqualTo("ABCDEF0123");
   }
 
+  @Test
+  public void simpleEmpty() {
+    assertThat(UnsignedByteArray.of(new byte[] {0}).empty()).isEqualTo(UnsignedByteArray.of(new byte[] {}));
+    assertThat(UnsignedByteArray.of(new byte[] {0, MAX_BYTE}).empty()).isEqualTo(UnsignedByteArray.of(new byte[] {}));
```
nkramer44, line 48: Why do we need these? `empty()` is a static method, so it's a bit weird to call `empty()` on an instance of `UnsignedByteArray`. Also, `empty()` just creates a new empty `UnsignedByteArray`, so why create a new `UnsignedByteArray` and then call `empty()`?

xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
```diff
@@ -41,4 +42,96 @@ public void lowerCaseOrUpperCase() {
...
+    assertThat(UnsignedByteArray.of(new byte[] {MAX_BYTE, MAX_BYTE}).empty().length()).isEqualTo(0);
+    assertThat(
+      UnsignedByteArray.of(new byte[] {0, MAX_BYTE}).empty().equals(UnsignedByteArray.of(new byte[] {}))
+    ).isTrue();
+  }
+
+  @Test
+  public void simpleLength() {
```
nkramer44, line 56: can we get rid of the "simple" prefix on all of these names? The fact that there is no "complicated" counterpart is a bit confusing lol

nkramer44 (changes requested):
Can you add the following tests as well?
1) One for `toByteArray()`
2) A test that calls `.equals()` on the same instance, like `assertThat(unsignedByteArray).isEqualTo(unsignedByteArray)`
3) A test that calls `.equals()` on an object of a different type. Something like
`assertThat(unsignedByteArray).isNotEqualTo(new Object())`

sappenin (changes requested):

xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
```diff
@@ -150,7 +162,7 @@ public void simpleHashcode() {
   }
 
   @Test
-  public void simpleToString() {
+  public void UnsignedByteArrayToString() {
```
sappenin, line 165:
```suggestion
  public void unsignedByteArrayToString() {
```

xrpl4j-address-codec/src/test/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArrayTest.java
sappenin: @nkramer44 I think this is now resolved with the latest commits from @mukulljangid - plz make another comment if you see something missing.

sappenin (approved):

nkramer44 (changes requested):
Small nit -- Can you add a test that calls `equals` with a non UnsignedByteArray type, like `assertThat(array.equals(new Object()).isFalse()`?

nkramer44 (approved):

PR #250, opened 2022-03-24
Add Max/Min IssuedCurrencyAmount value Strings

nkramer44:
Adds min and max constants in `IssuedCurrencyAmount`. Also adjust the acceptable exponent bounds in `AmountType` when encoding/decoding to/from binary.


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/v3/BinarySerializationTests.java
```diff
@@ -384,6 +387,184 @@ void serializeTrustLine() throws JsonProcessingException {
...
+  @Test
+  void serializeTrustSetWithMaxValueMinusOnePowerOfTen() throws JsonProcessingException {
+    TrustSet trustSet = TrustSet.builder()
+      .account(Address.of("r4V5fhT5KPCdhUanXCMpPUHiM4s6bXDVDH"))
+      .fee(XrpCurrencyAmount.ofDrops(10))
+      .sequence(UnsignedInteger.valueOf(5))
+      .limitAmount(IssuedCurrencyAmount.builder()
+        .currency("7872706C346A436F696E00000000000000000000")
+        .issuer(Address.of("rUcccYgErfgMifUHYDHed4ibTArjqfkU5R"))
+        .value(new BigDecimal(IssuedCurrencyAmount.MAX_VALUE).scaleByPowerOfTen(-1).toString())
```
sappenin, line 432: These exponent +-1 tests are useful (so we should keep them) but I was originally thinking just MAX-1 and MAX+1 to detect an "off-by-one" error and detect what happens on overflow. Could we add those too?

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/v3/BinarySerializationTests.java
nkramer44: We could add those, but XRP Ledger only understands a precision of 16. Our binary codec throws a different kind of error of the precision is off. Since I was trying to test the change in exponent bounds, I figured this was more relevant.

xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -27,8 +27,8 @@
   public static final int NATIVE_AMOUNT_BYTE_LENGTH = 8;
   public static final int CURRENCY_AMOUNT_BYTE_LENGTH = 48;
   private static final int MAX_IOU_PRECISION = 16;
-  private static final int MIN_IOU_EXPONENT = -96;
-  private static final int MAX_IOU_EXPONENT = 80;
+  private static final int MIN_IOU_EXPONENT = -81;
```
sappenin, line 30: Maybe add a Javadoc note here with a link to xrpl.org explaining why these don't match -96 and 80?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/v3/BinarySerializationTests.java
sappenin: Yeah, agree - I think they're both important.

https://twitter.com/codinghorror/status/506010907021828096

sappenin (approved):
Nice work!!

Issue #77, opened 2021-03-09
Add Constants for TransactionResult Codes


sappenin:
See Transaction result codes [here](https://xrpl.org/transaction-results.html). It would be useful to have these defined in constants so projects using this library don't accidentally introduce typos.

E.g.: 

```
String TES = "tes";
String TES_SUCCESS = TES + "SUCCESS";
```

PR #253, opened 2022-04-04
constants for TransactionResult codes

mukulljangid:
Addresses #77


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionResultCode.java
```diff
@@ -0,0 +1,146 @@
+package org.xrpl.xrpl4j.model.transactions;
+
+/**
+ * The rippled server summarizes transaction results with result codes, which appear
+ * in fields such as engine_result and meta.TransactionResult.
+ */
+public class TransactionResultCode {
```
nkramer44, line 7: Small nit, but can we rename this to `TransactionResultCodes`? Usually when you have a class that just contains a bunch of static methods, the class name is plural. The Guava library does this a lot, like `Strings`, `Lists`, `Maps` etc

nkramer44 (changes requested):

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionResultCodes.java
```diff
@@ -0,0 +1,146 @@
+package org.xrpl.xrpl4j.model.transactions;
+
+/**
+ * The rippled server summarizes transaction results with result codes, which appear
+ * in fields such as engine_result and meta.TransactionResult.
+ */
+public class TransactionResultCodes {
```
sappenin, line 7: Per our discussion, let's make a unit test for this one.

sappenin (changes requested):

sappenin (approved):

sappenin (approved):

Issue #221, opened 2022-02-01
Add helper method to order multisig signatures


sappenin:
Currently, we do this in an IT, but there's no "sugar" to help a developer do this correctly. 

See https://xrpl.org/transaction-malleability.html#transaction-malleability and code in SM.

E.g., 

```
 .sorted(
        Comparator.comparing(
          signature -> new BigInteger(addressCodec.decodeAccountId(signature.signer().account()).hexValue(), 16)
        )
      )
```

sappenin:
This is likely related to #152. It's possible if we simply re-order correctly in the encoder/decoder, then this issue may not be necessary.

PR #257, opened 2022-04-19
sort signers in the encoder and add test to verify hash

mukulljangid:
Rippled expects signers to be sorted by address as a part of submit_multisigned.
This PR sorts the list of signers in the encoder right before serializing it. 

Addresses: #221

### DIFF
--- a/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
+++ b/xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
@@ -22,20 +22,28 @@ package org.xrpl.xrpl4j.codec.binary.types;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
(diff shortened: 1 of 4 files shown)


nkramer44 (changes requested):
Looks good! Just a couple of comments.

Also, can you try to add some tests to `BinarySerializationTests` with signers in order and out of order?

xrpl4j-binary-codec/src/main/java/org/xrpl/xrpl4j/codec/binary/types/STObjectType.java
```diff
@@ -98,7 +105,29 @@ public STObjectType fromJson(JsonNode node) {
         continue;
       }
 
-      JsonNode fieldNode = node.get(fieldName);
+      JsonNode fieldNode;
+      // rippled expects signers canonically based on address
+      if (fieldName.equals("Signers")) {
+        final AddressCodec addressCodec = new AddressCodec();
+        ArrayNode arrayNode = (ArrayNode) node.get(fieldName);
+        List<JsonNode> jsonNodeList = new ArrayList<>();
```
sappenin, line 113: We need unit test coverage of this in addition to the ITs...that codecov tool in IDEA is your friend! :)

sappenin (changes requested):

sappenin (approved):

nkramer44 (approved):

Issue #258, opened 2022-04-21
LedgerIndexBound constraint is too strict


sappenin:
[This constraint](https://github.com/XRPLF/xrpl4j/blob/main/xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/common/Wrappers.java#L174) on LedgerIndexBound is too strict. For example, some responses to the `account_tx` method will return a response like this:

```
{"account":"r9m6MwViR4GnUNqoGXGa8eroBrZ9FAPHFS","marker":{"ledger":61363309,"seq":2},"ledger_index_max":61391079,"ledger_index_min":0,"transactions":[],"validated":true,"status":"success"}
```

sappenin:
This is a bug in the clio node powering s2.ripple.com. This bug will be fixed via https://github.com/XRPLF/clio/issues/148

PR #259, opened 2022-04-28
remove the strict `0` value check for ledger_index

mukulljangid:
Addresses: https://github.com/XRPLF/xrpl4j/issues/258


mukulljangid:
the one above is an unrelated change, but wanted to get this in, and didn't want to make a separate PR.

sappenin:
Before we merge this, we need to verify if this problem wasn't actually just caused by a bug in clio.

sappenin:
Looks like this is a bug in clio, so let's close this PR. Once I get the clio issue link I'll link it into #258.

PR #265, opened 2022-05-27
remove ledger value checks.

mukulljangid:
These checks were essentially incorrect ( from what I through my conversation with Nathan ). The docs read `earliest ledger actually searched for transactions` and won't guarantee the same values.

This comes to my attention when I ran RPC's on Clio specifically.

### DIFF
--- a/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AccountTransactionsIT.java
+++ b/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AccountTransactionsIT.java
@@ -104,8 +104,6 @@ public class AccountTransactionsIT {
         .build()
     );
 
-    assertThat(results.ledgerIndexMinimum()).isEqualTo(minLedger);
-    assertThat(results.ledgerIndexMaximum()).isEqualTo(maxLedger);
     assertThat(results.transactions()).hasSize(7);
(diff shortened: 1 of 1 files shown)


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AccountTransactionsIT.java
```diff
@@ -104,8 +104,6 @@ public void listTransactionWithIndexRange() {
         .build()
     );
 
-    assertThat(results.ledgerIndexMinimum()).isEqualTo(minLedger);
-    assertThat(results.ledgerIndexMaximum()).isEqualTo(maxLedger);
```
sappenin, line 108: I don't think we should remove these checks. Instead, I think we should either (1) Only talk to https://s2-clio.ripple.com or (2) Somehow figure out how to only talk to a reporting server, or (3) Get xrplcluster.com to support JSON-RPC and talk to that.

Overall, #1 is the future, but CLIO seems to be giving incorrect values, so I'm hesitant to use it. Simply removing these checks doesn't seem like the right path either though since these checks are serving a purpose.

sappenin (changes requested):

nkramer44 (approved):

sappenin (approved):

PR #272, opened 2022-06-06
add Base58EncodedSecret Model

mukulljangid:


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
```diff
@@ -192,4 +197,14 @@ public String toString() {
...
+
+  @Value.Immutable
+  @Wrapped
+  abstract static class Base58EncodedSecret extends Wrapper<String> {
+
+    @Value.Check
+    public void validateBase58EncodedSecret() {
+      assertDoesNotThrow(() -> Base58.decode(this.value()));
+    }
+  }
```
nkramer44, line 209: This isn't really providing any value as is. We already make sure the `String` you pass in to `fromBase58EncodedSecret` is Base58 encoded on line 125, so now our code has to do 2 Base58 decodes when it should really only need one. Base58 encoding/decoding is an expensive operation, so we don't want to do this unnecessarily. 

I could see this class being useful if we added a derived field containing the decoded `String`, and then using that field on line 125 instead of doing another decode.

xrpl4j-crypto-parent/xrpl4j-crypto-core/pom.xml
```diff
@@ -85,6 +85,11 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>compile</scope>
```
sappenin, line 91:
```suggestion
      <scope>test</scope>
```

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/keys/Seed.java
```diff
@@ -1,15 +1,20 @@
 package org.xrpl.xrpl4j.crypto.core.keys;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
sappenin, line 3: Jupiter imports should only exist in test classes, but not non-tests.

sappenin (changes requested):

xrpl4j-crypto-parent/xrpl4j-crypto-core/pom.xml
sappenin: This is now fixed.

sappenin (approved):

PR #277, opened 2022-06-10
adds ITs for NFT Transactions

mukulljangid:
* adds a CustomEnvironment to use in ITs
* Builds NftDevnetEnvironment on top of CustomEnvironment


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NfTokenIT.java
```diff
@@ -0,0 +1,464 @@
...
+    this.scanForResult(
+      () -> {
+        try {
+          return xrplClient.accountNfts(wallet.classicAddress());
+        } catch (JsonRpcClientErrorException e) {
+          throw new RuntimeException(e);
+        }
+      },
+      result -> result.accountNfts().size() == previousLength + 1
+    );
```
nkramer44, line 73: What if we include a URI in the NFT that we mint, and check that an NFT with the same URI exists in this check

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/XrplEnvironment.java
```diff
@@ -36,16 +36,20 @@ public interface XrplEnvironment {
...
     boolean isTestnetEnabled = System.getProperty("useTestnet") != null;
-    logger.info("useTestnet set to {}, using {} for testing.",
-      isTestnetEnabled, isTestnetEnabled ? "testnet" : "local rippled"
-    );
-    return isTestnetEnabled ? new TestnetEnvironment() : new LocalRippledEnvironment();
+    if (isTestnetEnabled) {
+      logger.info("System property 'useTestnet' detected; Using Testnet for integration testing.");
+      return new TestnetEnvironment();
+    } else {
+      logger.info("Neither 'useTestNet' nor 'useNftDevnet' nor 'useCustom' System properties detected." +
```
nkramer44, line 49: Since we removed the `useNftDevnet` sys property, let's fix this log statement

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NfTokenIT.java
```diff
@@ -0,0 +1,464 @@
...
+import org.xrpl.xrpl4j.tests.environment.XrplEnvironment;
+import org.xrpl.xrpl4j.wallet.Wallet;
+
+public class NfTokenIT extends AbstractIT {
+
+  private static final XrplEnvironment nftDevnetEnvironment = new CustomEnvironment(
+    HttpUrl.parse("http://xls20-sandbox.rippletest.net:51234"),
+    HttpUrl.parse("https://faucet-nft.ripple.com")
+  );
+  private static final XrplClient xrplClient = nftDevnetEnvironment.getXrplClient();
```
nkramer44, line 39: This can just be `xrplClient()`

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SignerListSetIT.java
```diff
@@ -51,11 +52,12 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
-public class SignerListSetIT extends AbstractIT {
+public class SignerListSetIT extends BaseIT {
 
   protected final ObjectMapper objectMapper = ObjectMapperFactory.create();
   protected final XrplBinaryCodec binaryCodec = new XrplBinaryCodec();
   protected final KeyPairService keyPairService = DefaultKeyPairService.getInstance();
+  private final XrplClient xrplClient = xrplClient();
```
nkramer44, line 60: We can pull all of these into `BaseIT` as a field

nkramer44 (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/IssuedCurrencyIT.java
```diff
@@ -42,10 +44,13 @@
...
 import java.util.List;
 
-public class IssuedCurrencyIT extends AbstractIT {
+public class IssuedCurrencyIT extends BaseIT {
+
+  private final XrplClient xrplClient = xrplClient();
 
   @Test
   public void issueIssuedCurrencyBalance() throws JsonRpcClientErrorException {
+    assumeFalse(System.getProperty("useNftDevnet") != null);
```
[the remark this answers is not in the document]
mukulljangid: Sorry. It was the devnet env where Issued Currency stuff wasn't supported. Now that we are not running all the IT's I'll remove this.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SignerListSetIT.java
mukulljangid: Argh. This was dumb.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BaseIT.java
```diff
@@ -0,0 +1,16 @@
+package org.xrpl.xrpl4j.tests;
+
+import org.xrpl.xrpl4j.client.XrplClient;
+import org.xrpl.xrpl4j.tests.environment.XrplEnvironment;
+
+public class BaseIT extends AbstractIT {
+
+  private static XrplEnvironment xrplEnvironment = XrplEnvironment.getConfiguredEnvironment();
+
+  protected static XrplClient xrplClient = xrplEnvironment.getXrplClient();
```
nkramer44, line 10:
```suggestion
  protected XrplClient xrplClient = xrplClient();
```

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NfTokenIT.java
```diff
@@ -0,0 +1,475 @@
...
+  @Test
+  void mintAndBurn() throws JsonRpcClientErrorException {
+    Wallet wallet = createRandomAccount();
+
+    AccountInfoResult accountInfoResult = this.scanForResult(
+      () -> this.getValidatedAccountInfo(wallet.classicAddress())
+    );
+
+    // nft mint
+    NfTokenUri uri = NfTokenUri.ofPlainText("ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf4dfuylqabf3oclgtqy55fbzdi");
```
nkramer44, line 90: You can't use the same URI for all of the NFTs you are minting in these ITs and reliably use the check below. If this test runs after `mint()` for example, that NFT will already exist, and you won't actually be checking that the NFT you are minting in this test got minted.

nkramer44 (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/NfTokenIT.java
mukulljangid: In each of these tests, we have different accounts generating them. So verifying shouldn't be a problem.

nkramer44: Ah good call. Nevermind then

nkramer44 (approved):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -493,6 +497,20 @@ public AccountInfoResult accountInfo(AccountInfoRequestParams params) throws Jso
...
+  /**
+   * Return AccountNftsResult for an {@link Address}.
+   *
+   * @param account to get the NFTs for.
+   *
+   * @return {@link AccountNftsResult} containing list of accounts for an address.
+   * @throws JsonRpcClientErrorException If {@code jsonRpcClient} throws an error.
+   */
+  public AccountNftsResult accountNfts(Address account) throws JsonRpcClientErrorException {
+    return this.accountNfts(
```
sappenin, line 509: Please add preconditions to assert non-null inputs on any functions.

```suggestion
    Objects.requireNonNull(account);
    return this.accountNfts(
```

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -493,6 +497,20 @@ public AccountInfoResult accountInfo(AccountInfoRequestParams params) throws Jso
...
+      AccountNftsRequestParams.builder().account(account).build()
+    );
+  }
+
   /**
    * Get the {@link AccountNftsResult} for the account specified in {@code params} by making an account_channels
```
sappenin, line 515: Is `account_channels` a typo?

```suggestion
   * Get the {@link AccountNftsResult} for the account specified in {@code params} by making an account_nfts
```

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/nft/SellOffer.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  @JsonProperty("Flags")
+  Flags.NfTokenOfferFlags flags();
+
+  /**
+   * The ledger object ID of this offer.
+   *
+   * @return The {@link String} index.
+   */
+  @JsonProperty("nft_offer_index")
```
sappenin, line 52: Is this casing correct? Standard XRPL API values have historically been cased differently (e.g., see line 44).

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BaseIT.java
```diff
@@ -0,0 +1,16 @@
+package org.xrpl.xrpl4j.tests;
+
+import org.xrpl.xrpl4j.client.XrplClient;
+import org.xrpl.xrpl4j.tests.environment.XrplEnvironment;
+
+public class BaseIT extends AbstractIT {
```
sappenin, line 6: From what I can tell, this class is not necessary. We can move the internals of this class up into `AbstractIT` and then remove this class entirely. Once done, the override in `NfTokenIT` will still work properly. Here's what the values in `AbstractIT` should look like:

```java
private static XrplEnvironment xrplEnvironment = XrplEnvironment.getConfiguredEnvironment();

protected XrplClient xrplClient = xrplClient();

protected XrplEnvironment xrplEnvironment() {
  return xrplEnvironment;
}

protected XrplClient xrplClient() {
    return xrplEnvironment().getXrplClient();
}
```

You'll then want to do the following:

1. Remove any references to `xrpClient()` and replace with `xrpClient`.
2. Get rid of the client declaration in `NfTokenIT` near line 39. The value in `AbstractIT` will work fine.

sappenin (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
mukulljangid: It is

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BaseIT.java
nkramer44: @sappenin I actually suggested this pattern to mukul, but I could be wrong. I wanted to avoid calling `XrplEnvironment.getConfiguredEnvironment` statically on all tests if you're just running a test that uses a custom environment. For example, if you only wanted to run the NFT tests and we call `XrplEnvironment.getConfiguredEnvironment()` in `AbstractIT`, and the configured environment is local rippled, your tests will end up spinning up a rippled container even though you overrode the `XrplEnvironment` in your test class.

If you only call `XrplEnvironment.getConfiguredEnvironment()` in `BaseIT`, a rippled container only gets started up if you are running a test that extends `BaseIT`, no?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -75,19 +75,38 @@ public abstract class AbstractIT {
...
 
+  protected static XrplEnvironment xrplEnvironment;
+
+  /**
+   * Helper method to initialize the {@link XrplEnvironment} and to also allow sub-classes to use a different
+   * environment. This construction allows sub-classes to choose an environment that doesn't spin-up a test container.
+   *
+   * @return An {@link XrplEnvironment}.
+   */
+  protected XrplEnvironment xrplEnvironment() {
```
sappenin, line 90: @mukulljangid @nkramer44 This is the pattern I had in mind. It allows three important things for our ITs:

1. It allows this `AbstractIT` or any sub-classes to use the `xrplEnvironment` variable directly so that if a developer forgets to call `xrplEnvironment()` as a function, things will still work.

2. It allows a potential future where we might run these ITs in parallel.

3. It enables the goal that @nkramer44 was hoping for which is that any IT that doesn't need a Testcontainer-based rippled won't spin one up un-necessarily. You can see now that `NfTokenIT` doesn't use a test-container due to the override of `xrplEnvironment()` in place in that class.

sappenin (approved):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BaseIT.java
sappenin: Yes, I agree with your goal @nkramer44 -- see my latest comment and latest commit. I think what I pushed accomplishes both of our goals.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -75,19 +76,33 @@ public abstract class AbstractIT {
...
 
+  protected static XrplEnvironment xrplEnvironment;
+
+  /**
+   * Helper method to initialize the {@link XrplEnvironment} and to also allow sub-classes to use a different
+   * environment. This construction allows sub-classes to choose an environment that doesn't spin-up a test container.
+   *
+   * @return An {@link XrplEnvironment}.
+   */
+  @BeforeAll
```
sappenin, line 91: @mukulljangid @nkramer44 This is the pattern I had in mind. It allows three important things for our ITs:

1. It allows this `AbstractIT` or any sub-classes to use the `xrplEnvironment` variable directly so that if a developer forgets to call `xrplEnvironment()` as a function, things will still work.

2. It allows a potential future where we might run these ITs in parallel.

3. It enables the goal that @nkramer44 was hoping for which is that any IT that doesn't need a Testcontainer-based rippled won't spin one up un-necessarily. You can see now that `NfTokenIT` doesn't use a test-container (due to the override of `initXrplEnvironment()` in place in that class) when executed on its own. However, when all tests are run together, the _other_ tests will initialize the xrplEnvironment correctly using the variant in `AbstractIT`.

Issue #275, opened 2022-06-08
Use FeeUtils.computeFeeDynamically in v3 ITs


mukulljangid:
Similar to https://github.com/XRPLF/xrpl4j/pull/260

mukulljangid:
PR has been merged. closing this

PR #280, opened 2022-06-23
Mj/use feeutils v3

mukulljangid:
Addresses #275


sappenin:
Let's consider the changes in https://github.com/XRPLF/xrpl4j/pull/282 before merging this one.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -229,11 +250,9 @@ public String toString() {
      */
     @Value.Check
     protected void check() {
-      Preconditions.checkState(
-        FluentCompareTo.is(value()).lessThanOrEqualTo(UnsignedLong.valueOf(MAX_XRP_IN_DROPS)),
-        String.format(
-          "XRP Amounts may not exceed %s drops (100B XRP, denominated in Drops)", FORMATTER.format(MAX_XRP_IN_DROPS))
-      );
+      Preconditions.checkState(FluentCompareTo.is(value()).lessThanOrEqualTo(UnsignedLong.valueOf(MAX_XRP_IN_DROPS)),
+        String.format("XRP Amounts may not exceed %s drops (100B XRP, denominated in Drops)",
```
sappenin, line 254: My formatter is off - I'll fix this one and some of the others.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -0,0 +1,421 @@
...
+   * current size of the job queue as compared to its total capacity.
+   *
+   * @param feeResult {@link FeeResult} object obtained by querying the ledger (e.g., via an `XrplClient#fee()` call).
+   *
+   * @return {@link XrpCurrencyAmount} value of the fee that should be used for the transaction.
+   *
+   * @see "https://xrpl.org/fee.html"
+   * @see "https://github.com/XRPL-Labs/XUMM-App/blob/master/src/services/LedgerService.ts#L244"
+   */
+  public static XrpCurrencyAmount calculateFeeDynamically(final FeeResult feeResult) {
```
sappenin, line 77: @mukulljangid @nkramer44 Two questions:

1. I don't like the name of this method because it's unclear what it's doing -- the usage of the word "dynamically" isn't very helpful to me, and it also doesn't really conform to the other public method in this class i.e., `computeMultiSigFee`). Do you have better ideas here? I was thinking maybe just `computeFee` but am curious to hear other opinions here. Xumm calls this `getAvailableNetworkFee`, so we could use `computeAvailableNetworkFee`
2. The Xumm version of this method actually [returns a three-choice object](https://github.com/XRPL-Labs/XUMM-App/blob/8e6e795648b97d0067380baa43bedea9e4b5ea61/src/services/LedgerService.ts#L295) (`low`, `medium` and `high`) and then shows these values to users so they can pick depending on their appetite -- e.g., a Xumm user may not want to always pay the highest fee, and Xumm seems to allow users to select this. I think we should do the same -- anyone using xrpl4j shouldn't be forced into using the fee that we choose for them, because an xrpl4j developer might want to override this behavior. Therefore, I propose that we return an object that contains the low/medium/high fees. WDYT?

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: This is fixed.

xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
mukulljangid: I like `computeNetworkFees` as well. I agree with point 2 as well. I'll make the change and add javadoc to guide the user.

xrpl4j-model/src/test/java/org/xrpl/xrpl4j/model/client/fees/FeeUtilsTest.java
```diff
@@ -68,7 +68,8 @@ public void simpleComputeMultiSigFee() {
       .index(Hash256.of("A9C28A28B85CD533217F5C0A0C7767666B093FA58A0F2D80026FCC4CD932DDC7"))
       .build();
 
-    assertThat(computeMultisigNetworkFees(feeResult, object)).isEqualTo(XrpCurrencyAmount.ofDrops(60));
+    assertThat(computeMultisigNetworkFees(feeResult, object).recommendedFee())
```
sappenin, line 71: Let's assert the actual values for `low`, `medium` and `high` to ensure they are as expected, just to make sure we're wiring up values correctly.

PR #303, opened 2022-09-29
fix `Signature` serialization and deserialization

### LINKED ISSUE #304: v3: serialize and deserialize `Signature` correctly
Currently, `Signature` is being serialized as: `{
  signature: {
    value: "12345"
  }
}`
Where as it should look like this: `{
  signature: "12345"
}`

This is a breaking change hence it will be part of v3.

mukulljangid:
fixes #304

### DIFF
--- a/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/Signature.java
+++ b/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/Signature.java
@@ -20,15 +20,28 @@ package org.xrpl.xrpl4j.crypto.signing;
  * =========================LICENSE_END==================================
  */
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.io.BaseEncoding;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Derived;
 import org.xrpl.xrpl4j.codec.addresses.UnsignedByteArray;
 
+import java.io.IOException;
+
 /**
  * Represents a digital signature for a transaction that can be submitted to the XRP Ledger.
  */
 @Value.Immutable
+@JsonSerialize(using = Signature.SignatureSerializer.class)
+@JsonDeserialize(using = Signature.SignatureDeserializer.class)
 public interface Signature {
 
   /**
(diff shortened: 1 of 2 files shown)


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/test/java/org/xrpl/xrpl4j/crypto/signing/SignatureTest.java
```diff
@@ -0,0 +1,61 @@
...
+    assertThat(signature).isEqualTo(deserialized);
+
+    WrappedSignature wrappedSignature = WrappedSignature.builder()
+      .signature(signature)
+      .build();
+    String serializedWrappedSignature = objectMapper.writeValueAsString(wrappedSignature);
+
+    String serialized = "{\"signature\":\"E2ACD61C90D93433402B1F704DA38DF72876B6788C2C05B3196E14BC711AECFF14A7D627643" +
+      "9A198D8B4880EE2DB544CF351A8CE231B3340F42F9BF1EDBF5104\"}";
+    assertThat(serializedWrappedSignature).isEqualTo(serialized);
```
mukulljangid, line 42: Checkout this serialized object! @florent-uzio

florent-uzio: Ah nice, thanks @mukulljangid , once this is merged and used in the cbdc-backend we will be able to merge our UI branch.
https://github.com/xpring-eng/cbdc-admin-ui/pull/323

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/Signature.java
```diff
@@ -56,4 +69,27 @@ static ImmutableSignature.Builder builder() {
...
+   */
+  class SignatureDeserializer extends JsonDeserializer<Signature> {
+
+    @Override
+    public Signature deserialize(JsonParser jsonParser, DeserializationContext ctxt)
+      throws IOException {
+      JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+      return Signature.builder().value(UnsignedByteArray.fromHex(node.asText())).build();
+    }
+  }
```
nkramer44, line 94: Let's put these in `xrpl4j-model` in the `jackson.modules` package

nkramer44 (changes requested):
I think if we merge this into `main` it would technically introduce a breaking change, which I don't want to do. In the v3 branch this `Signature` is deprecated and replaced by `org.xrpl.xrpl4j.crypto.core.signing.Signature`. I think we should leave this `Signature` alone, switch over to the v3 branch and make the change on the new `Signature` class.

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/Signature.java
mukulljangid: `Signature` is  a part of the `xrpl4j-crypto-core` module and might introduce circular dependency if we add crypto-core as a dependency in `xrpl4j-model` .

mukulljangid:
Closing this. And opening a new PR which changes the model in v3 `Siganture`.

PR #305, opened 2022-09-30
add serializer and deserializer for `Signature`

mukulljangid:
addresses: https://github.com/XRPLF/xrpl4j/issues/304

### DIFF
--- a/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
+++ b/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
@@ -12,8 +12,8 @@ import org.xrpl.xrpl4j.codec.addresses.UnsignedByteArray;
  * Represents a digital signature for a transaction that can be submitted to the XRP Ledger.
  */
 @Value.Immutable
-@JsonSerialize(using = SignatureSerializer.class)
-@JsonDeserialize(using = SignatureDeserializer.class)
+@JsonSerialize(as = ImmutableSignature.class, using = SignatureSerializer.class)
+@JsonDeserialize(as = ImmutableSignature.class, using = SignatureDeserializer.class)
 public interface Signature {
 
   /**
--- a/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
+++ b/xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
@@ -12,8 +12,8 @@ import org.xrpl.xrpl4j.codec.addresses.UnsignedByteArray;
  * Represents a digital signature for a transaction that can be submitted to the XRP Ledger.
  */
 @Value.Immutable
-@JsonSerialize(as = ImmutableSignature.class)
-@JsonDeserialize(as = ImmutableSignature.class)
+@JsonSerialize(using = SignatureSerializer.class)
+@JsonDeserialize(using = SignatureDeserializer.class)
 public interface Signature {
 
   /**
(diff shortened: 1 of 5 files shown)


xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/SignatureSerializer.java
```diff
@@ -0,0 +1,17 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+
+import java.io.IOException;
+
+/**
+ * Custom Jackson serializer for {@link Signature}es.
+ */
+class SignatureSerializer extends JsonSerializer<Signature> {
```
nkramer44, line 12: Let's move this and the deserializer into the same package as the other serializers/deserializers and add them to the `Xrpl4jModule` so you don't need the `using = SignatureSerializer` and `using = SignatureDeserializer.class`

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
```diff
@@ -12,8 +12,8 @@
  * Represents a digital signature for a transaction that can be submitted to the XRP Ledger.
  */
 @Value.Immutable
-@JsonSerialize(as = ImmutableSignature.class)
-@JsonDeserialize(as = ImmutableSignature.class)
+@JsonSerialize(using = SignatureSerializer.class)
+@JsonDeserialize(using = SignatureDeserializer.class)
```
nkramer44, line 16: These will probably get removed once you add the serializer/deserializer to the Xrpl4jModule, but for future reference, you need both the `as` and the `using` fields in these annotations for things to work correctly with immutables.

nkramer44 (changes requested):

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
mukulljangid: https://github.com/XRPLF/xrpl4j/pull/303#discussion_r984657149

nkramer44 (approved):

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/SignatureSerializer.java
sappenin: This still needs to be addressed, correct?

sappenin (changes requested):

mukulljangid:
> This still needs to be addressed, correct?

Not at the moment. That would need bigger refactor of modules

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/SignatureSerializer.java
nkramer44: The problem is that `Signature` is in xrpl4j-crypto, which depends on `xrpl4j-model`, but all of the Jackson modules are in `xrpl4j-model`, so there's no way to move these without creating a circular dependency. We probably want to extract all the Jackson stuff into it's own module at some point, but I think that should be future work. WDYT?

xrpl4j-crypto-parent/xrpl4j-crypto-core/src/main/java/org/xrpl/xrpl4j/crypto/core/signing/Signature.java
sappenin: This will be fixed via #306.

sappenin (approved):

Issue #299, opened 2022-08-08
Use new `serverInformation` in ITs


mukulljangid:

nkramer44:
This was fixed in #369 so I'm closing this issue.

PR #332, opened 2022-12-07
Use new serverInformation in ITs

rootVegtables:
Fixes: #299

### DIFF
--- a/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
+++ b/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
@@ -9,6 +9,7 @@ import org.slf4j.LoggerFactory;
 import org.xrpl.xrpl4j.client.JsonRpcClientErrorException;
 import org.xrpl.xrpl4j.client.XrplClient;
 import org.xrpl.xrpl4j.model.client.serverinfo.ServerInfo;
+import org.xrpl.xrpl4j.model.client.serverinfo.ServerInfoResult;
 
 import java.util.concurrent.TimeUnit;
 
@@ -49,8 +50,8 @@ public class ServerInfoIT {
     }
   }
 
-  private String getType(ServerInfo info) {
-    return info.map(
+  private String getType(ServerInfo result) {
+    return result.map(
       rippled -> "rippled",
       clio -> "clio",
       reporting -> "reporting"
--- a/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
+++ b/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
@@ -50,8 +50,8 @@ public class ServerInfoIT {
     }
   }
 
-  private String getType(ServerInfo result) {
-    return result.map(
+  private String getType(ServerInfo serverInfo) {
+    return serverInfo.map(
       rippled -> "rippled",
       clio -> "clio",
       reporting -> "reporting"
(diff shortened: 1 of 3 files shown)


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
```diff
@@ -49,8 +50,8 @@ public void testServerInfoAcrossAllTypes() throws JsonRpcClientErrorException, I
     }
   }
 
-  private String getType(ServerInfo info) {
-    return info.map(
+  private String getType(ServerInfo result) {
```
sappenin, line 53: This rename seems a bit off. We should either leave it alone or rename it to `serverInfo`.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/ServerInfoIT.java
```diff
@@ -9,6 +9,7 @@
 import org.xrpl.xrpl4j.client.JsonRpcClientErrorException;
 import org.xrpl.xrpl4j.client.XrplClient;
 import org.xrpl.xrpl4j.model.client.serverinfo.ServerInfo;
+import org.xrpl.xrpl4j.model.client.serverinfo.ServerInfoResult;
```
sappenin, line 12: This import seems unused.

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/LocalRippledEnvironment.java
```diff
@@ -55,7 +55,7 @@ public XrplClient getXrplClient() {
 
   @Override
   public void fundAccount(Address classicAddress) {
-    // accounts are funded from the genesis account that holds all XRP when the ledger container starts.
+    // accounts are funded  from the genesis account that holds all XRP when the ledger container starts.
```
nkramer44, line 58: Let's revert this

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/RippledContainer.java
```diff
@@ -99,9 +100,9 @@ public RippledContainer start() {
   /**
    * Start contain with given interval for closing ledgers.
    *
-   * @param acceptIntervalMillis
+   * @param acceptIntervalMillis the startup time limit.
```
nkramer44, line 103: This is actually the interval at which we close ledgers in the rippled container, so I would say something like `The number of milliseconds between ledger version closes`

nkramer44 (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/RippledContainer.java
```diff
@@ -132,7 +133,11 @@ private void waitForLedgerTimeToSync() {
 
   private ZonedDateTime getLedgerTime() {
     try {
-      return getXrplClient().serverInfo().time();
+      return getXrplClient().serverInformation().info().map(
+        rippled -> rippled.time(),
+        clio -> null,
```
sappenin, line 138: We shouldn't return `null` from this method because it breaks the contract (i.e., `ZonedDateTime` isn't `Optional`.

That said, what if for `clio` and `reporting` we simply returned the host's time. There's no other time to get anyway, and up above where this method is used, the "wait until system time matches the host time" will be essentially reduced towards 0.

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/RippledContainer.java
sappenin: Also, a unit test for this method should exist (might need to make the method `package-private` to pull that off.

nkramer44:
@rootVegtables We ended up having to do this work in order to get a v3.0.0 out, so going to close this PR

PR #336, opened 2022-12-15
Fix CI Builds

nkramer44:
Our github actions pipeline runs several maven builds for different version of Java. Every build was running on the `ubuntu-latest`, however Github has recently updated it's "latest" version of ubuntu. I'm not totally sure why the builds fail in the latest ubuntu version, but pegging ubuntu to version 20.04 seems to fix the issue.

Additionally, this PR changes `NfTokenITs` to use the configured `XrplEnvironment` for all ITs instead of a `CustomEnvironment` pointed at NFT Devnet. NFT Devnet has been deprecated and has been experiencing issues lately, which exposed the fact that overwriting `AbstractIT.xrplEnvironment` in `NfTokenIT` will cause all ITs run after `NfTokenIT` to point to NFT devnet. Removing this logic in `NfTokenId` both fixes this problem and will help down the line when NFT Devnet is fully decommissioned.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -6,7 +6,7 @@ on:
     types: [assigned]
 jobs:
   build_java8:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-21.10
     steps:
       # Checks-out the repository under $GITHUB_WORKSPACE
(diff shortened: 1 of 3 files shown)


sappenin (approved):
LGTM!

Issue #341, opened 2023-01-04
Need to Make Destination Tag to be Unsigned Long


txysystems:
Hello. I'm a developer and a big fan of ripple. If you will belief me, IT IS THE BEST CRYPTOCURRENCY FOR ME.
My issue is that, I'm working on developing a ripple wallet, using the official java library (xrpl4j), but I realize that most of the crypto exchange platforms uses destination tags higher than the maximum value of UnsignedInteger specified in the api.
To overcome this limitation, I hereby plead that THE DESTINATION TAG VARIABLE TYPE SHOULD BE CHANGED FROM UNSIGNEDINTEGER TO UNSIGNEDLONG.
Thanks in anticipation of a relieving response.
Love you all.

nkramer44:
Hi @txysystems, I'm not aware of any exchanges that use destination tags larger than a 32-bit unsigned integer. In fact, I'm not sure how that would be possible, considering the XRPL protocol enforces that destination and source tags are 32-bit unsigned integers. the `UnsignedInteger` class represents a 32-bit unsigned integer, so should be large enough to hold any destination tag. 

Can you tell me which on which exchange specifically you are seeing these larger destination tags?

sappenin:
Closing this as `wont-fix` since `UnsignedInteger` holds unsigned integers up to 2^32-1 (i.e., that can fit into 32-bits), and per the docs [here](https://xrpl.org/source-and-destination-tags.html):

> In transactions, both source and destination tags are formatted as 32-bit unsigned integers.

You can also see in transactions (e.g., a [Payment](https://xrpl.org/payment.html)) that destination tags are typed as `UInt32`.

Please re-open or make a comment here if there's still a concern.

PR #342, opened 2023-01-05
Releases/v2.5

txysystems:
Thanks for creating and maintaining this wonderful library.
The issue i'm having is that, some crypto exchanges like binance allocate some deposit tags to their customers, and this deposit tags are numbers higher than the maximum limit of UnsignedInteger currently specified by the xrpl4j library, thereby making it difficult to send fund to their wallet via the api.
In view of the above, i am pleading that the **UnsignedInteger field type specified in the library for destination tags should be changed to UnsignedLong**.
Thank you as i await your response / action.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <groupId>org.xrpl</groupId>
   <artifactId>xrpl4j-parent</artifactId>
   <packaging>pom</packaging>
-  <version>2.5.0</version>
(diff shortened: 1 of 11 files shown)


nkramer44:
@txysystems I responded to #341. Issues are the best way to request changes to the library, or you can feel free to propose the change yourself by opening a PR. Closing this PR as we never merge release branches back to `main`.

PR #344, opened 2023-01-19
Add AMM support

nkramer44:
**Do not merge this until the AMM amendment is endorsed on XRPL Mainnet**

This PR adds support for XLS-30 AMMs.

### New Transaction Types
* `AmmCreate`
* `AmmDeposit`
* `AmmBid`
* `AmmVote`
* `AmmWithdraw`
* `AmmDelete`

### New Client Methods
* `XrplClient.ammInfo` calls `amm_info`, taking `AmmInfoRequestParams` as a parameter and returning an `AmmInfoResult`

### Other Changes
* XLS-30 introduces a new serializable type called `Issue` (used to specify `Asset` and `Asset2` in an AMM). This PR adds an `Issue` type (called `IssueType`) to the binary codec module to encode/decode these values.
* This PR includes both objects for the `AMM` ledger entry as well as the `amm_info` result object. While `AmmInfoResult.amm()` is very similar to the `AMM` ledger entry, the rippled API formats the AMM object differently in responses to `amm_info`. The `AmmObject` in this PR is basically inaccessible because `XrplClient` doesn't support the `ledger_entry` RPC, but it will be good to have in the codebase once we add that method


sappenin (approved):
LGTM

PR #355, opened 2023-01-25
Improve MultiSignedTransaction interface

nkramer44:
This PR does the following:
* Adds a static instance of `XrplBinaryCodec`
* Removes `TransactionHashUtils` (v3 version)
* Removes the SHA256 field from `PrivateKey` (v3 version)
* Removes `SignableTransaction` (v3 version) because it is unused
* Derives `SignedTransaction.signedTransactionBytes` and `MultiSignedTransaction.signedTransaction` based on `unsignedTransaction` and `signatureAndPublicKeySet`
* Adds a new method to `XrplClient` `submitMultisigned(MultiSignedTransaction)`


xrpl4j-model/src/main/java/org/xrpl/xrpl4j/model/transactions/Payment.java
```diff
@@ -30,6 +30,7 @@
 
 import java.util.List;
 import java.util.Optional;
+import javax.annotation.Nullable;
```
sappenin, line 33: This import is unused.

sappenin (changes requested):
One minor nit, otherwise LGTM!

PR #356, opened 2023-01-25
Reorganize modules

### LINKED ISSUE #185: Refactor project modules to improve developer experience
In order to use xrpl4j `v1.x` and `v2.x`, a developer must import a variety of maven coordinates in their `pom.xml` project files.

Instead, it would be preferable to have (nearly) all library code in a single module called `xrpl4j-core` so that a developer only has to import a single dependency to utilize the entire library.

Once complete, the following will be the only two modules in the project:

* `xrpl4j-core` (e.g., a service that creates signatures but doesn't need a client)
* `xrpl4j-intergration-tests` (not in bom)

Maybe...
* `xrpl4j-crypto` (parent module (to allow for both BC and Java17 implementations)
* `xrpl4j-bom` (needed if there's more than 1 module that devs would use).
* `xrpl4j-client` (e.g., do we want to force anyone pulling in `core` to also take a WS and RPC client)

nkramer44:
Partially fixes #185 by condensing xrpl4j-model, xrpl4j-crypto-core, xrpl4j-crypto-parent, xrpl4j-keypairs, xrpl4j-binary-codec, and xrpl4j-address-codec into the xrpl4j-core module.

I kept the package structures the same for ease of review, but we should discuss if we want to keep the package structure this way or change it.


sappenin (approved):
LGTM!

PR #357, opened 2023-01-25
Remove v2 ITs and move v3 ITs into root package

nkramer44:
Removes all v2 ITs from the org.xrpl.xrpl4j.tests package and moves all v3 ITs from the org.xrpl.xrpl4j.tests.v3 packages into the org.xrpl.xrpl4j.tests package


sappenin (approved):
LGTM!

Issue #363, opened 2023-01-26
Decompose Flags into separate classes in the `.flags` package


sappenin:
Issue #32 should be addressed in-tandem with this issue.

PR #368, opened 2023-01-26
Decompose Flags classes and remove tfFullyCanonicalSig from flags builders

### LINKED ISSUE #32: Remove tfFullyCanonicalSig flag from all Flags Builders
The `tfFullyCanonicalSig` transaction flag was recently deprecated, and no new transactions submitted without this flag will be accepted.  Currently, we do allow users to set `tfFullyCanonicalSig` to `false` for transactions which have custom flags, so we should always set `tfFullyCanonicalSig` to `true`.  This can be done by removing the `tfFullyCanonicalSig(boolean)` method from every `TransactionFlags` `Builder` and always setting it to `true`.

nkramer44:
Fixes #32 and #363


xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/PaymentFlagsJsonTests.java
```diff
@@ -45,35 +45,7 @@ public void testJson() throws JsonProcessingException, JSONException {
       .destination(Address.of("r4BPgS7DHebQiU31xWELvZawwSG2fSPJ7C"))
       .amount(XrpCurrencyAmount.ofDrops(25000000))
       .fee(XrpCurrencyAmount.ofDrops(10))
```
sappenin, line 47: Is it not necessary to test Payment without flags?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BinarySerializationTests.java
```diff
@@ -386,7 +388,7 @@ void serializeTrustLine() throws JsonProcessingException {
       .build();
 
 
-    String expectedBinary = "1200142200020000240000002C63D6438D7EA4C68000000000000000000000000000574347000000" +
+    String expectedBinary = "1200142280020000240000002C63D6438D7EA4C68000000000000000000000000000574347000000" +
```
sappenin, line 391: I assume this change due to default canonical setting?

nkramer44: Exactly. The only difference is in the 9th character

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/PaymentFlagsJsonTests.java
nkramer44: Probably should. I'll put this back.

sappenin (approved):
LGTM!

PR #369, opened 2023-01-27
Remove all Deprecated classes

### LINKED ISSUE #365: Remove Deprecated Classes
Now that v3 code has been merged to `main`, we should delete all deprecated classes. We should take notes on the upgrade path for each deprecated class as we do this to help with writing a migration plan.

nkramer44:
Fixes #365


sappenin (approved):
Awesome!

PR #389, opened 2023-01-31
Remove SignatureWithPublicKey

sappenin:


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Signer.java
```diff
@@ -51,7 +52,10 @@ static ImmutableSigner.Builder builder() {
    * @return The {@link Address} of the signer account.
    */
   @JsonProperty("Account")
-  Address account();
+  @Default
```
sappenin, line 55: This should likely be `@Derived` but we probably want to leave it as `@Default` so that if this value comes back from rippled, we honor the value supplied by the server as opposed to a derived value in Java.

WDYT @nkramer44?

nkramer44: Agreed

V3_MIGRATION.md
```diff
@@ -220,10 +220,10 @@ xrplClient.submit(signedTransaction);
 To create a `MultiSignedTransaction`, use `BcSignatureService` like this:
 
 ```java
-Set<SignatureWithPublicKey> signers = Lists.newArrayList(aliceKeyPair, bobKeyPair).stream()
+Set<Signer> signers = Lists.newArrayList(aliceKeyPair, bobKeyPair).stream()
   .map(keyPair -> {
       Signature signedPayment = signatureService.multiSign(keyPair.privateKey(), unsignedPayment);
-      return SignatureWithPublicKey.builder()
+      return Signer.builder()
```
nkramer44, line 226: What do you think about returning `Signer` from `SignatureService.multiSign`?

nkramer44 (changes requested):

nkramer44 (approved):

V3_MIGRATION.md
nkramer44: We decided that returning `Signer` from `multiSign` would require the implementation of `multiSign` to call `this.derivePublicKey`. In most cases, the public key will already exist in memory at the time of calling `multiSign`, so returning a `Signer` would introduce an extraneous calculation or call to an external system. This seems like it could create performance problems for developers, so we chose to keep this as `Signature`

Issue #431, opened 2023-06-28
Add missing fields to NfTokenObject


nkramer44:
NfTokenObject is supposed to represent the NFTs returned from the `account_nfts` RPC, however it only contains the `NFTokenID` and `URI` fields. The response contains [some other fields](https://xrpl.org/account_nfts.html#response-format) that we should add to `NFTokenObject`

sappenin:
This is now fixed via #432

Issue #426, opened 2023-06-22
Things to change in v4 (maybe)


sappenin:
- Transaction Flags are technically [optional](https://github.com/XRPLF/rippled/blob/71d7d67fa39af3af5f36e136a7b6e5ddced08e49/src/ripple/protocol/impl/TxFormats.cpp#L30), so the Java API should make them optional in each transaction. 
  - E.g., in Payment, the API should be `Optional<PaymentFlags> flags()` with the `default` implementation removed.
  - This is more idiomatically Java than the v3 solution, which was implemented via #425.
  - Note that LedgerObject flags are _not_ optional, so we need to make sure that classes like `AccountRootObject` do not make their flags `Optional`.

sappenin:
Consider this list for any new API endpoints or breaking changes: https://github.com/XRPLF/rippled/issues?q=label%3A%22API+Change%22+is%3Aclosed

PR #432, opened 2023-06-29
Add objects for NFTokenPages

### LINKED ISSUE #430: Add NFTokenPage Object
Previously, the `NFTokenPage` ledger object was inaccessible via xrpl4j because it required a `ledger_entry` RPC which we don't support (yet). Therefore, we never added a `LedgerObject` for `NFTokenPage`. However, once https://github.com/XRPLF/rippled/pull/4352 is running on ledger, `NFTokenPage`s will show up in responses to `account_objects`.

nkramer44:
Fixes #430 and #431


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/NfTokenFlags.java
```diff
@@ -0,0 +1,114 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =========================LICENSE_END==================================
+ */
+
+import org.xrpl.xrpl4j.model.transactions.AccountSet;
+
+/**
+ * A set of static {@link Flags} which can be set on {@link AccountSet} transactions.
```
sappenin, line 26: This comment is incorrect.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountObjectsRequestParams.java
```diff
@@ -61,6 +61,7 @@ static ImmutableAccountObjectsRequestParams.Builder builder() {
   static AccountObjectsRequestParams of(Address classicAddress) {
     return builder()
       .account(classicAddress)
+      .ledgerSpecifier(LedgerSpecifier.CURRENT)
```
sappenin, line 64: Why is `CURRENT` the chosen default here?

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/accounts/AccountObjectsRequestParams.java
nkramer44: This was actually a bug that I found that we didn't have test coverage for. Before this change, `AccountObjectsRequestParams.of` would throw an `IllegalStateException` always. Many of the other request params  classes (ie `AccountInfoRequestParams`) have a similar method, and they all default to `LedgerSpecifier.CURRENT`, so this is consistent with those.

sappenin: That makes sense -- however, shouldn't we choose `VALIDATED` in the default case?

nkramer44: After speaking with @sappenin, we decided to potentially change the default from `CURRENT` to `VALIDATED` in v4 (see #426), as changing this default would require changing other defaults, which would be a breaking change.

PR #440, opened 2023-07-07
Add support for book_offers RPC

### LINKED ISSUE #411: Add BookOffers order book method
BookOffers path and order book method does not appear to be implemented in the xrpl4j api model

nkramer44:
Fixes #411


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -639,7 +598,6 @@ public <T extends Transaction> TransactionResult<T> transaction(
    * Get the contents of a ledger by sending a ledger method request.
    *
    * @param params The {@link LedgerRequestParams} to send in the request.
-   *
```
sappenin, line 642: I think your formatter is off.

sappenin (approved):

PR #461, opened 2023-07-31
Support for NetworkID

### LINKED ISSUE #427: Add `network_id` to ServerInfo respones
This value was added to rippled `v1.11.0`. Unclear if this is planned for reporting-mode (likely will occur for Clio), but in any event should _probably_ be optional for backward compatibility...?

nkramer44:
Adds support for the `NetworkID` transaction field. This field is `Optional` in `Transaction`, as it is only required on networks with ID > 1024.

Fixes #427

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/NetworkIdSerializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/NetworkIdSerializer.java
@@ -1,0 +1,46 @@
+package org.xrpl.xrpl4j.model.jackson.modules;
+
+/*-
+ * ========================LICENSE_START=================================
+ * xrpl4j :: core
+ * %%
+ * Copyright (C) 2020 - 2023 XRPL Foundation and its contributors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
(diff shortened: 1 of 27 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/NetworkIdSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Custom Jackson serializer for {@link NetworkId}s.
+ */
+public class NetworkIdSerializer extends StdScalarSerializer<NetworkId> {
+
+  /**
+   * No-args constructor.
+   */
+  public NetworkIdSerializer() {
+    super(NetworkId.class, false);
```
sappenin, line 39:
```suggestion
    super(NetworkId.class);
```

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/BinarySerializationTests.java
```diff
@@ -134,9 +136,10 @@ public void serializeAccountDelete() throws JsonProcessingException {
       .sequence(UnsignedInteger.valueOf(2470665))
       .destination(Address.of("rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe"))
       .destinationTag(UnsignedInteger.valueOf(13))
+      .networkId(NetworkId.of(UnsignedInteger.MAX_VALUE))
```
sappenin, line 139: Let's add one test that is explicitly indicates (via its name) that the test is testing the absence of a network id on a particular transaction's binary encoding.

sappenin (approved):

PR #465, opened 2023-08-01
Fix NfTokenBuyOffers and NfTokenSellOffers RequestParams/Results

nkramer44:
I noticed that there were some missing fields in `NfTokenBuyOffersRequestParams`, `NfTokenBuyOffersResult`, `NfTokenSellOffersRequestParams` and `NfTokenSellOfferResult`.

* Both request params classes were missing a `LedgerSpecifier` field. I defaulted both of them to `VALIDATED` to keep the existing behavior
* Both result classes were missing `limit` and `marker` fields, which made it impossible to page through results.
* `BuyOffer` and `SellOffer` both had `Amount` and `Flags` as the `@JsonProperty` field names, but the actual json uses `amount` and `flags`. This wasn't actually an issue because Jackson seems to handle the wrong casing, but should be fixed regardless


sappenin (approved):
LGTM!

PR #466, opened 2023-08-01
Add nft_info support

nkramer44:
Adds support for the Clio [`nft_info`](https://xrpl.org/nft_info.html) RPC.

Also, `TransferFee.ofPercent` was not `public`, and was also incorrectly implemented. It used 2 decimal places of precision, but transfer fees use 3 decimal places of precision (ie it would translate `45.99%` into `4599` when it should translate it to `45,990`. This is not a breaking change because the `ofPercent` method was not publicly accessible.


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/nft/NftInfoResult.java
```diff
@@ -0,0 +1,118 @@
...
+    return ImmutableNftInfoResult.builder();
+  }
+
+  /**
+   * A unique identifier for the non-fungible token (NFT).
+   *
+   * @return An {@link NfTokenId}.
+   */
+  @JsonProperty("nft_id")
+  NfTokenId nfTokenId();
```
sappenin, line 42:
```suggestion
  NfTokenId nftId();
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -352,12 +352,12 @@ public String toString() {
      * @param percent of type {@link BigDecimal}
      * @return {@link TransferFee}
      */
-    static TransferFee ofPercent(BigDecimal percent) {
```
sappenin, line 355: Can we add this to the Javadoc?

```
Valid values for this field are between 0 and 50000 inclusive, allowing transfer rates of between 0.00% and 50.00% in increments of 0.001. If this field is provided, the transaction MUST have the [tfTransferable flag](https://xrpl.org/nftokenmint.html#nftokenmint-flags) enabled.
```

Also, add an `@see https://xrpl.org/nftokenmint.html#nftokenmint-flags`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -352,12 +352,12 @@ public String toString() {
...
+    public static TransferFee ofPercent(BigDecimal percent) {
```
sappenin, line 355:
```suggestion
    public static TransferFee ofPercent(BigDecimal percent) {
       Objects.requireNonNull(percent);
```

sappenin (changes requested):
Minor nits.

sappenin (approved):

PR #479, opened 2023-08-11
Allow for No-Op AccountSet SetFlag and ClearFlag values

### LINKED ISSUE #477: LedgerResult Deserialization: AccountSet SetFlag/ClearFlag can be any 32 bit number
The docs say that the `AccountSet` `SetFlag` and `ClearFlag` must be a number between 1 and 15 (ie must be no more than 4 lowest bits set). However, there are some historical `AccountSet` transactions that has `SetFlag` or `ClearFlag` set to 4294967295, which is that max unsigned 32 bit number. This transaction does nothing because the `AccountSet` transactor checks for equality between `SetFlag`/`ClearFlag` and certain `asf` flags, but it does break our deserialization. This is because `AccountSetFlag` is an enum that wraps an `int`, and 4294967295 is too large for Java signed ints. 

The `@JsonCreator` on `AccountSetFlag` is public, so we can't just make it take a `long` instead of an `int` without a breaking change. However, we should probably add another method that takes a `long` and returns a new enum value called `UNKNOWN` or `NONE` if the flag is unrecognized rather than throwing an exception.

We should also consider filing a bug in rippled, as the transactor should likely limit the `SetFlag` and `ClearFlag` to the established `asf` flag values and return an error code if not.

Ledgers where this is a problem:
* 596C086FA83918033DF1A7FF31CA1D3E3526DE0533217A8A324D97DF7CA39E98

nkramer44:
Fixes #477

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
@@ -141,6 +141,8 @@ public interface AccountSet extends Transaction {
       // This can happen if:
       //   1. A developer sets clearFlagRawValue manually in the builder
       //   2. JSON has ClearFlag and jackson sets clearFlagRawValue.
+      // This value will never be negative due to XRPL representing this kind of flag as an unsigned number,
+      // so no lower bound check is required.
       if (clearFlagRawValue().get().longValue() <= 15) {
         // Set clearFlag to clearFlagRawValue if clearFlagRawValue matches a valid AccountSetFlag variant.
         return AccountSet.builder().from(this)
@@ -232,7 +234,8 @@ public interface AccountSet extends Transaction {
       // This can happen if:
       //   1. A developer sets setFlagRawValue manually in the builder
       //   2. JSON has ClearFlag and jackson sets setFlagRawValue.
-
+      // This value will never be negative due to XRPL representing this kind of flag as an unsigned number,
+      // so no lower bound check is required.
       if (setFlagRawValue().get().longValue() <= 15) {
         // Set setFlag to setFlagRawValue if setFlagRawValue matches a valid AccountSetFlag variant.
         return AccountSet.builder().from(this)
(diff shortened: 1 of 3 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
```diff
@@ -65,25 +65,185 @@ default AccountSetTransactionFlags flags() {
...
+      return this;
+    } else if (clearFlag().isPresent() && !clearFlagRawValue().isPresent()) {
+      // This can only happen if the developer only set clearFlag(). In this case, we need to set clearFlagRawValue to
+      // match clearFlag.
+      return AccountSet.builder().from(this)
+        .clearFlagRawValue(UnsignedInteger.valueOf(clearFlag().get().getValue()))
+        .build();
+    } else { // clearFlag not present and clearFlagRawValue is present
+      // This can happen if:
+      //   1. A developer sets clearFlag2 manually in the builder
```
sappenin, line 142: Typo - no clearflag2

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountSet.java
```diff
@@ -65,25 +65,185 @@ default AccountSetTransactionFlags flags() {
...
+      //   2. JSON has ClearFlag and jackson sets clearFlag2.
+      if (clearFlagRawValue().get().longValue() <= 15) {
```
sappenin, line 144: I think zero could creep in here, no?

nkramer44: It could, but `AccountSetFlag.NONE` exists and is mapped to zero so that should be fine.

Another way to do this is to call `AccountSetFlag.forValue` and catch the `IllegalArgumentException` as an indicator that the raw value isn't representable by `AccountSetFlag`

sappenin, line 144:
```suggestion
      // This value will never be negative due to XRPL representing this kind of flag as an unsigned number, 
      // so no lower bound check is required. 
      if (clearFlagRawValue().get().longValue() <= 15) {
```

sappenin (approved):

PR #487, opened 2023-09-28
Fixes #486

### LINKED ISSUE #486: Inconsistent Length of Private Key Byte Array in `deriveKeyPair` Method
### Inconsistent Length of Private Key Byte Array in `deriveKeyPair` Method

I stumbled upon this browsing Github and not understanding how the private key was 33 bytes without a prefix.  

### Description
The `deriveKeyPair` method occasionally produces private keys with lengths of either 32 or 33 bytes. This inconsistency is attributed to the behavior of the `BigInteger.toByteArray()` method. It returns the byte array in two's-complement form and occasionally prepends a zero byte to ensure the number is not negative.

```java
      private static KeyPair deriveKeyPair(final UnsignedByteArray seedBytes, final int accountNumber) {
        Objects.requireNonNull(seedBytes);

        // private key needs to be a BigInteger so we can derive the public key by multiplying G by the private key.
        final BigInteger privateKeyInt = derivePrivateKey(seedBytes, accountNumber);
        final UnsignedByteArray publicKeyInt = derivePublicKey(privateKeyInt);

        return KeyPair.builder()
          .privateKey(PrivateKey.of(UnsignedByteArray.of( /*>>>*/ privateKeyInt.toByteArray() /*<<<*/ )))
          .publicKey(PublicKey.fromBase16EncodedPublicKey(
            UnsignedByteArray.of(publicKeyInt.toByteArray()).hexValue()
          ))
          .build();
      }

```

The length of the byte array `byteArray` is sometimes 33 bytes and sometimes 32 bytes, which could lead to potential issues downstream where a fixed-length byte array might be expected.

### Expected Behavior
Consistent length for the private key's byte array.

### Actual Behavior
Variable length of 32 or 33 bytes for the private key's byte array.

### Why
The inconsistency is due to the two's-complement representation that `BigInteger.toByteArray()` uses. Sometimes a 0 byte is prepended to make sure the number is not interpreted as negative.

### Note
The JavaScript edition of ripple-keypairs uses a 33-byte representation with a 0 byte prefix. The secp256k1 curve generally uses 32-byte scalars naturally.

### Possible Solution
Normalize the private key to a consistent length. Either:

- Consistently use a 33-byte representation with a 0 byte prefix for compatibility with other Ripple products.
- Or normalize to a 32-byte scalar which aligns with the natural length used in secp256k1.

### Fixtures

| Seed Hex                           | Private Key Hex                                                      | Private Key Length |
|------------------------------------|----------------------------------------------------------------------|--------------------|
| `358c75d5ad5e2ff5e19256978f18f0b9` | `00fbd60c9c99ba3d4706a449c30e4d61dcc3811e23ef69291f98a886cec6a8b0b5` | 33                 |
| `b7e99c6bb9786494238d2ba84eabe854` | `7030cbd40d6961e625ad73159a4b463aa42b4e88cc2248ac49e1edcb50af2924`   | 32                 |

sappenin:
1. Add a precondition to PrivateKey.of that enforces the length to 33 bytes (starting with either  0x00 or 0xED).
2. Update BCKeyUtils to add/remove the padding for secp keys (the way we do for Ed keys).
3. Add a new function to PrivateKey called `valueWithoutPrefix` which returns the natural 32 private key bytes.
4. Update Secp256k1KeyPairService in Seed.java to prefix with private keys with 0x00 if only 32 bytes.
5. Validate that `PrivateKey.keyType()` works properly via unit tests.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
@@ -28,7 +28,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
-
 import javax.security.auth.Destroyable;
 
 /**
@@ -116,6 +115,23 @@ public class UnsignedByteArray implements Destroyable {
   }
 
   /**
+   * Construct a new {@link UnsignedByteArray} that contains this byte array's bytes, but with enough `0x00` prefix
+   * padding bytes such that the final length of the returned value is {@code minFinalByteLength}.
+   *
+   * @param minFinalByteLength The minimum length, in bytes, that the final result must be zero-byte prefix-padded to.
+   *                           If this number is greater-than {@code #length}, then this value will be reduced to
+   *                           {@code #length}.
+   *
+   * @return A copy of this {@link UnsignedByteArray} that has been zero-byte prefix-padded such that its final length
+   *   is at least {@code minFinalByteLength}.
+   */
+  public UnsignedByteArray withPrefixPadding(int minFinalByteLength) {
+    Preconditions.checkArgument(minFinalByteLength >= 0, "minFinalByteLength must not be negative");
+
+    return withPrefixPadding(this.toByteArray(), minFinalByteLength);
+  }
+
+  /**
    * Creates a new {@link UnsignedByteArray} from a byte array by prefixing enough zero-pad bytes to ensure that the
    * result has at least {@code minFinalByteLength} bytes.
    *
@@ -178,23 +194,6 @@ public class UnsignedByteArray implements Destroyable {
       unsignedBytes.add(i, UnsignedByte.of(0));
     }
     return unsignedBytes;
-  }
-
-  /**
-   * Construct a new {@link UnsignedByteArray} that contains this byte array's bytes, but with enough `0x00` prefix
-   * padding bytes such that the final length of the returned value is {@code minFinalByteLength}.
-   *
-   * @param minFinalByteLength The minimum length, in bytes, that the final result must be zero-byte prefix-padded to.
-   *                           If this number is greater-than {@code #length}, then this value will be reduced to
-   *                           {@code #length}.
-   *
-   * @return A copy of this {@link UnsignedByteArray} that has been zero-byte prefix-padded such that its final length
-   *   is at least {@code minFinalByteLength}.
-   */
-  public UnsignedByteArray withPrefixPadding(int minFinalByteLength) {
-    Preconditions.checkArgument(minFinalByteLength >= 0, "minFinalByteLength must not be negative");
-
-    return withPrefixPadding(this.toByteArray(), minFinalByteLength);
   }
 
   /**
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+
 import javax.security.auth.Destroyable;
 
 /**
@@ -76,82 +77,6 @@ public class UnsignedByteArray implements Destroyable {
     unsignedBytes.add(first);
     Collections.addAll(unsignedBytes, rest);
     return new UnsignedByteArray(unsignedBytes);
-  }
-
-  /**
-   * Creates an {@link UnsignedByteArray} from the bytes of a supplied {@link BigInteger}. If the length of the
-   * resulting array is not at least {@code minFinalByteLength}, then the result is prefix padded with `0x00` bytes
-   * until the final array length is {@code minFinalByteLength}.
-   *
-   * <p>This function primarily exists to ensure that transformation of secp256k1 private keys from one form to another
-   * (e.g., from {@link BigInteger} to a byte array) are done in a consistent manner, always yielding the desired number
-   * of bytes. For example, secp256k1 private keys are 32-bytes long naturally. However, when transformed to a byte
-   * array via {@link BigInteger#toByteArray()}, the result will not always have the same number of leading zero bytes
-   * that one might expect. Sometimes the returned array will have 33 bytes, one of which is a zero-byte prefix pad that
-   * is meant to ensure the underlying number is not represented as a negative number. Other times, the array will have
-   * fewer than 32 bytes, for example 31 or even 30, if the byte array has redundant leading zero bytes.
-   *
-   * <p>Thus, this function can be used to normalize a byte array with a desired number of 0-byte padding to ensure
-   * that the resulting byte array is always the desired {@code minFinalByteLength} (e.g., in this library, secp256k1
-   * private keys should always be comprised of a 32-byte natural private key with a one-byte `0x00` prefix pad).
-   *
-   * @param amount             A {@link BigInteger} to convert into an {@link UnsignedByteArray}.
-   * @param minFinalByteLength The minimum length, in bytes, that the final result must be. If the final byte length is
-   *                           less than this number, the resulting array will be prefix padded to increase its length
-   *                           to this number.
-   *
-   * @return An {@link UnsignedByteArray} with a length of at least {@code minFinalByteLength}.
-   *
-   * @see "https://github.com/XRPLF/xrpl4j/issues/486"
-   */
-  public static UnsignedByteArray from(final BigInteger amount, int minFinalByteLength) {
-    Objects.requireNonNull(amount);
-    Preconditions.checkArgument(amount.signum() >= 0, "amount must not be negative");
-    Preconditions.checkArgument(minFinalByteLength >= 0, "minFinalByteLength must not be negative");
-
-    // Return the `amount` as an UnsignedByteArray, but with the proper zero-byte prefix padding.
-    return withPrefixPadding(amount.toByteArray(), minFinalByteLength);
-  }
-
-  /**
-   * Construct a new {@link UnsignedByteArray} that contains this byte array's bytes, but with enough `0x00` prefix
-   * padding bytes such that the final length of the returned value is {@code minFinalByteLength}.
-   *
-   * @param minFinalByteLength The minimum length, in bytes, that the final result must be zero-byte prefix-padded to.
-   *                           If this number is greater-than {@code #length}, then this value will be reduced to
-   *                           {@code #length}.
-   *
-   * @return A copy of this {@link UnsignedByteArray} that has been zero-byte prefix-padded such that its final length
-   *   is at least {@code minFinalByteLength}.
-   */
-  public UnsignedByteArray withPrefixPadding(int minFinalByteLength) {
-    Preconditions.checkArgument(minFinalByteLength >= 0, "minFinalByteLength must not be negative");
-
-    return withPrefixPadding(this.toByteArray(), minFinalByteLength);
-  }
-
-  /**
-   * Creates a new {@link UnsignedByteArray} from a byte array by prefixing enough zero-pad bytes to ensure that the
-   * result has at least {@code minFinalByteLength} bytes.
-   *
-   * @param bytes              A byte array to convert into an {@link UnsignedByteArray}.
-   * @param minFinalByteLength The minimum length, in bytes, that the final result must be prefix-padded to. If this
-   *                           number is greater-than {@code amount#length}, then this value will be reduced to
-   *                           {@code amount#length}.
-   *
-   * @return A zero-byte prefix-padded {@link UnsignedByteArray} with a length of at least {@code minFinalByteLength}.
-   */
-  private static UnsignedByteArray withPrefixPadding(final byte[] bytes, int minFinalByteLength) {
-    Preconditions.checkArgument(minFinalByteLength >= 0, "minFinalByteLength must not be negative");
-
-    if (bytes.length > minFinalByteLength) { // <-- Increase `minFinalByteLength` to be at least bytes.length
-      minFinalByteLength = bytes.length;
-    }
-
-    final int numPadBytes = minFinalByteLength - bytes.length; // <-- numPadBytes will never be negative
-    byte[] resultBytes = new byte[minFinalByteLength];
-    System.arraycopy(bytes, 0, resultBytes, numPadBytes, bytes.length);
-    return UnsignedByteArray.of(resultBytes);
   }
 
   /**
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
@@ -86,7 +86,7 @@ public class UnsignedByteArray implements Destroyable {
    * <p>This function primarily exists to ensure that transformation of secp256k1 private keys from one form to another
    * (e.g., from {@link BigInteger} to a byte array) are done in a consistent manner, always yielding the desired number
    * of bytes. For example, secp256k1 private keys are 32-bytes long naturally. However, when transformed to a byte
-   * array via `BigInteger.toByteArray()`, the result will not always have the same number of leading zero bytes that
+   * array via {@link BigInteger#toByteArray()}, the result will not always have the same number of leading zero bytes that
    * one might expect. Sometimes the returned array will have 33 bytes, one of which is a zero-byte prefix pad that is
    * meant to ensure the underlying number is not represented as a negative number. Other times, the array will have
    * fewer than 32 bytes, for example 31 or even 30, if the byte array has redundant leading zero bytes.
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
@@ -28,7 +28,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
-
 import javax.security.auth.Destroyable;
 
 /**
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+
 import javax.security.auth.Destroyable;
 
 /**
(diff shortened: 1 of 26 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
```diff
@@ -386,11 +389,24 @@ private static KeyPair deriveKeyPair(final UnsignedByteArray seedBytes, final in
...
+        // No need to prefix secp256k1 public keys because they are always 33 bytes.
+        final UnsignedByteArray prefixedPublicKey = UnsignedByteArray.of(publicKeyInt.toByteArray());
+
         return KeyPair.builder()
-          .privateKey(PrivateKey.of(UnsignedByteArray.of(privateKeyInt.toByteArray())))
-          .publicKey(PublicKey.fromBase16EncodedPublicKey(
-            UnsignedByteArray.of(publicKeyInt.toByteArray()).hexValue()
-          ))
+          .privateKey(PrivateKey.of(prefixedPrivateKey))
+          .publicKey(PublicKey.fromBase16EncodedPublicKey(prefixedPublicKey.hexValue()))
```
nkramer44, line 409: Should we call `PublicKey.builder().value(prefixedPublicKey).build()` here rather than encode to hex and then parse from hex?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
```diff
@@ -100,8 +100,10 @@ public static PrivateKey toPrivateKey(final Ed25519PrivateKeyParameters ed25519P
   public static PrivateKey toPrivateKey(final ECPrivateKeyParameters ecPrivateKeyParameters) {
     // Convert the HEX representation of the BigInteger into bytes.
     byte[] privateKeyBytes = BaseEncoding.base16().decode(ecPrivateKeyParameters.getD().toString(16).toUpperCase());
-    final UnsignedByteArray privateKeyUba = UnsignedByteArray.of(privateKeyBytes);
-    return PrivateKey.of(privateKeyUba);
+    return PrivateKey.of(
+      UnsignedByteArray.of(PrivateKey.SECP256K1_PREFIX) // <-- Per #486, always append 0x00 to secp256k1 private keys
+        .append(UnsignedByteArray.of(privateKeyBytes))
```
nkramer44, line 105: Are we sure `ecPrivateKeyParameters.getD()` always returns 32 bytes? Is it worth just double checking here so we don't accidentally add a 0x00 prefix to an already prefixed array?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
```diff
@@ -386,11 +389,24 @@ private static KeyPair deriveKeyPair(final UnsignedByteArray seedBytes, final in
         final BigInteger privateKeyInt = derivePrivateKey(seedBytes, accountNumber);
         final UnsignedByteArray publicKeyInt = derivePublicKey(privateKeyInt);
 
+        // Both ed25519 and secp256k1 _private_ keys are always 32 bytes long. However, in this library, both types of
```
sappenin, line 392: This comment is incorrect - fixing it now.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
```diff
@@ -100,8 +100,10 @@ public static PrivateKey toPrivateKey(final Ed25519PrivateKeyParameters ed25519P
   public static PrivateKey toPrivateKey(final ECPrivateKeyParameters ecPrivateKeyParameters) {
     // Convert the HEX representation of the BigInteger into bytes.
     byte[] privateKeyBytes = BaseEncoding.base16().decode(ecPrivateKeyParameters.getD().toString(16).toUpperCase());
-    final UnsignedByteArray privateKeyUba = UnsignedByteArray.of(privateKeyBytes);
-    return PrivateKey.of(privateKeyUba);
+    return PrivateKey.of(
```
nkramer44, line 103: Just to double check -- We don't need to do any prefix stripping in `toEcPrivateKeyParams`?

sappenin: This has been fixed. I'm now padding all public and private key byte arrays to ensure they're always appropriately sized (regardless of what BigInteger's conversion operations yield to us).

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
sappenin: From some cursory tests, it seems to always yield 32 bytes. But just in case I added pad-check code.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/addresses/UnsignedByteArray.java
```diff
@@ -75,6 +78,83 @@ public static UnsignedByteArray of(UnsignedByte first, UnsignedByte... rest) {
...
+   *                           to this number.
+   *
+   * @return An {@link UnsignedByteArray}with a length of at least {@code minFinalByteLength} (possibly via zero-byte
+   *   prefix padding).
+   *
+   * @see "https://github.com/XRPLF/xrpl4j/issues/486"
+   */
+  public static UnsignedByteArray from(final BigInteger amount, int minFinalByteLength) {
+    Objects.requireNonNull(amount);
+    Preconditions.checkArgument(amount.signum() >= 0, "amount must not be negative");
```
sappenin, line 110: TODO: Remove.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
```diff
@@ -32,41 +33,188 @@
...
+    final UnsignedByte prefixByte = value.get(0); // <-- relies upon the above length check.
+    if (ED2559_PREFIX.equals(prefixByte)) {
+      return new PrivateKey(value.slice(1, 33), KeyType.ED25519);
+    } else if (SECP256K1_PREFIX.equals(prefixByte)) {
+      return new PrivateKey(value.slice(1, 33), KeyType.SECP256K1);
+    } else {
+      throw new IllegalArgumentException(String.format(
+        "Constructing a PrivateKey with raw bytes requires a one-byte prefix in front of the 32 natural bytes of a" +
+          " private key. Use the prefix `0xED` for ed25519 private keys, or `0x00` for secp256k1 private keys. " +
+          "Length was %s bytes.", value.length())
```
sappenin, line 105: We should emit the prefix here in addition to the length

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
```diff
@@ -385,12 +386,18 @@ private static KeyPair deriveKeyPair(final UnsignedByteArray seedBytes, final in
         // private key needs to be a BigInteger so we can derive the public key by multiplying G by the private key.
         final BigInteger privateKeyInt = derivePrivateKey(seedBytes, accountNumber);
         final UnsignedByteArray publicKeyInt = derivePublicKey(privateKeyInt);
+        // TODO: FIXME with padding in derivePublicKey
```
sappenin, line 389: TODO: Need to fix this.

sappenin: Fixed via 10e70ecef6b26e0f58860bc9100e43c440ea8572

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
```diff
@@ -234,7 +236,7 @@ public static ECPrivateKeyParameters toEcPrivateKeyParams(final PrivateKey priva
     Preconditions.checkArgument(privateKey.keyType() == KeyType.SECP256K1, "KeyType must be SECP256K1");
 
     final BigInteger privateKeyInt = new BigInteger(
-      BaseEncoding.base16().encode(privateKey.value().toByteArray()), 16
+      BaseEncoding.base16().encode(privateKey.valueWithNaturalBytes().toByteArray()), 16
```
sappenin, line 239: TODO: Check

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
```diff
@@ -97,7 +97,7 @@ protected synchronized Signature edDsaSign(
     final byte[] privateKeyBytes = new byte[32];
     try {
       // Remove ED prefix byte (if it's there)
-      System.arraycopy(privateKey.value().toByteArray(), 1, privateKeyBytes, 0, 32);
+      System.arraycopy(privateKey.valueWithNaturalBytes().toByteArray(), 0, privateKeyBytes, 0, 32);
```
sappenin, line 100: TODO: Do we need `privateKeyBytes` now that we can get the natural bytes?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
```diff
@@ -127,8 +127,7 @@ protected synchronized Signature ecDsaSign(final PrivateKey privateKey, final Un
     Objects.requireNonNull(transactionBytes);
 
     final UnsignedByteArray messageHash = HashingUtils.sha512Half(transactionBytes);
-
-    final BigInteger privateKeyInt = new BigInteger(privateKey.value().toByteArray());
+    final BigInteger privateKeyInt = new BigInteger(privateKey.valueWithPrefixedBytes().toByteArray());
```
sappenin, line 130: TODO: Check

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
```diff
@@ -32,41 +33,189 @@
...
+
+  /**
+   * Accessor for the byte value in {@link #value()} but in a more natural form (i.e., the size of the returned value
+   * will be 32 bytes). Natural ed25519 or secp256k1 private keys will ordinarily contain only 32 bytes. However, in
+   * XRPL, private keys are represented with a single-byte prefix (i.e., `0xED` for ed25519 and `0x00` for secp256k1
+   * keys).
+   *
+   * @return An instance of {@link UnsignedByteArray}.
+   */
+  public UnsignedByteArray valueWithNaturalBytes() {
```
nkramer44, line 177: The name of this method seems off for some reason... what if we just call it `naturalBytes()` or `naturalBytesValue()`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
```diff
@@ -32,41 +33,189 @@
...
+
+  /**
+   * Accessor for the byte value in {@link #value()} but in a more natural form (i.e., the size of the returned value
+   * will be 32 bytes). Natural ed25519 or secp256k1 private keys will ordinarily contain only 32 bytes. However, in
+   * XRPL, private keys are represented with a single-byte prefix (i.e., `0xED` for ed25519 and `0x00` for secp256k1
+   * keys).
+   *
+   * @return An instance of {@link UnsignedByteArray}.
+   */
+  public UnsignedByteArray valueWithPrefixedBytes() {
```
nkramer44, line 196: Same here -- `prefixedBytes()` or `prefixedBytesValue()`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
```diff
@@ -32,41 +33,189 @@
...
+    // Check for empty value, which can occur if this PrivateKey is "destroyed" but still in memory.
+    if (value.length() == 0) {
+      return UnsignedByteArray.empty();
+    } else {
+      // Note: value.slice() will take a view of the existing UBA, then `.toByteArray()` will perform a copy, which is
+      // what we want in order to enforce immutability of this PrivateKey (because Java 8 doesn't support immutable byte
+      // arrays).
```
nkramer44, line 203: I think this comment doesn't apply here anymore. Maybe it should go on line 96?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
nkramer44: I think this should probably be `privateKey.valueWithNaturalBytes`? Probably doesn't matter because the prefix will be 0x00 which won't affect the value of the BigInteger, but un-prefixed seems more correct

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/TestConstants.java
```diff
@@ -36,18 +37,39 @@ public interface TestConstants {
...
+  /**
+   * Helper method to return a newly constructed {@link PrivateKey} with its own copy of bytes.
+   *
+   * @return A {@link PrivateKey}.
+   */
+  static PrivateKey getEcPrivateKey() {
+    return PrivateKey.fromNaturalBytes(
+      UnsignedByteArray.of(BaseEncoding.base16().decode(EC_PRIVATE_KEY_HEX)), KeyType.SECP256K1
+    );
+  }
```
nkramer44, line 72: We should revert this and keep using `EC_PRIVATE_KEY` in our tests to make sure we are always copying and can't mutate PrivateKey bytes.

nkramer44 (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
sappenin: I actually like `naturalBytes`, but if we do that then I suggest we rename the other one to `prefixedBytes`. 

This seems actually preferable overall. One minor nit is developer ergonomics: it won't be possible to type `key.value...` and have auto-complete work.

But that's such a small thing, and perhaps it's better to move away from the `value` nomenclature. WDYT?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
sappenin: Fixed in 901847cb59bc7ecafb03057c7d20ef7d4b3c60a8

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
sappenin: Fixed via 92fb265f95d77220f2c567d242e3de867ad50c0c

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
sappenin: Turns out we can just use `publicKeyInt` directly. Fixed in 68b3292cfab0eb332263f7c40b9a51c44d344b1b

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
sappenin: Fixed via efb49a88bb64ed2967eede944892ffe711fc1124

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
nkramer44: Agree, I like `naturalBytes` and `prefixedBytes`

sappenin: Fixed in 9cf65c4a869426120fa30e7dbf750b6ac659fa75

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PrivateKey.java
sappenin: Fixed via 9cf65c4a869426120fa30e7dbf750b6ac659fa75

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
```diff
@@ -234,7 +237,7 @@ public static ECPrivateKeyParameters toEcPrivateKeyParams(final PrivateKey priva
     Preconditions.checkArgument(privateKey.keyType() == KeyType.SECP256K1, "KeyType must be SECP256K1");
 
     final BigInteger privateKeyInt = new BigInteger(
-      BaseEncoding.base16().encode(privateKey.value().toByteArray()), 16
+      BaseEncoding.base16().encode(privateKey.naturalBytes().toByteArray()), 16
```
sappenin, line 240: Check this

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
```diff
@@ -96,17 +96,14 @@ protected synchronized Signature edDsaSign(
...
     try {
-      // Remove ED prefix byte (if it's there)
-      System.arraycopy(privateKey.value().toByteArray(), 1, privateKeyBytes, 0, 32);
-      Ed25519PrivateKeyParameters privateKeyParameters = new Ed25519PrivateKeyParameters(
-        privateKeyBytes, 0
-      );
+      System.arraycopy(privateKey.naturalBytes().toByteArray(), 0, privateKeyBytes, 0, 32);
+      Ed25519PrivateKeyParameters privateKeyParameters = new Ed25519PrivateKeyParameters(privateKeyBytes, 0);
+
+      final byte[] signableBytes = signableTransactionBytes.toByteArray();
```
sappenin, line 102: check this.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtilsTest.java
```diff
@@ -73,10 +74,10 @@ void ecPrivateKeyParametersToPrivateKeyAndBack() {
 
     // To PrivateKey
     PrivateKey privateKey = BcKeyUtils.toPrivateKey(ecPrivateKeyParameters);
-    assertThat(Base58.encode(privateKey.value().toByteArray()))
-      .isEqualTo("EnYwxojogCYKG3F5Bf7zvcZjo76pEqKwG9wGH14JngcV");
-    assertThat(BaseEncoding.base16().encode(privateKey.value().toByteArray()))
-      .isEqualTo("D12D2FACA9AD92828D89683778CB8DFCCDBD6C9E92F6AB7D6065E8AACC1FF6D6");
+    assertThat(Base58.encode(privateKey.prefixedBytes().toByteArray()))
```
sappenin, line 77: TODO: Check this.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/bc/BcSignatureService.java
sappenin: This is fixed via 408012bffa7f56ffe73cef6b7fd865a84087ec74

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtilsTest.java
sappenin: This is good.

nkramer44 (approved):

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/TestConstants.java
sappenin: @nkramer44 and I decided to keep this code, but add unit tests to validate PrivateKey immutability, which was added via 46db60ceef1788f3e0646cea567eaf1a3298e413.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/bc/BcKeyUtils.java
sappenin: Fixed.

sappenin:
Merging this now that all tests passing except for two subsets of ITs.

- `build_testnet_reporting` is failing only on AMM and Clawback because neither are yet enabled in `TestNet`.
- `build_testnet_clio`  is failing only on AMM and Clawback because neither are yet enabled in `TestNet` and failing on  NFToken because of neither https://github.com/XRPLF/clio/pull/736 nor https://github.com/XRPLF/clio/pull/780 are deployed yet.

PR #489, opened 2023-10-10
Add support for ledger_entry RPC

nkramer44:
Adds support for the `ledger_entry` RPC command in XrplClient.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerEntryResult.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerEntryResult.java
@@ -1,0 +1,126 @@
+package org.xrpl.xrpl4j.model.client.ledger;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+import org.xrpl.xrpl4j.model.client.XrplResult;
+import org.xrpl.xrpl4j.model.client.common.LedgerIndex;
+import org.xrpl.xrpl4j.model.ledger.LedgerObject;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+
+import java.util.Optional;
+
+/**
+ * The result of a {@code ledger_entry} RPC call.
+ *
+ * @param <T> The type of {@link LedgerObject} contained in the result.
+ */
+@Immutable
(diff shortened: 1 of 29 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerEntryResult.java
```diff
@@ -0,0 +1,126 @@
...
+  default boolean validated() {
+    return false;
+  }
+
+  /**
+   * The ID of the ledger entry returned.
+   *
+   * @return The {@link Hash256} representing the ID of the ledger entry.
+   */
+  Hash256 index();
```
sappenin, line 125: Double-checking -- this is never empty, correct?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/TicketObject.java
```diff
@@ -110,4 +110,11 @@ default Flags flags() {
    */
   @JsonProperty("TicketSequence")
   UnsignedInteger ticketSequence();
+
+  /**
+   * The unique ID of the {@link TicketObject}.
+   *
+   * @return A {@link Hash256}.
+   */
+  Hash256 index();
```
sappenin, line 119: Thinking a bit more - should this be pushed up into `LedgerObject`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/ledger/LedgerEntryResult.java
nkramer44: I don't think so. None of the other `LedgerObject`s have it as `Optional` and I've never seen those break during deserialization.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/TicketObject.java
nkramer44: I'm always a bit hesitant to push fields up to a super class in xrpl4j. I _think_ all ledger object responses have this field, but the code in rippled is unclear and who knows if all future ledger objects will have this field in RPC responses. If it turns out one particular `LedgerObject` has an optional `index` field, or it doesn't have an `index` field and we push `index` up to `LedgerObject`, we'd have to break every single `LedgerObject`, whereas if we copy this field in each `LedgerObject` subclass, the damage is limited to just one object. wdyt?

sappenin: Yeah, probably the right way to go -- I suppose for now this is just an implementation detail, and does leave room for something different (as you point out), as long as we don't omit this field by accident on some `LedgerObject` that should have it.

sappenin (approved):

Issue #266, opened 2022-05-29
Publish DevNet/TestNet/MainNet Environment classes so that other projects can use them


sappenin:
The classes in [org.xrpl.xrpl4j.test.environment](https://github.com/XRPLF/xrpl4j/tree/main/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment) would be useful to other projects. We should move these to the `src/test/java` folder (and also a variety of AbstractIT) so that they can be accessed from external projects.

sappenin:
Another feature that should be built as part of this ticket is some kind of map for each environment that allows us to enable/disable tests (e.g., if we specify `xrplcluster.com` then that Environment object could be queried to see if any particular test suite should or should not be executed.

That way, [this build](https://github.com/XRPLF/xrpl4j/actions/runs/6869821469/job/18683404546?pr=494) will "pass" (and not run AMM tests, for example, because xrplcluster.com doesn't yet support that feature).

PR #494, opened 2023-11-14
Point Reporting Mode tests to xrplcluster

nkramer44:
`ReportingModeMainnetEnvironment` was previously pointed at `https://s2-reporting.riple.com` which was recently shut down. `https://xrplcluster.com` is powered by reporting mode servers, so this PR changes `ReportingModeMainnetEnvironment` to point to `xrplcluster.com`


sappenin:
Looks like the Reporting IT failed due to `Clawback` and `AMM`, so that feels like a "passing test" (i.e., shouldn't block this PR). We should probably make a way to twiddle features off for a given environment. I've updated https://github.com/XRPLF/xrpl4j/issues/266 with this idea.

sappenin (approved):

nkramer44:
@sappenin yes, we've been merging PRs that only fail testnet reporting ITs due to clawback and AMMs, and have no problems doing so here. Merging.

PR #498, opened 2023-11-28
DID Support

### LINKED ISSUE #496: Add support for DID
See specification here: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0040d-decentralized-identity

This is deployed in v2.0 rippled versions and above, but is not yet amendment enabled.

nkramer44:
Fixes #496. Adds XLS-40d DID support.

Includes new transactions:
* `DidSet`
* `DidDelete`

Includes new ledger enty:
* `DidObject`


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidDataSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * No-args constructor.
+   */
+  public DidDataSerializer() {
+    super(DidData.class, false);
+  }
+
+  @Override
+  public void serialize(DidData DidData, JsonGenerator gen, SerializerProvider provider) throws IOException {
+    gen.writeString(DidData.value());
```
sappenin, line 44: Should this data actually be hex-encoded? The spec shows this as a `string` in JSON, but in rippled it's just a `BLOB`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidUriDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * No-args constructor.
+   */
+  public DidUriDeserializer() {
+    super(DidUri.class);
+  }
+
+  @Override
+  public DidUri deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+    return DidUri.of(jsonParser.getText());
```
sappenin, line 44: Should this data actually be hex-encoded? The spec shows this as a `string` in JSON, but in rippled it's just a `BLOB`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidUriSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * No-args constructor.
+   */
+  public DidUriSerializer() {
+    super(DidUri.class, false);
+  }
+
+  @Override
+  public void serialize(DidUri didUri, JsonGenerator gen, SerializerProvider provider) throws IOException {
+    gen.writeString(didUri.value());
```
sappenin, line 44: Should this data actually be hex-encoded? The spec shows this as a `string` in JSON, but in rippled it's just a `BLOB`.

xrpl4j-core/src/main/resources/definitions.json
```diff
@@ -140,40 +141,40 @@
...
     ],
     [
       "Transaction",
       {
-        "nth": 1,
+        "nth": 257,
         "isVLEncoded": false,
         "isSerialized": false,
-        "isSigningField": true,
+        "isSigningField": false,
```
sappenin, line 157: Why are these changing?

xrpl4j-core/src/main/resources/definitions.json
```diff
@@ -2759,7 +2781,7 @@
     "tecKILLED": 150,
     "tecHAS_OBLIGATIONS": 151,
     "tecTOO_SOON": 152,
-    "tecHOOK_ERROR": 153,
+    "tecHOOK_REJECTED": 153,
```
sappenin, line 2784: This seems unrelated to DID -- not objecting, just double-checking.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DidIT.java
```diff
@@ -0,0 +1,270 @@
...
+    XrpCurrencyAmount fee = FeeUtils.computeNetworkFees(xrplClient.fee()).recommendedFee();
+
+    DidSet didSet = DidSet.builder()
+      .account(sourceAccountInfo.accountData().account())
+      .sequence(sourceAccountInfo.accountData().sequence())
+      .fee(fee)
+      .lastLedgerSequence(sourceAccountInfo.ledgerIndexSafe().unsignedIntegerValue().plus(UnsignedInteger.valueOf(4)))
+      .signingPublicKey(sourceKeyPair.publicKey())
+      .data(DidData.of("617474657374"))
+      .uri(DidUri.of("6469645F6578616D706C65"))
```
sappenin, line 192: What does rippled do if this isn't HEX?

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidDataSerializer.java
nkramer44: It should be hex encoded, but I'm not currently checking that it is in `DidData`, `DidDocument` or `DidUri`. I could do that, but it adds an extra hex decode, and I've generally stopped adding validation to our models because rippled often breaks those validations. However, if a developer sets one of these fields to a non-hex String and tries to serialize it to sign, the binary codec will throw an error because it's expecting a hex string, so their errors will at least be caught before submitting to rippled.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidUriDeserializer.java
nkramer44: See comment above

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidUriSerializer.java
nkramer44: See comment above

xrpl4j-core/src/main/resources/definitions.json
nkramer44: I generally run [xrpl-codec-gen](https://github.com/RichardAH/xrpl-codec-gen) to generate this file whenever there are new transactions or fields. Mayukha's [recent commit](https://github.com/RichardAH/xrpl-codec-gen/commit/d3691947bfcc334a5c1a9322ad2e53626adcc5ea) changes these fields `isSigningField` value to align with her rippled PR, which explains these changes.

xrpl4j-core/src/main/resources/definitions.json
nkramer44: See comment above -- this was generated by xrpl-codec-gen

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DidIT.java
nkramer44: See comment above -- I can't actually submit a transaction with a non-hex value for any of these fields because the binary codec will throw an exception.

You could talk me into validating that `DidData`, `DidUri`, and `DidDocument` are hex encoded in those objects so developers get better exceptions...

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/DidDataSerializer.java
sappenin: Yes, that's reasonable. I'm not too worried in this particular case because the input on line 44 is going to be coming from rippled, and it _should_ be HEX.

On the constructors of these threee (`DidData.of(...)` it _might_ be worth adding validation that only HEX can be supplied there -- but to you're point, the binary codec will throw. Maybe this relates to https://github.com/XRPLF/xrpl4j/issues/500, and can be fixed there -- but the binary codec errors aren't very helpful.

This isn't a blocker for me here, so if you don't want to do anything for now, I'm good with that.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DidIT.java
sappenin: Maybe make a comment in https://github.com/XRPLF/xrpl4j/issues/500 to improve the binary codec errors.

sappenin (approved):

PR #508, opened 2023-12-05
Support XRPFees amendment

nkramer44:
Adjusts the `SetFee` pseudo-transaction to work when the `XRPFees` amendment goes live.


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,107 @@ static ImmutableSetFee.Builder builder() {
...
    * @return An {@link UnsignedInteger} cost of ref transaction.
+   *
+   * @deprecated Prefer {@link #maybeReferenceFeeUnits()} over this field.
+   */
+  @Deprecated
+  @Nullable
+  @Value.Default
+  @JsonIgnore
+  default UnsignedInteger referenceFeeUnits() {
+    return maybeReferenceFeeUnits().orElse(null);
```
sappenin, line 98: This feels like a breaking change. 

E.g., imagine an application using an older version of xrpl4j  is calling `referenceFeeUnits`. Then, that code upgrades to the version of xrpl4j that has this change (but the application usage of xrpl4j isn't changed because there's just a deprecation warning). If that application starts talking to a server with `XRPFees` amendment enabled, that code will throw a NPE likely.

I need to think more about this one -- unless the argument is that what I wrote above can't happen? Or maybe just is unlikely to happen?

nkramer44: I think you're right. Anyone using this method and not checking for null will get a NPE if XRPFees is enabled. Unfortunately I can't think of a way to support the XRPFees change without making this `@Nullable`. We could `@Value.Default` it to `UnsignedInteger.ZERO` to avoid creating NPE, but that almost seems harder to catch as a developer than getting a bunch of NPE in your logs.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
-   * @return A hex {@link String} basefee value.
+   * @return A hex {@link String} baseFee value.
+   *
+   * @deprecated Prefer {@link #baseFeeDrops()} over this field because the XRPFees ammendment now serializes this
+   *   object's base fee into XRP drops.
    */
-  @JsonProperty("BaseFee")
-  String baseFee();
+  @Value.Default
+  @Deprecated
```
nkramer44, line 66: Remove deprecation

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
    *
    * @return An {@link UnsignedInteger} cost of ref transaction.
+   *
+   * @deprecated Prefer {@link #maybeReferenceFeeUnits()} over this field.
+   */
+  @Deprecated
+  @Nullable
+  @Value.Default
+  @JsonIgnore
+  default UnsignedInteger referenceFeeUnits() {
```
nkramer44, line 98: Remove deprecation, make a breaking change of just making this `Optional`, and get rid of `maybeReferenceFeeUnits`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
   /**
    * The base reserve, in drops.
    *
-   * @return An {@link UnsignedInteger} base reverse value in {@link org.xrpl.xrpl4j.model.client.fees.FeeDrops}.
+   * @return An {@link UnsignedInteger} base reserve value in {@link org.xrpl.xrpl4j.model.client.fees.FeeDrops}.
+   *
+   * @deprecated Prefer {@link #reserveBaseDrops()} over this field.
+   */
+  @Value.Default
+  @Deprecated
```
nkramer44, line 119: Remove deprecation and add comment

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
 
   /**
    * The base reserve, in drops.
    *
-   * @return An {@link UnsignedInteger} base reverse value in {@link org.xrpl.xrpl4j.model.client.fees.FeeDrops}.
+   * @return An {@link UnsignedInteger} base reserve value in {@link org.xrpl.xrpl4j.model.client.fees.FeeDrops}.
+   *
+   * @deprecated Prefer {@link #reserveBaseDrops()} over this field.
+   */
+  @Value.Default
```
nkramer44, line 118: Use derived

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
    *
-   * @return A hex {@link String} basefee value.
+   * @return A hex {@link String} baseFee value.
+   *
+   * @deprecated Prefer {@link #baseFeeDrops()} over this field because the XRPFees ammendment now serializes this
+   *   object's base fee into XRP drops.
    */
-  @JsonProperty("BaseFee")
-  String baseFee();
+  @Value.Default
```
nkramer44, line 65: Use derived

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/SetFee.java
```diff
@@ -53,34 +57,108 @@ static ImmutableSetFee.Builder builder() {
...
   /**
    * The incremental reserve, in drops.
    *
    * @return An {@link UnsignedInteger} incremental reserve in {@link org.xrpl.xrpl4j.model.client.fees.FeeDrops}.
+   *
+   * @deprecated Prefer {@link #reserveIncrementDrops()} over this field.
+   */
+  @Deprecated
+  @JsonIgnore
+  @Value.Default
```
nkramer44, line 146: Use derived

sappenin (approved):
LGTM!

PR #512, opened 2023-12-18
Fix EscrowFinish fulfillment and condition parsing

### LINKED ISSUE #513: LedgerResult Deserialization: Invalid but parseable EscrowFinish fulfillment
Certain fulfillments and conditions that show up in validated (but not executed) EscrowFinish transactions are able to be parsed by `CryptoConditionReader`, but are actually malformed (for example if a Fulfillment has a well formed first 32 bytes, but has extra bytes attached afterward). This causes an exception to be thrown when deserializing the EscrowFinish from JSON. 

Offending transactions:
- https://livenet.xrpl.org/transactions/138543329687544CDAFCD3AB0DCBFE9C4F8E710397747BA7155F19426F493C8D

nkramer44:
Fixes #513 

As seen in transaction [`138543329687544CDAFCD3AB0DCBFE9C4F8E710397747BA7155F19426F493C8D`](https://livenet.xrpl.org/transactions/138543329687544CDAFCD3AB0DCBFE9C4F8E710397747BA7155F19426F493C8D), sometimes `EscrowFinish` transactions can show up in validated ledger transaction sets with an invalid fulfillment or condition. We added a `fulfillmentRawValue` and `conditionRawValue`, and normalization functions to try to make `fulfillment()` equivalent to `fulfillmentRawValue` and `condition()` equivalent to `conditionRawValue` in #483. However, we missed a special case where the fulfillment or condition could be well formed in the first 32 bytes, but contain extra bytes afterward (in the case of 138543329687544CDAFCD3AB0DCBFE9C4F8E710397747BA7155F19426F493C8D, the fulfillment is duplicated), which causes an exception to be thrown by the normalization function.

This PR fixes this issue by checking that the parsed fulfillment or condition is equivalent to the fulfillment/condition raw value. If it is not equivalent, we will not set `fulfillment` or `condition`. If it is, we will.


sappenin (approved):
LGTM!

Issue #527, opened 2024-04-02
xrpl4j fail to get negative balance value


hyeongcheolkim:
xrpl4j fail to get '-2298000050' because type is unsignedlong. but i think it should be signedlong.

    XrplClient xrp = new XrplClient(HttpUrl.get("https://xrplcluster.com/"));
    try {

        ImmutableLedgerRequestParams params = LedgerRequestParams.builder()
                .ledgerSpecifier(LedgerSpecifier.of(UnsignedInteger.valueOf(354575L)))
                .transactions(true)
                .build();
        ImmutableJsonRpcRequest request = JsonRpcRequest.builder().method("ledger").addParams(params).build();
        JsonNode jsonNode = xrp.getJsonRpcClient().postRpcRequest(request);
        System.out.println("jsonNode = " + jsonNode);

        LedgerResult ledger = xrp.ledger(params);
        System.out.println("ledger = " + ledger);

    } catch (Exception e) {
        throw new RuntimeException(e);
    }

when i run this code. exception occured.

>          org.xrpl.xrpl4j.client.JsonRpcClientErrorException: com.fasterxml.jackson.databind.JsonMappingException: value (-2298000050) is outside the range for an unsigned long value (through reference chain: org.xrpl.xrpl4j.model.transactions.ImmutableTransactionMetadata$Json["AffectedNodes"]->java.util.ArrayList[0]->       org.xrpl.xrpl4j.model.transactions.metadata.ImmutableModifiedNode$Json["PreviousFields"]->       org.xrpl.xrpl4j.model.transactions.metadata.ImmutableMetaAccountRootObject$Json["Balance"]) (through reference chain: org.xrpl.xrpl4j.model.client.ledger.ImmutableLedgerResult$Json["ledger"]->       org.xrpl.xrpl4j.model.ledger.ImmutableLedgerHeader$Json["transactions"]->java.util.ArrayList[0])

hyeongcheolkim:
i think balance type should be SigendLong. not UnsignedLong.
![image](https://github.com/XRPLF/xrpl4j/assets/72899681/77f5bcda-7d88-4320-b36f-5bd73f2711c5)

nkramer44:
Hi @hyeongcheolkim, this is a [known issue](https://github.com/XRPLF/xrpl4j/issues/474). There are a few historical (very very old) ledgers that seem to contain negative XRP balances, but the bug that caused that issue in rippled has since been fixed. 

[...]

Issue #475, opened 2023-08-10
LedgerResult Deserialization: Large XRP Amounts in TakerPays/TakerGets


nkramer44:
There are several historical transactions where either the `TakerPays` or `TakerGets` fields have an XRP amount that is greater than the maximum XRP (100,000,000,000,000,000 drops). This causes a deserialization failure because `XrpCurrencyAmount` limits amounts to 100,000,000,000,000,000 drops. We should figure out how these transactions have such large amounts because this should not be allowed by rippled.

Ledgers where this is a problem:
* DB7F3411A5F56726671CDC7E378ABDB852243A6AB3241195841FA278131895FF
* A9E115D8E7399BABCE3A33AB7ACB7E542C170DB0FF2D8D813F64916966DBF596
* 478A2D1485DA131AA9BE11D863B5198A23AFB06EB5026B544A9CF667FB906F3A
* B18B5429A1C8FA5D523DA4B716734885AEA5675C9D99D5083CB7AA2111CD98F5
* 8AC3BEA1A098A65FCAEC246510CC472784F370ECCA368CF1C6E5C3435BBFA48E
* F8DBBD6BAE6E6D9A3734D7DCEFB66985614416C06A5585C82E58A0615BDF52C6
* FB620BD9A9773022FEA5C9114BEA3F155252D405B2EDE3C3F29DB06F3FF6E452
* E43F6FB3946CFB572E9DF822CBD449DB7A3B47A3089F60B8769793F3D6D539AB
* 4944275473A983074C0C35CC83FC2CA41AE7E741209E187C7C2008587F74D035
* 064E75D27F61D2EBB2949E4CAD9C2518C4ACDD0A5528F32F0B20B97B07D2ECAD
* 028A7CE0A27759F02C94A34DE92675CC18A86DD89CCC39588621EE6440B92ED6
* CF682E98547634E9C6D44652BADFC1E60283E5B8CE689EF75EA23C78C35931C4
* 1FC05817B97101DE3967E1B747E9C0E03183058476F2CC09A7FAC06DB16B808A
* C3FC0DAF5E757E1F427FFE5064FDC5285E27079847C6114315BACA694428A672
* 4F5EA977211870A199394E75078671D55304904A4FAB7D9F8D01AD2E3040C085
* 1FB3037BA5DC640CA5472879DF72B51F1223AC1EF48CD5B00E24101D8A3AE91C
[...]

PR #535, opened 2024-05-20
Fixes #474

### LINKED ISSUE #474: LedgerResult Deserialization: Negative Balance in AccountRoot AffectedNode
Some historical transactions have metadata for a change in an `AccountRoot` object where the `Balance` of the account is negative. `AccountRootObject` and `MetaAccountRootObject` `balance` is typed as an `XrpCurrencyAmount` which doesn't allow for negative numbers. We should figure out how this is possible because negative balances shouldn't be allowed.

Ledgers where this is a problem:
* ED4395F2C503963E0E9A37F36F5ACC8A60DD9911A1F9E61F71AF555003C1F53B
* 3FA4E5A1B2DA2F27FD2E3DD3AB7487441F65F4349126F4AFC9F67FD07ECB7D3D
* D840709E9075A5DC5D2C6D7E45728B76102B3A5CD22BCF4793C47959A3ECEC2A
* 3E6571E1869FD143E6558251007F511CDF008EBA62F585E2B3FFD51191ED4B44
* 892B13707B9031258D57ED9F42BAAC2BD63EC553E177F5DE0F63BD1EE124F24A
* BD93D52FDDE2E57825FC038B6CE8B19F22388FC2D96161AB75844A5D793113AE
* 4D08EE303097D13775C22674672527055625E8115635C0A0FF8BA714F9474286
* 024BA9BD6BF8C5289C19F742514168701523975B261198EDC6707B9032B051D1

sappenin:
Fixes #474 by adding the `isNegative` function `CurrencyAmount`, and then properly supporting negative value detection in XRP and IssuedCurrency amounts.

This fix should also fix issue #527.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
@@ -31,6 +31,7 @@ import org.xrpl.xrpl4j.codec.addresses.UnsignedByteArray;
 import org.xrpl.xrpl4j.codec.binary.BinaryCodecObjectMapperFactory;
 import org.xrpl.xrpl4j.codec.binary.math.MathUtils;
 import org.xrpl.xrpl4j.codec.binary.serdes.BinaryParser;
+import org.xrpl.xrpl4j.model.immutables.FluentCompareTo;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -90,12 +91,13 @@ class AmountType extends SerializedType<AmountType> {
     }
     BigDecimal value = new BigDecimal(amount);
     if (!value.equals(BigDecimal.ZERO)) {
+      final FluentCompareTo<BigDecimal> fluentValue = FluentCompareTo.is(value);
       if (value.signum() < 0) { // `value` is negative
-        if (value.compareTo(MAX_NEGATIVE_XRP) > 0 || value.compareTo(MIN_DROPS) < 0) {
+        if (fluentValue.greaterThan(MAX_NEGATIVE_XRP) || fluentValue.lessThan(MIN_DROPS)) {
           throw new IllegalArgumentException(String.format("%s is an illegal amount", amount));
         }
       } else { // `value` is positive
-        if (value.compareTo(MIN_XRP) < 0 || value.compareTo(MAX_DROPS) > 0) {
+        if (fluentValue.lessThan(MIN_XRP) || fluentValue.greaterThan(MAX_DROPS)) {
           throw new IllegalArgumentException(String.format("%s is an illegal amount", amount));
         }
       }
@@ -187,7 +189,7 @@ class AmountType extends SerializedType<AmountType> {
   private UnsignedByteArray getAmountBytes(BigDecimal number) {
     BigInteger paddedNumber = MathUtils.toPaddedBigInteger(number, 16);
     byte[] amountBytes = ByteUtils.toByteArray(paddedNumber, 8);
-    amountBytes[0] |= 0x80;
+    amountBytes[0] |= (byte) 0x80;
     if (number.compareTo(BigDecimal.ZERO) > 0) {
       amountBytes[0] |= 0x40;
     }
@@ -197,8 +199,8 @@ class AmountType extends SerializedType<AmountType> {
       throw new IllegalArgumentException("exponent out of range");
     }
     UnsignedByte exponentByte = UnsignedByte.of(97 + exponent - 15);
-    amountBytes[0] |= exponentByte.asInt() >>> 2;
-    amountBytes[1] |= (exponentByte.asInt() & 0x03) << 6;
+    amountBytes[0] |= (byte) (exponentByte.asInt() >>> 2);
+    amountBytes[1] |= (byte) ((exponentByte.asInt() & 0x03) << 6);
 
     return UnsignedByteArray.of(amountBytes);
   }
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
@@ -158,7 +158,7 @@ class AmountType extends SerializedType<AmountType> {
           UnsignedLong
             .valueOf(isValueNegative ? value.asText().substring(1) : value.asText())
             .toString(16),
-          16 // <-- 64 / 16
+          16 // <-- 64 / 4
         )
       );
       final byte[] rawBytes = number.toByteArray();
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
@@ -92,11 +92,11 @@ class AmountType extends SerializedType<AmountType> {
     if (!value.equals(BigDecimal.ZERO)) {
       if (value.signum() < 0) { // `value` is negative
         if (value.compareTo(MAX_NEGATIVE_XRP) > 0 || value.compareTo(MIN_DROPS) < 0) {
-          throw new IllegalArgumentException(amount + " is an illegal amount");
+          throw new IllegalArgumentException(String.format("%s is an illegal amount", amount));
         }
       } else { // `value` is positive
         if (value.compareTo(MIN_XRP) < 0 || value.compareTo(MAX_DROPS) > 0) {
-          throw new IllegalArgumentException(amount + " is an illegal amount");
+          throw new IllegalArgumentException(String.format("%s is an illegal amount", amount));
         }
       }
     }
@@ -181,9 +181,7 @@ class AmountType extends SerializedType<AmountType> {
     result.append(currency);
     result.append(issuer);
 
-    return new
-
-      AmountType(result);
+    return new AmountType(result);
   }
 
   private UnsignedByteArray getAmountBytes(BigDecimal number) {
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
@@ -9,9 +9,9 @@ package org.xrpl.xrpl4j.codec.binary.types;
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -41,7 +41,9 @@ import java.math.BigInteger;
 class AmountType extends SerializedType<AmountType> {
 
   public static final BigDecimal MAX_DROPS = new BigDecimal("1e17");
+  public static final BigDecimal MIN_DROPS = new BigDecimal("-1e17");
   public static final BigDecimal MIN_XRP = new BigDecimal("1e-6");
+  public static final BigDecimal MAX_NEGATIVE_XRP = new BigDecimal("-1e-6");
 
   public static final String DEFAULT_AMOUNT_HEX = "4000000000000000";
   public static final String ZERO_CURRENCY_AMOUNT_HEX = "8000000000000000";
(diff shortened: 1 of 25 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -199,11 +228,44 @@ public static XrpCurrencyAmount ofDrops(UnsignedLong drops) {
...
-      if (FluentCompareTo.is(amount).notEqualTo(BigDecimal.ZERO)) {
-        Preconditions.checkArgument(FluentCompareTo.is(amount).greaterThanEqualTo(SMALLEST_XRP));
+    public static XrpCurrencyAmount ofXrp(final BigDecimal amount) {
+      Objects.requireNonNull(amount);
+
+      if (FluentCompareTo.is(amount).equalTo(BigDecimal.ZERO)) {
+        return ofDrops(UnsignedLong.ZERO);
+      }
+
+      // Whether positive or negative, clamp the amount to 0 if it's too small, or 100B if it's too big.
```
sappenin, line 238: @nkramer44 I need to fix this comment (the code isn't clamping right now, but is instead enforcing preconditions like the old code. I _think_ that's the right choice, in order to keep things functioning "as-is" but we've been bit so many times by invalid assumptions around numbers, I'll pose the question here.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -247,12 +322,16 @@ public XrpCurrencyAmount minus(XrpCurrencyAmount other) {
...
     public XrpCurrencyAmount times(XrpCurrencyAmount other) {
-      return XrpCurrencyAmount.of(this.value().times(other.value()));
+      return XrpCurrencyAmount.ofDrops(this.value().times(other.value()), this.isNegative() | other.isNegative());
     }
 
     @Override
     public String toString() {
-      return this.value().toString();
+      if (isNegative()) {
+        return "-" + this.value().toString();
```
sappenin, line 331: @nkramer44 WDYT about this one? In this case, I do this because `XrpCurrencyAmount` always holds an `UnsignedLong` (the `isNegative()` function is use to detect negativity since the Unsigned value can't hold a negative.

One could argue that this `toString` should always show a positive value, to conform with the `UnsignedLong` underlying.

Note: The details informing this choice are subtly different in the case of `IssueCurrencyAmount`.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmountTest.java
```diff
@@ -0,0 +1,51 @@
...
+  void toStringIssuedCurrentAmount() {
+    // Negative values.
+    IssuedCurrencyAmount issuedCurrency = IssuedCurrencyAmount.builder()
+      .currency("USD")
+      .issuer(Address.of("rP9JR5JTEqaVYbXHtiqR5YvBeoWQeMBipS"))
+      .value("-1.00")
+      .build();
+    assertThat(issuedCurrency.toString()).isEqualTo(
+      "IssuedCurrencyAmount{" +
+        "value=-1.00, " +
```
sappenin, line 22: See my comment above about whether this should emit as a negative string, or a positive one. In rippled JSON RPC responses, these emit as negative values (there is no `isNegative` in there) so from that perspective, this format is more aligned. However, it does feel weird to have a negative number with an `isNegative` property.

Maybe we exclude `isNegative` from the JSON of `IssuedCurrencyAmount` and that will make things clearer.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
nkramer44: I think leaving preconditions here is ok for one main reason, which is that this method is never invoked during serialization, and will only be invoked by a developer. If, for some reason, they need to create an `XrpCurrencyAmount` that is smaller than the smallest XRP or larger than the largest XRP amount, they can just call `ofDrops` or `of`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -213,8 +275,13 @@ public static XrpCurrencyAmount ofXrp(BigDecimal amount) {
      * @return A {@link BigDecimal} representing this value denominated in whole XRP units.
      */
     public BigDecimal toXrp() {
-      return new BigDecimal(this.value().bigIntegerValue())
```
nkramer44, line 216: Was this a bug? If `value()` is drops, we should have been scaling by 10^-6, right?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
nkramer44: I think this is the right thing to do. If you deserialize one of those payloads that has a negative XRP amount and then call `toString` on that amount, it would be misleading/incorrect to _not_ show a `-` before the number IMO

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmountTest.java
nkramer44: I wonder if `isNegative` should live in `CurrencyAmount` or if it should just be a property of `XrpCurrencyAmount`...

If we want to keep it in `CurrencyAmount`, I think it's fine to include it in the `toString` implementation.

nkramer44 (approved):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: No, this was working normally before -- it's just difficult to see the change-set here in GH. Line 224/286 right below is unchanged, so the scaling was happening there (and still is).

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: I _think_ you're right. One final counter-point here would be to emit something like `1 XRP [positive]` or `1 XRP [negative]` so that nobody accidentally uses this `toString` to show or parse real numeric values.

Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: I'm going to capture this idea for future consideration in https://github.com/XRPLF/xrpl4j/issues/475 to unblock this PR (Leaving  unblock this PR (we can consider this change as part of that ticket's work).

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: Resolving this one for now - feel free to re-open if I missed something.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: Resolving this for now in the interest of maintaining something as-close-as-possible to existing functionality.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmountTest.java
sappenin: Agreed. I lean slightly towards leaving `isNegative` in the top-level interface, so resolving this one for now. I created #547 for later tracking.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -88,8 +90,14 @@ private static void assertXrpIsValid(String amount) {
     }
     BigDecimal value = new BigDecimal(amount);
     if (!value.equals(BigDecimal.ZERO)) {
-      if (value.compareTo(MIN_XRP) < 0 || value.compareTo(MAX_DROPS) > 0) {
-        throw new IllegalArgumentException(amount + " is an illegal amount");
+      if (value.signum() < 0) { // `value` is negative
+        if (value.compareTo(MAX_NEGATIVE_XRP) > 0 || value.compareTo(MIN_DROPS) < 0) {
```
sappenin, line 94: @nkramer44 - validate my logic here for handling negative numbers.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -142,14 +150,21 @@ public AmountType fromJson(JsonNode value) throws JsonProcessingException {
...
-          64 / 4
+          UnsignedLong
+            .valueOf(isValueNegative ? value.asText().substring(1) : value.asText())
+            .toString(16),
+          16 // <-- 64 / 16
         )
       );
-      byte[] rawBytes = number.toByteArray();
-      rawBytes[0] |= 0x40;
+      final byte[] rawBytes = number.toByteArray();
```
sappenin, line 162: @nkramer44 - validate my logic here for handling negative numbers.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -142,14 +150,21 @@ public AmountType fromJson(JsonNode value) throws JsonProcessingException {
     if (value.isValueNode()) {
       assertXrpIsValid(value.asText());
 
-      UnsignedByteArray number = UnsignedByteArray.fromHex(
+      final boolean isValueNegative = value.asText().startsWith("-");
```
sappenin, line 153: @nkramer44 - validate my logic here for handling negative numbers.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -142,14 +150,21 @@ public AmountType fromJson(JsonNode value) throws JsonProcessingException {
...
+      final UnsignedByteArray number = UnsignedByteArray.fromHex(
         ByteUtils.padded(
-          UnsignedLong.valueOf(value.asText()).toString(16),
-          64 / 4
+          UnsignedLong
+            .valueOf(isValueNegative ? value.asText().substring(1) : value.asText())
+            .toString(16),
+          16 // <-- 64 / 16
```
nkramer44, line 159:
```suggestion
          16 // <-- 64 / 4
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
```diff
@@ -142,14 +150,21 @@ public AmountType fromJson(JsonNode value) throws JsonProcessingException {
...
         )
       );
-      byte[] rawBytes = number.toByteArray();
-      rawBytes[0] |= 0x40;
+      final byte[] rawBytes = number.toByteArray();
+      if (isValueNegative) {
+        rawBytes[0] |= 0x00;
```
nkramer44, line 164: Isn't `rawBytes[0] |= 0x00` the same as just `rawBytes[0]`? Wondering if we should just do 
```
if (!isValueNegative) {
  rawBytes[0] |= 0x40;
}
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
sappenin: Great catch! Updated.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/AmountType.java
sappenin: Great catch! Updated.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -206,11 +236,50 @@ public static XrpCurrencyAmount ofDrops(UnsignedLong drops) {
...
+     * construct an {@link XrpCurrencyAmount} correctly (i.e., we assume that no developer is constructing a negative
+     * XRP amount because the {@link UnsignedLong} precondition in any legacy code would not allow them to do such a
+     * thing without throwing an exception). Finally, due to the way this class merely adds new static builders to
+     * augment existing code, legacy code should continue to work normally.
+     *
+     * @return {@code true} if this amount is negative; {@code false} otherwise (i.e., if the value is 0 or positive).
+     */
+    @Default
+    @Override
+    @JsonIgnore // <-- This is not actually part of the binary serialization format, so exclude from JSON
```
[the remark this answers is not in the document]
sappenin: Agreed. Fixed via 44ea58f76833ee1ac0153617e61ba3064fef6af3

PR #550, opened 2024-09-13
Make XrplClient timeouts configurable

nkramer44:
Allows developers to set connect and read timeouts on `XrplClients`.

### DIFF
--- a/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
+++ b/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
@@ -28,6 +28,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Range;
 import com.google.common.primitives.UnsignedInteger;
 import com.google.common.primitives.UnsignedLong;
+import feign.Request;
+import feign.Request.Options;
 import okhttp3.HttpUrl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


theotherian (approved):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -127,6 +130,30 @@ public XrplClient(final HttpUrl rippledUrl) {
...
+   *
+   * @param rippledUrl         The {@link HttpUrl} of the node to connect to.
+   * @param connectTimeout     A {@code long} indicating the client's connect timeout.
+   * @param connectTimeoutUnit The {@link TimeUnit} that {@code connectTimeout} is denominated in.
+   * @param readTimeout        A {@code long} indicating the client's read timeout.
+   * @param readTimeoutUnit    The {@link TimeUnit} that {@code readTimeout} is denominated in.
+   */
+  public XrplClient(
+    HttpUrl rippledUrl,
+    long connectTimeout,
```
sappenin, line 144: Any reason not use two `java.time.Duration` parameters here instead of the current four?

If that's easy, I prefer it. However, we could always add an override constructor later if you want this one to merge quickly (esp since this class is marked `@Beta`).

sappenin (approved):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
sappenin: This is resolved.

PR #564, opened 2024-10-29
Serialize non-absent AccountSet `domain` values

nkramer44:
To unset the `Domain` field on an `AccountRoot`, you have to set `AccountSet.domain` to an empty String. In xrpl4j, the Jackson ObjectMapper is configured globally to only include NON_EMPTY fields, which means that `AccountSet.domain` wasn't getting serialized to JSON if it was an empty string, which means that it didn't show up in the binary serialization of the transaction. Therefore, when developers correctly set `AccountSet.domain` to an empty string to clear the AccountRoot field, the transaction would end up being a no-op because `Domain` wasn't present in the serialized transaction.


sappenin (approved):
LGTM!

PR #596, opened 2025-02-17
Draft: Add support for AMM Clawback

agutierrez0:
Opened a Draft MR to see if I am doing something wrong with how I am setting up the integration tests using AMMClawback.

I saw there was a `createAMM` method, but I noticed in the AMMClawback docs that a flag (Allow Trustline Clawback) needed to be enabled in order to use this new transaction. In my integration test, I do the following:

1. enable rippling
2. enable this Allow Trustline Clawback flag
3. create a trustline from trader wallet to issuer wallet for "TST"
4. create a trustline from trader wallet to issuer wallet for "ZFT"
5. send "ZFT" from issuer wallet to trader wallet
6. send "TST" from issuer wallet to trader wallet
7. create an AMMCreate transaction with `amount` being `100` "ZFT", and `amount2` being `250` "TST"

i get back:
```
org.opentest4j.AssertionFailedError: 
expected: "tesSUCCESS"
 but was: "tecNO_PERMISSION"
Expected :"tesSUCCESS"
Actual   :"tecNO_PERMISSION"
```

I see that the error code mentions the following:
```
tecNO_PERMISSION | At least one of the deposit assets cannot be used in an AMM.
```

i have tried a few different currency options, but no luck

i have also been able to do this same transaction with the xrpl.js library using all the same values, except what gets autofilled from their `submitAndWait` function. Any idea what I could be doing wrong?

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/AmmClawbackTransactionFlags.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/AmmClawbackTransactionFlags.java
@@ -1,0 +1,113 @@
+package org.xrpl.xrpl4j.model.flags;
+
+/*-
+ * ========================LICENSE_START=================================
+ * xrpl4j :: core
+ * %%
+ * Copyright (C) 2020 - 2023 XRPL Foundation and its contributors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =========================LICENSE_END==================================
+ */
+
+import org.xrpl.xrpl4j.model.transactions.AmmClawback;
+
+/**
+ * {@link TransactionFlags} for {@link AmmClawback} transactions.
+ */
+public class AmmClawbackTransactionFlags extends TransactionFlags {
+  /**
+   * Constant for an unset flag.
+   */
+  protected static final AmmClawbackTransactionFlags UNSET = new AmmClawbackTransactionFlags(0);
+
+  /**
+   * Constant for the {@code tfClawTwoAssets} flag.
+   */
+  protected static final AmmClawbackTransactionFlags CLAW_TWO_ASSETS = new AmmClawbackTransactionFlags(0x00000001);
+
+  private AmmClawbackTransactionFlags(long value) {
+    super(value);
+  }
+
+  private AmmClawbackTransactionFlags() {
+  }
+
+  /**
+   * Construct {@link AmmClawbackTransactionFlags} with a given value.
+   *
+   * @param tfFullyCanonicalSig The long-number encoded flags value of this {@link AmmClawbackTransactionFlags}.
+   *
+   * @return New {@link AmmClawbackTransactionFlags}.
+   */
+  public static AmmClawbackTransactionFlags of(boolean tfFullyCanonicalSig) {
+    return new AmmClawbackTransactionFlags(
+        Flags.of(
+            tfFullyCanonicalSig ? CLAW_TWO_ASSETS : UNSET
+        ).getValue()
+    );
+  }
+
+  /**
+   * Construct an empty instance of {@link AmmClawbackTransactionFlags}.
+   *
+   * @return An empty {@link AmmClawbackTransactionFlags}.
+   */
+  public static AmmClawbackTransactionFlags empty() {
+    return new AmmClawbackTransactionFlags();
+  }
+
+  /**
+   * Create a new {@link Builder}.
+   *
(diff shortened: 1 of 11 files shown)


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/environment/RippledContainer.java
```diff
@@ -69,7 +69,7 @@ public class RippledContainer {
    * No-args constructor.
    */
   public RippledContainer() {
-    rippledContainer = new GenericContainer<>("rippleci/rippled:2.2.0-rc3")
+    rippledContainer = new GenericContainer<>("rippleci/rippled:2.3.0")
```
agutierrez0, line 72: I noticed when I first tried submitting an AMMClawback transaction, I kept getting some "Unknown field" error, so I changed the version here. Let me know if there is a better version to use

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AmmIT.java
```diff
@@ -288,6 +292,170 @@ void depositAndWithdraw() throws JsonRpcClientErrorException, JsonProcessingExce
...
+        )
+        .tradingFee(TradingFee.ofPercent(BigDecimal.ONE))
+        .lastLedgerSequence(issuerAccount.ledgerIndexSafe().plus(UnsignedInteger.valueOf(100)).unsignedIntegerValue())
+        .signingPublicKey(traderKeyPair.publicKey())
+        .build();
+
+    // cannot get a successful AMMCreate transaction submitted, keep getting tecNO_PERMISSION
+    SingleSignedTransaction<AmmCreate> signedCreate = signatureService.sign(traderKeyPair.privateKey(), ammCreate);
+    SubmitResult<AmmCreate> submitResult = xrplClient.submit(signedCreate);
+    assertThat(submitResult.engineResult()).isEqualTo(TransactionResultCodes.TES_SUCCESS);
```
agutierrez0, line 456: This is where I keep getting `tecNO_PERMISSION`. I am able to successfully get an AMMCreate transaction submitted if I call the `ammCreate` function before doing enabling the allow trust line clawback flag.

agutierrez0: @nkramer44  helped me identify that this was happening because we did not have the AMMClawback Amendment enabled. i updated `rippled.cfg` to add `AMMClawback` and I am no longer getting this error.

agutierrez0:
will close this PR for now and re-open once one ready or if i run into some other issue.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/AmmClawbackTransactionFlags.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * Construct {@link AmmClawbackTransactionFlags} with a given value.
+   *
+   * @param tfFullyCanonicalSig The long-number encoded flags value of this {@link AmmClawbackTransactionFlags}.
+   *
+   * @return New {@link AmmClawbackTransactionFlags}.
+   */
+  public static AmmClawbackTransactionFlags of(boolean tfFullyCanonicalSig) {
```
nkramer44, line 53: This parameter should be renamed to tfClawTwoAssets

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AmmClawback.java
```diff
@@ -0,0 +1,37 @@
...
+  @JsonProperty("Asset")
+  Issue asset();
+
+  @JsonProperty("Asset2")
+  Issue asset2();
+
+  @JsonProperty("Amount")
+  Optional<CurrencyAmount> amount();
+
+  static ImmutableAmmClawback.Builder builder() {
```
nkramer44, line 34: builder() should go at the top of this interface

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AmmClawback.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Optional;
+
+/**
+ * Object mapping for the {@code AMMClawback} transaction.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableAmmClawback.class)
+@JsonDeserialize(as = ImmutableAmmClawback.class)
+@Beta
+public interface AmmClawback extends Transaction {
```
nkramer44, line 20: You'll also need to define the `flags` field, like this:

```
@JsonProperty("Flags")
@Value.Default
default AmmClawbackTransactionFlags flags() {
  return AmmClawbackTransactionFlags.empty();
}
```

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtilsTest.java
```diff
@@ -843,6 +844,33 @@ void addSignatureToAmmCreate() {
     addSignatureToTransactionHelper(ammCreate);
   }
 
+  @Test
+  void addSignatureToAmmClawback() {
```
nkramer44, line 848: Also need a test for `signatureUtils.addMultiSignaturesToTransaction`

nkramer44 (changes requested):
Looking good so far!

PR #637, opened 2025-10-08
Support frozenBalances in gateway_balances result

nkramer44:
There's an undocumented field `frozen_balances` in the response to `gateway_balances` that contains the balances of all frozen trustlines with the given account. Adding support for that field in this PR.


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/accounts/GatewayBalancesResult.java
```diff
@@ -92,6 +92,18 @@ default GatewayBalancesObligations obligations() {
     return GatewayBalancesObligations.builder().build();
   }
 
+  /**
+   * Amounts issued to addresses where the trustline is frozen. Note that any counterparties to a frozen trustline
+   * included in {@code hotwallet} are not included in this field's balances.
```
sappenin:
```suggestion
   * included in {@link GatewayBalancesRequestParams#hotWallets()} are not included in this field's balances.
```

sappenin (changes requested):

PR #674, opened 2026-01-27
Fixes #666: Update `definitions.json` with new values

### LINKED ISSUE #666: Update definitions.json with latest values from Mainnet
This will pave the way for supporting Batch in xrpl4j.

We'll want to add a README into the folder to explain how to generate the definitions. Something like:

```
You can query any running rippled node (including public ones) for its definitions via the server_definitions command.

```bash
curl -X POST -d '{
    "method": "server_definitions",
    "params": [{}]
}' https://s1.ripple.com:51234/
```

Note: The output from the RPC call includes a "result" wrapper, so you would need to strip that out to get the raw JSON
used by the codec.

sappenin:
* Update definitions.json with the latest values from XRPL Mainnet.
* Add a README to describe how to get updated values for `definitions.json`

### DIFF
--- a/xrpl4j-core/src/main/resources/README.md
+++ b/xrpl4j-core/src/main/resources/README.md
@@ -22,7 +22,7 @@ If you have a local build of the `xrpld` C++ server, you can export the definiti
 
 If you don't have a local build of the `xrpld` C++ server, you can use the `xrpl.js` library (specifically the
 ripple-binary-codec package) to fetch the definitions. Read more about this option in the
-[xrpl.js documentation](https://github.com/ripple/xrpl.js/blob/main/packages/ripple-binary-codec/README.md#updating-definitions).
+[xrpl.js documentation](https://github.com/XRPLF/xrpl.js/blob/main/CONTRIBUTING.md#updating-definitionsjson).
 
 ### Option 3: Using a Live `rippled` Node
 


xrpl4j-core/src/main/resources/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Option 1: Using the `rippled` Binary
+
+If you have a local build of the `xrpld` C++ server, you can export the definitions directly from the source code.
+
+```bash
+# Export definitions to a local file
+./rippled --get_definitions > definitions.json
+```
+
+### Option 2: Using the `xrpl.js` Library
```
Patel-Raj11, line 21: I think we don't have `update-definitions` script in xrpl.js. The steps we follow in xrpl.js are mentioned here - https://github.com/XRPLF/xrpl.js/blob/4716780a1fe3df0ff1f255ef23072b22ed06b03e/CONTRIBUTING.md?plain=1#L167

xrpl4j-core/src/main/resources/README.md
```diff
@@ -0,0 +1,46 @@
+# XRPL `definitions.json` Generation Guide
```
Patel-Raj11, line 1: For my understanding: The `definitions.json` file that gets merged into main branch of xrpl4j would always be generated from rippled's `develop` branch right? Since, sometimes xrpl4j supports a feature that is on `Devnet` but not on `Mainnet`.

sappenin: That's a great point - let me update things here.

sappenin: So, it turns out that most of the changes that one might have thought existed on develop are already out in production as part of the v3.0 rippled release. Here's the diff between that and the `develop` branch of xrpld: 51bcbad12250fc7727703ba601e6e0b8fa1b20a8

xrpl4j-core/src/main/resources/README.md
sappenin: Thanks for clarifying that - I updated this file to instead just refer to the xrpl.js docs. See 14722b7388be700c4f28cc1612811999e6159172

xrpl4j-core/src/main/resources/README.md
sappenin: I'm just thinking - maybe as a sanity check, could you generate a definitions.json and compare it to this one (just to make sure I've got the correct one)?

Patel-Raj11: Yes, I checked, the contents are same when I generate definitions.json from `develop` branch using xrpl.js.

xrpl4j-core/src/main/resources/README.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+# Export definitions to a local file
+./rippled --get_definitions > definitions.json
+```
+
+### Option 2: Using the `xrpl.js` Library
+
+If you don't have a local build of the `xrpld` C++ server, you can use the `xrpl.js` library (specifically the
+ripple-binary-codec package) to fetch the definitions. Read more about this option in the
+[xrpl.js documentation](https://github.com/ripple/xrpl.js/blob/main/packages/ripple-binary-codec/README.md#updating-definitions).
```
Patel-Raj11:
```suggestion
[xrpl.js documentation](https://github.com/XRPLF/xrpl.js/blob/main/CONTRIBUTING.md#updating-definitionsjson).
```

Patel-Raj11 (approved):

PR #680, opened 2026-02-01
Fix build error Java 25

### LINKED ISSUE #679: Build Error Java 25
When building xrpl4j-core using Jdk 25, Immutable classes are not generated.

This behaviour is known from Immutables lib team and is documented (see Maven Classic in https://immutables.github.io/getstarted.html)

By applying their fix (adding compiler args in maven-compiler-plugin) it works.

Sadly by doing this, it doesn't work anymore using Java 8. I didn't opened a PR for this reason, this is not a proper fix.

VLEFF:
Fixes #679 by using Maven Modern Immutable configuration defined in https://immutables.github.io/getstarted.html

Also fixes a little javadoc error that is blocking build using java 25

Built and tested using both jdk 25 and jdk 8 : Unit Test are all green

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -60,7 +60,7 @@ jobs:
     strategy:
       matrix:
         # test against each major Java version (Java 8 built separately above for codecov upload)
-        java: [ 11, 16, 17, 21, 25 ]
+        java: [ 11, 16, 17, 21 ]
     steps:
       # Checks-out the repository under $GITHUB_WORKSPACE
       - uses: actions/checkout@v5


pom.xml
sappenin: This change I'm proposing is not strictly necessary, but we might as well upgrade since we're touching the config. `3.15.0` is still Java8 compatible.

```suggestion
          <version>3.15.0</version>
```

pom.xml
```diff
@@ -135,7 +135,7 @@
       </dependency>
       <dependency>
         <groupId>org.immutables</groupId>
-        <artifactId>value</artifactId>
+        <artifactId>value-annotations</artifactId>
         <version>2.9.3</version>
         <scope>provided</scope>
```
sappenin: Just echoing the docs from Immutables. `provided` has its meaning, but I think the extra comment is useful, for clarity.

```suggestion
        <scope>provided</scope> <!-- compile only -->
```

xrpl4j-client/pom.xml
```diff
@@ -47,7 +47,7 @@
     </dependency>
     <dependency>
       <groupId>org.immutables</groupId>
-      <artifactId>value</artifactId>
+      <artifactId>value-annotations</artifactId>
```
sappenin, line 50: You need to sign your commits in order for the build to pass. 
Read more here:

* https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
* https://stackoverflow.com/questions/37737096/signing-an-existing-commit-with-gpg

VLEFF: Done!

sappenin (approved):

sappenin (approved):

VLEFF:
Maybe it's due to the addition of 25 in the build that is breaking everything?
I don't see "build_jdk_temurin_other (25)" in the expected checks

sappenin (approved):

sappenin:
IDK, something is broken in Github for this PR (or maybe something weird with the branch or unsigned commits?). There were some issues with Github actions earlier this week, too, which might be making GH act oddly here? 

I hate to make you do extra work, but I've scoured the settings in the repo around outside contributors and GH actions, and they all look correct. So,  any chance you'd be open to creating a new branch and new PR to see if that can get GH actions working?

Otherwise, I can try this on my end and submit a PR for you tomorrow or Friday.

(Also open to some other ideas?)

VLEFF:
Well it's really the first time I'm dealing with github Action (I'm used to gitlab CI though) so I can't really tell you much.
No problem for creating new branch and PR.

PR #684, opened 2026-02-03
Fixes #683 Allow empty PublicKey

### LINKED ISSUE #683: Add Empty `PublicKey`
Per this comment, it would be nice to be able to construct an empty `PublicKey` so that if a user specifies this in an inner transaction (or potentially in a multisig transactions), the builder for `PublicKey` won't throw when `PublicKey.builder().build()` is called.

The comment in that issue focuses on using an empty `PublicKey` when constructing `Payment` transactions to be used as an inner transaction in a Batch transaction. However, this issue is likely more generally applicable to any Batch inner transaction where the `SigningPublicKey` should be an empty string.

Ordinarily, when constructing transactions like this, the `SigningPublicKey` should simply be omitted in the builder, as can be seen in the Batch ITs [here](https://github.com/XRPLF/xrpl4j/blob/df/batch/add-its/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/BatchTransactionIT.java#L1176). This works because the `Transaction` super-class has an Immutables `@Default` method that sets the `SigningPublicKey` to an empty string when not specified in a builder.

That said, It would be nice to be able to specify an empty PublicKey, as described in the comment. At the least, the builder should not throw when constructed with default values.

sappenin:
* Allow an empty PublicKey to be built using an empty builder
* Add unit tests to verify functionality.


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/PublicKey.java
```diff
@@ -118,7 +121,10 @@ static PublicKey fromBase16EncodedPublicKey(final String base16EncodedPublicKey)
    *
    * @return An instance of {@link UnsignedByteArray}.
    */
-  UnsignedByteArray value();
+  @Default
```
sappenin, line 124: After considering this further, we should remove the empty builder so that developers don't accidentally omit an actual `value`. Instead, we should update `EMPTY_PUBLIC_KEY` to be this:

```java
PublicKey EMPTY_PUBLIC_KEY = PublicKey.builder().value(UnsignedByteArray.empty()).build();
```

That way, if someone wants to overtly use an empty public key, it will be obvious they are do as much.

Issue #671, opened 2026-01-27
Upgrade Jackson ObjectMapper Factory configuration


sappenin:
Issue #656 introduced a more recent Jackson dependency (`v2.21`). As of that version, certain functions in the ObjectMapper configuration are now deprecated, and there is updated guidance for how to configure the ObjectMapper. 

At some point we should consider updating to the new configuration style. However, this will require an xrpl4j major version bump, so we should consider not addressing this issue for awhile because some users of the library may prefer to use xrpl4j with older Jackson library versions, and this new configuration won't work with those older versions.

Some options to consider for addressing this issue:

1. Use the current (legacy) configuration as a fallback if an Exception is thrown during configuration (e.g., due to an enum, function or class not being found). This should work for all library users, but will produce different output, so diagnosing production issues might more difficult. If we stay with the "deprecated" ObjectMapper methods (like registerModule), those methods still exist in 2.21 for compatibility. This allows the same library code to run on both 2.14 and 2.21.

2. If xrpl4j absolutely needs 2.21 features but users are stuck on an older version like v2.14, we can shade and relocate Jackson. This might rename `com.fasterxml.jackson` to `org.xrpl.shaded.jackson`, which would fix the conflict but increases the JAR size and make it harder (though not impossible) for users to customize the ObjectMapper.

3. Consider a library-wide mechanism for setting the `ObjectMapper` used by the library. In this way, the configuration can be supplied by the library user. The risk here is that the configuration might be incorrect, since it's supplied by the user.

5. Delay this update for some time (Possibly to when we update to Jackson v3) to give Jackson a chance to provide a configuration parameter to control this behavior in a more predictable fashion (e.g., see https://github.com/FasterXML/jackson-databind/issues/3284).

It's likely that option 1 or 2 are the most pragmatic choice.

Following is an updated configuration for `ObjectMapperFactory` that conforms to Jackson `v2.21`: 

```java
public class ObjectMapperFactory {

  /**
   * Construct an {@link ObjectMapper} that can be used to serialize and deserialize JSON.
   *
   * @return An {@link ObjectMapper}.
   */
  public static ObjectMapper create() {

    JavaTimeModule javaTimeModule = new JavaTimeModule();

    // 1. Define the exact pattern that rippled produces for date/times (also that unit tests expect (+0000))
    // 'X' is the ISO Zone Offset. 'XXXX' specifically produces '+0000'
[...]

PR #689, opened 2026-02-05
Added account permission delegation

cybele-ripple:
Adds support for account permission delegation ([XLS-0075](https://xls.xrpl.org/xls/XLS-0075-permission-delegation.html))


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+/**
+ * {@link Permission} inner object with PermissionValue details.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutablePermission.class)
+@JsonDeserialize(as = ImmutablePermission.class)
+public interface Permission {
```
Patel-Raj11, line 41: I think we should introduce new enum for `GranularPermission` and add two static methods here that can build a `Permission` object using `TransactionType` and `GranularPermission` enum value.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * The maximum number of permissions that can be delegated in a single transaction.
+   */
+  int PERMISSIONS_MAX_LENGTH = 10;
+
+  /**
+   * Set of transaction types that cannot be delegated.
+   */
+  Set<String> NON_DELEGABLE_TRANSACTIONS = Sets.newHashSet(
```
Patel-Raj11: We should use `TransactionType` enum so that string value remains consistent.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * The transaction permissions (represented by transaction type strings) that the account has been granted.
+   *
+   * @return A {@link List} of {@link PermissionWrapper}s.
+   */
+  @JsonProperty("Permissions")
+  @JsonInclude(JsonInclude.Include.ALWAYS)
+  List<PermissionWrapper> permissions();
+
```
Patel-Raj11, line 113: We could add another check to throw an error if `Permission` object has a string value that does not belong to a `TransactionType` enum or GranularPermission. Like completely random Permission (e.g. "abc") should fail here.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
```diff
@@ -433,6 +434,10 @@ public <T extends Transaction> SingleSignedTransaction<T> addSignatureToTransact
       transactionWithSignature = CredentialDelete.builder().from((CredentialDelete) transaction)
         .transactionSignature(signature)
         .build();
+    } else if (DelegateSet.class.isAssignableFrom(transaction.getClass())) {
```
Patel-Raj11: nit: Since we are adding tests in `SignatureUtilsTest` for all transaction types, we should add it for `DelegateSet`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/DelegateObject.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+
+/**
+ * This object represents a set of permissions that an account has delegated to another account.
+ * {@link DelegateSet} transactions create these objects.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableDelegateObject.class)
+@JsonDeserialize(as = ImmutableDelegateObject.class)
+public interface DelegateObject extends LedgerObject {
```
Patel-Raj11, line 48: Unrelated change but commenting it out here:

1. We need to add Delegate to LedgerEntryRequestParams as well and create MetaDelegateObject as well.

2. Add tests in `LedgerEntryResultTest` and for point 1 as well.

Credential's PR can be referred to see what is missing for newly added ledger object and ledger entry as that PR introduced `Credential` object and this one introduces `Delegate` object.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateSetIT.java
```diff
@@ -0,0 +1,349 @@
+package org.xrpl.xrpl4j.tests;
```
Patel-Raj11, line 1: I think we need to add `Delegate` to the base transaction.

Also, if that field is present, it will sing the transaction and pay the fees - https://xls.xrpl.org/xls/XLS-0075-permission-delegation.html#4-transactions-common-fields

After implementing the above, we can write an integration test that verifies if Delegate can actually submit a Payment transaction or not.

Patel-Raj11 (changes requested):

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
Patel-Raj11: This should be removed.

xrpl4j-core/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
Patel-Raj11: This should be removed.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateSetIT.java
Patel-Raj11: Good to have only one file for DelegateIT. This can me merged into the other one.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
```diff
@@ -43,14 +45,61 @@ public DefaultDefinitionsProvider(final ObjectMapper objectMapper) {
...
+   * - Transaction type permissions are transaction type code + 1
+   *
+   * @param definitions The base {@link Definitions} loaded from definitions.json
+   * @return A {@link Map} of permission names to their numeric values
+   */
+  private Map<String, Integer> generatePermissionValues(Definitions definitions) {
+    Map<String, Integer> permissionValues = new HashMap<>();
+
+    // Add granular permissions (starting at 65537)
+    permissionValues.put("TrustlineAuthorize", 65537);
```
Patel-Raj11: This list is out of sync from `GranularPermission` enum. Could we use the same Enum here?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
```diff
@@ -43,14 +45,61 @@ public DefaultDefinitionsProvider(final ObjectMapper objectMapper) {
...
+    permissionValues.put("MPTokenIssuanceAuthorize", 65544);
+    permissionValues.put("MPTokenIssuanceUnlock", 65545);
+    permissionValues.put("MPTokenIssuanceLock", 65546);
+    permissionValues.put("MPTokenIssuanceTransfer", 65547);
+    permissionValues.put("MPTokenIssuanceClawback", 65548);
+
+    // Add transaction type permissions (transaction type code + 1)
+    // Exclude pseudo-transactions and non-delegatable transactions
+    definitions.transactionTypes().forEach((txType, txCode) -> {
+      // Skip Invalid (-1) and pseudo-transactions (>= 100)
```
Patel-Raj11: I didn't find the mention of `>= 100` transaction types should be not allowed in the spec. If that's not the case, can we remove the upper bound to allow future transaction types >= 100 to be delegable?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/UInt32Type.java
```diff
@@ -48,8 +52,45 @@ public UInt32Type fromJson(JsonNode value) {
     return new UInt32Type(UnsignedLong.valueOf(value.asText()));
   }
 
+  @Override
+  public UInt32Type fromJson(JsonNode value, FieldInstance fieldInstance) throws JsonProcessingException {
```
Patel-Raj11: I am not sure if need changes in this file. Can you remove these and check if integration tests works?

Patel-Raj11 (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
cybele-ripple: Mention of >= 100 is because transactions of types 100, 101, and 102 pseudotransactions. Since there may be future transactions that are over 100, I can remove this upper bound.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
```diff
@@ -363,6 +368,16 @@ public enum TransactionType {
...
   AMM_CLAWBACK("AMMClawback"),
+
+  /**
+   * The {@link TransactionType} for the Batch transaction.
+   *
+   * <p>This constant will be marked {@link Beta} until the Batch amendment is enabled on mainnet. Its API is subject to
+   * change.</p>
+   */
+  @Beta
+  BATCH("Batch"),
```
sappenin: Note: This branch was updated via GH to pull-in changes from `main`. Because Batch is now in `main`, this enum is a duplicate, and breaks the build. We'll want to remove this one to get the build working.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/UInt32Type.java
cybele-ripple: Removed changes to this file and the integration tests failed. Will keep for now

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
Patel-Raj11: @cybele-ripple Could we reuse the Enum here?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
```diff
@@ -26,6 +26,8 @@
 import com.google.common.io.Resources;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
 public class DefaultDefinitionsProvider implements DefinitionsProvider {
```
sappenin, line 45: Let's add a unit test for this called `DefaultDefinitionsProviderTest.java`. It should in particular test the result of `generatePermissionValues()` and also should cover the cases in line 93.

sappenin, line 45: Please add Javadoc to the class level. Consider something like:

```suggestion
/**
 * Default implementation of the {@link DefinitionsProvider} interface. This class is responsible for loading and
 * providing {@link Definitions} data from a JSON file, while also programmatically generating additional permission
 * values based on the loaded data.
 *
 * The definitions file is expected to be located at '/definitions.json' as a resource, and is parsed using the
 * specified {@link ObjectMapper}. Lazily initializes and memoizes the loaded {@link Definitions} data for improved
 * performance.
 */
public class DefaultDefinitionsProvider implements DefinitionsProvider {
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
```diff
@@ -0,0 +1,188 @@
...
+   * Construct a builder for this class.
+   *
+   * @return An {@link ImmutableDelegateSet.Builder}.
+   */
+  static ImmutableDelegateSet.Builder builder() {
+    return ImmutableDelegateSet.builder();
+  }
+
+  /**
+   * Set of {@link TransactionFlags}s for this {@link DelegateSet}, which only allows the {@code tfFullyCanonicalSig}
```
sappenin: This comment is a bit outdated due to Batch. Let's just adjust it to be more generic so we don't have to update these in every transaction in the future if another new Transaction flag is added.
```suggestion
   * Set of {@link TransactionFlags}s for this {@link DelegateSet}.
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
Patel-Raj11: Something like this:

`for (GranularPermission permission : GranularPermission.values()) {
      permissionValues.put(permission.value(), permission.numericValue());
    }`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/UInt32Type.java
Patel-Raj11: These changes have been removed and IT passes.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.xrpl.xrpl4j.model.transactions.Address;
+import org.xrpl.xrpl4j.model.transactions.DelegateSet;
+import org.xrpl.xrpl4j.model.transactions.Permission;
+import org.xrpl.xrpl4j.model.transactions.PermissionWrapper;
+import org.xrpl.xrpl4j.model.transactions.XrpCurrencyAmount;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+public class DelegateSetJsonTest extends AbstractJsonTest {
```
sappenin, line 43: Minor nit, but we try to add Javadoc to test classes in the following pattern (mostly to allow easy navigation in IDEs, but also just for consistency):

```suggestion
/**
 * Unit tests for {@link DelegateSet} JSON serialization and deserialization.
 */
public class DelegateSetJsonTest extends AbstractJsonTest {
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LedgerObject.java
```diff
@@ -164,6 +165,15 @@ enum LedgerEntryType {
      */
     NFTOKEN_PAGE("NFTokenPage"),
 
+    /**
+     * The {@link LedgerEntryType} for {@code Delegate} ledger objects.
+     *
+     * <p>This constant will be marked {@link Beta} until the featurePermissionDelegation amendment is enabled on mainnet.
```
Patel-Raj11: This line is going beyond 120 chars so build is failing. Adding line break will fix the issue.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
```diff
@@ -0,0 +1,188 @@
...
+  @JsonProperty("Authorize")
+  Address authorize();
+
+  /**
+   * The transaction permissions (represented by transaction type strings) that the account has been granted.
+   *
+   * @return A {@link List} of {@link PermissionWrapper}s.
+   */
+  @JsonProperty("Permissions")
+  @JsonInclude(JsonInclude.Include.ALWAYS)
```
sappenin, line 111: At first I thought this was incorrect, but I think it actually _is_ correct. I added a comment to an existing enhancement ticket (see [here](https://github.com/XRPLF/xrpl4j/issues/671#issuecomment-3937075415
)) to capture this.

At the same time, we need an extra unit test in `DelegateSetJsonTest.java` to cover the case where the permissions are empty. I'll post that in a different comment in that file.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
```diff
@@ -0,0 +1,111 @@
...
+
```
sappenin: Per [my comment](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-05713a5290db465ee8a57b96b9af5aff76fb39caac1b520d12e843b59e319919R109
) in `DelegateSet.java`, let's add one more negative test to increase coverage here:

```java
@Test
  public void testDelegateSetJsonWithNoPermissions() throws JsonProcessingException, JSONException {
    DelegateSet delegateSet = DelegateSet.builder()
      .account(Address.of("rsUiUMpnrgxQp24dJYZDhmV4bE3aBtQyt8"))
      .authorize(Address.of("rEhxGqkqPPSxQ3P25J66ft5TwpzV14k2de"))
      .sequence(UnsignedInteger.valueOf(2))
      .fee(XrpCurrencyAmount.ofDrops(10))
      .signingPublicKey(
        PublicKey.fromBase16EncodedPublicKey("ED87987410480E90474F7A02E0DA0CE4E6ABC8A1377864026A1FEE2718688B0B84")
      )
      .build();

    String json = "{" +
      "  \"Account\":\"rsUiUMpnrgxQp24dJYZDhmV4bE3aBtQyt8\"," +
      "  \"Authorize\":\"rEhxGqkqPPSxQ3P25J66ft5TwpzV14k2de\"," +
      "  \"Permissions\":[]," +
      "  \"Fee\":\"10\"," +
      "  \"Sequence\":2," +
      "  \"SigningPubKey\":\"ED87987410480E90474F7A02E0DA0CE4E6ABC8A1377864026A1FEE2718688B0B84\"," +
      "  \"TransactionType\":\"DelegateSet\"" +
      "}";

    assertCanSerializeAndDeserialize(delegateSet, json);
  }
```

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
```diff
@@ -0,0 +1,111 @@
...
+      "  \"Fee\":\"10\"," +
+      "  \"Sequence\":2," +
+      "  \"SigningPubKey\":\"ED87987410480E90474F7A02E0DA0CE4E6ABC8A1377864026A1FEE2718688B0B84\"," +
+      "  \"TransactionType\":\"DelegateSet\"" +
+      "}";
+
+    assertCanSerializeAndDeserialize(delegateSet, json);
+  }
+
+  @Test
```
sappenin, line 80: I think we need to cover a case or two that with JSON having granular permissions.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,748 @@
...
+  private NetworkId networkId;
+
+  @BeforeEach
+  public void setUp() throws JsonRpcClientErrorException {
+    // Query the network ID from the server
+    networkId = xrplClient.serverInformation().info().networkId()
+      .orElse(NetworkId.of(UnsignedInteger.ZERO));
+    logger.info("Using NetworkID: {}", networkId);
+  }
+
```
sappenin, line 89: I'm guessing these edge cases are probably not needed since they're testing things that should be covered by `xrpld` QE integration tests. However, if anything jumps out to you, feel free to add it. Here are some suggestions from AI (which are probably wrong for xrpl4j, but just in case):

1. ❌ **Negative test**: Delegate tries to execute transaction without permission
2. ❌ **Negative test**: Wrong delegate tries to use delegation
3. ❌ **Fee verification**: Confirm delegate pays fees (partially exists)
4. ❌ **Sequence verification**: Confirm delegating account's sequence increments

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
```diff
@@ -0,0 +1,89 @@
...
+ * {@link Permission} inner object with PermissionValue details.
+ *
+ * <p>This class will be marked {@link Beta} until the featurePermissionDelegation amendment is enabled on mainnet.
+ * Its API is subject to change.</p>
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutablePermission.class)
+@JsonDeserialize(as = ImmutablePermission.class)
+@Beta
+public interface Permission {
```
sappenin, line 41: Wondering about the naming here. Should we instead call this (per XLS-74) `AccountPermission`, for clarity?

Patel-Raj11 (commented):
Looks good for beta release.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Permission to unlock an MPToken issuance.
+   */
+  MPTOKEN_ISSUANCE_UNLOCK("MPTokenIssuanceUnlock", 65548);
+
+  private final String value;
+  private final int numericValue;
+
+  GranularPermission(String value, int numericValue) {
```
sappenin: Minor style nit, but prefer the following subtle changes (add Javadoc, make constructor params final, disallow null when possible):

```java
 /**
   * Constructs a {@code GranularPermission} instance with the specified value and numeric value.
   *
   * @param value        A string representing the name or identifier of the permission. This parameter cannot be null.
   * @param numericValue An integer representing the numeric value associated with the permission.
   */
  GranularPermission(final String value, final int numericValue) {
    this.value = Objects.requireNonNull(value);
    this.numericValue = numericValue;
  }
```

This is not so much of an issue for an enum constructor because of the way enums are created, but sharing it just for style consistency (here and in other places).

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @return The {@link GranularPermission} with the corresponding value, or null if not found.
+   */
+  public static GranularPermission forValue(String value) {
+    for (GranularPermission permission : GranularPermission.values()) {
+      if (permission.value.equals(value)) {
+        return permission;
+      }
+    }
+    return null;
```
sappenin: As much as possible, we disallow `null` in this project. Sometimes it's unavoidable, but we use `Optional` instead and aggressively fail on `null` anywhere in the library (this helps catch bugs in unit and integration testing, as opposed to production code). In fact, if we throw a NPE in production, we consider that to be a quality error on our part.

To that end, let's refactor to return `Optional` like this:

```java
public static Optional<GranularPermission> forValue(final String value) {
    Objects.requireNonNull(value);
    for (GranularPermission permission : GranularPermission.values()) {
      if (permission.value.equals(value)) {
        return Optional.of(permission);
      }
    }
    return Optional.empty();
  }
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Gets an instance of {@link GranularPermission} for the given string value.
+   *
+   * @param value The {@link String} value corresponding to a {@link GranularPermission}.
+   *
+   * @return The {@link GranularPermission} with the corresponding value, or null if not found.
+   */
+  public static GranularPermission forValue(String value) {
+    for (GranularPermission permission : GranularPermission.values()) {
+      if (permission.value.equals(value)) {
```
sappenin: WDYT: Should we make this comparison case-insensive, just in case?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/Definitions.java
```diff
@@ -77,4 +79,16 @@ public interface Definitions {
...
+  /**
+   * Permission values mappings (permission value to ordinal value).
+   * This field is not present in the generated definitions.json file and is populated
+   * programmatically by {@link DefaultDefinitionsProvider}.
+   *
+   * @return {@link Map} keyed by {@link String} with {@link Integer} values for all permission values.
+   */
+  @JsonProperty("PERMISSION_VALUES")
+  @Default
+  default Map<String, Integer> permissionValues() {
```
sappenin: Immutables should fill-in default collections for you automagically, so I _think_ this can be updated to this instead (unless you have some reason to keep this as-is):

```java
 @JsonProperty("PERMISSION_VALUES")
 Map<String, Integer> permissionValues();
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
sappenin: I see this comment as resolved, but when I run `SignatureUtilsTest.java` with coverage, it shows this as still uncovered.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
sappenin: I'm not seeing this fix implemented. Can we keep comments unresolved until they're pushed (to aid in code review)?

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/DelegateObjectJsonTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.json.JSONException;
+import org.junit.jupiter.api.Test;
+import org.xrpl.xrpl4j.model.AbstractJsonTest;
+import org.xrpl.xrpl4j.model.transactions.Address;
+import org.xrpl.xrpl4j.model.transactions.Hash256;
+import org.xrpl.xrpl4j.model.transactions.Permission;
+import org.xrpl.xrpl4j.model.transactions.PermissionWrapper;
+
+import java.util.Arrays;
+import java.util.Collections;
```
sappenin, line 35: This import appears to be unused.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/DelegateSetTest.java
```diff
@@ -0,0 +1,505 @@
...
+    assertThat(delegateSet.transactionType()).isEqualTo(TransactionType.DELEGATE_SET);
+    assertThat(delegateSet.account()).isEqualTo(Address.of("rsUiUMpnrgxQp24dJYZDhmV4bE3aBtQyt8"));
+    assertThat(delegateSet.authorize()).isEqualTo(authorize);
+    assertThat(delegateSet.permissions()).hasSize(2);
+    assertThat(delegateSet.permissions().get(0).permission().permissionValue()).isEqualTo("Payment");
+    assertThat(delegateSet.permissions().get(1).permission().permissionValue()).isEqualTo("TrustSet");
+  }
+
+  @Test
+  public void testDelegateSetWithoutPermissions() {
```
sappenin: Let's add one more variant that simply omits the `.permissions` call to make sure the Immutable setup is correct:

```java
 @Test
  public void testDelegateSetWithoutPermissions() {
    // DelegateSet without permissions should work (removes delegation)
    Address authorize = Address.of("rEhxGqkqPPSxQ3P25J66ft5TwpzV14k2de");

    DelegateSet delegateSet = DelegateSet.builder()
      .account(Address.of("rsUiUMpnrgxQp24dJYZDhmV4bE3aBtQyt8"))
      .fee(XrpCurrencyAmount.ofDrops(10))
      .sequence(UnsignedInteger.valueOf(2))
      .authorize(authorize)
      // .permissions(Collections.emptyList()) <-- Validate 
      .build();

    assertThat(delegateSet.authorize()).isEqualTo(authorize);
    assertThat(delegateSet.permissions()).isEmpty();
  }
```

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/DelegateSetTest.java
```diff
@@ -0,0 +1,505 @@
...
+
+      assertThat(delegateSet.permissions()).hasSize(1);
+      assertThat(delegateSet.permissions().get(0).permission().permissionValue()).isEqualTo(txType.value());
+    }
+  }
+
+  @Test
+  public void testAllNonDelegatableTransactionTypesWithEnum() {
+    // Test all non-delegatable transaction types using TransactionType enum
+    TransactionType[] nonDelegatableTypes = {
```
sappenin: Should we instead reuse `DelegateSet#NON_DELEGABLE_TRANSACTIONS` here?

Overall, I worry too about future transaction added to the library that eventually become untested, either here or in `testAllDelegatableTransactionTypes`. Maybe we should first take all transactions in the enum, and remove the undelegateble ones? That way, if anyone adds a new TransactionType in the future, it will get included in here automatically.

WDYT?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,748 @@
...
+    AccountInfoResult updatedDelegatingAccountInfo = this.scanForResult(
+      () -> this.getValidatedAccountInfo(delegatingAccountKeyPair.publicKey().deriveAddress())
+    );
+
+    // Get delegate account info to check balance before payment
+    final AccountInfoResult delegateAccountInfoBefore = this.scanForResult(
+      () -> this.getValidatedAccountInfo(delegateKeyPair.publicKey().deriveAddress())
+    );
+
+
```
sappenin: Minor nit: extra newlines.

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
cybele-ripple: Changes made here: https://github.com/XRPLF/xrpl4j/pull/689/changes/a270901977a8fe39fe3654eaab44501e54a6c0c0#diff-05713a5290db465ee8a57b96b9af5aff76fb39caac1b520d12e843b59e319919L53

Patel-Raj11: I resolved this because the values were at least using `TransactionType` enum to get string value. But, yes we should do this - `Set<TransactionType> NON_DELEGABLE_TRANSACTIONS`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
Patel-Raj11: I think if we allow here, it will end up failing in binary codec (due to the way it's implemented). If we allow binary codec to do case-insensitive mapping as well it should be good. But when we deserialize from JSON, it will always be in CamelCase (for both TransactionTypes and GranularPermissions) since we won't know what was the original input.

I am good with case-insensitive comparison at both places.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
cybele-ripple: Changed based on your suggestion

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
cybele-ripple: Changed

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
cybele-ripple: Added extra unit test

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
cybele-ripple: Added test

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
cybele-ripple: Improved the following tests in DelegateIT:
`testPaymentWithDelegate()` - added to test to ensure that the delegate pays the transaction fees, test delegate's sequence increment

Added the following tests to DelegateIT: 
`testDelegateWithoutPermission()` - tests using transaction without proper permissions
`testWrongDelegateCannotUsePermission()` - tests that specific delegate with proper permissions can use permission
`testCannotDelegateToSelf()` - tests that a delegate cannot give permissions to self

xrpl4j-bom/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>xrpl4j-parent</artifactId>
     <groupId>org.xrpl</groupId>
-    <version>HEAD-SNAPSHOT</version>
+    <version>HEAD-SNAPSOT</version>
```
sappenin: Looks like a typo.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
cybele-ripple: Changes Permission to AccountPermission throughout

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
cybele-ripple: Refactored to return optional

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
cybele-ripple: Changed to case insensitive

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
cybele-ripple: Added javadoc

cybele-ripple: Added DefaultDefinitionsProviderTest.java

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/DelegateSetJsonTest.java
cybele-ripple: Added `testDelegateSetJsonWithGranularPermissions()` which tests serialization/deserialization, verifies correct JSON format with different granular permissions

xrpl4j-core/package-lock.json
cybele-ripple: removed

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
cybele-ripple: Updated to use Set<TransactionType> NON_DELEGABLE_TRANSACTIONS

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
cybele-ripple: Added coverage for this line and reran tests with coverage to confirm

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
cybele-ripple: Removed enum

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
cybele-ripple: Added this change

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/Definitions.java
cybele-ripple: Added

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/DelegateSetTest.java
cybele-ripple: Test now takes all TransactionType vals
As of now, new delegable transaction types just need to be added to the TransactionType enum. Adding non delegable transactions just involves an additional change to a single file so I think the current implementation is fine.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,939 @@
...
+ * is not yet implemented in any released version of rippled. The amendment was disabled in rippled 2.6.1
+ * due to a bug and has not been re-enabled. These tests will fail with "temDISABLED" until XLS-75 is
+ * implemented and enabled on a rippled node.</p>
+ *
+ * <p>To run these tests (they will fail): mvn test -Dtest=DelegateIT -DenableXLS75Tests=true</p>
+ */
+@DisabledIf(value = "shouldNotRun", disabledReason = "XLS-75 PermissionDelegation is not yet implemented in rippled")
+public class DelegateIT extends AbstractIT {
+
+  static boolean shouldNotRun() {
```
Patel-Raj11, line 74: I think we should remove this now and add amendment name to `xrpld.cfg`.

cybele-ripple: Fixed [here](https://github.com/XRPLF/xrpl4j/pull/689/commits/ddd3a395bc2cc8860296297e5cb013305e2affb0)

xrpl4j-bom/pom.xml
cybele-ripple: Fixed

xrpl4j-integration-tests/src/test/resources/xrpld/xrpld.cfg
```diff
@@ -230,6 +230,7 @@ fixFrozenLPTokenTransfer
 # 2.5.0 Amendments
 Batch
 PermissionedDEX
+PermissionDelegation
```
Patel-Raj11: @cybele-ripple The amendment name should be `PermissionDelegationV1_1`. Due to that ITs are failing.

Patel-Raj11 (changes requested):

xrpl4j-integration-tests/src/test/resources/xrpld/xrpld.cfg
Patel-Raj11: @cybele-ripple I think tests are still failing after this commit - [db3cf6b](https://github.com/XRPLF/xrpl4j/pull/689/commits/db3cf6b33836e9f860057178230a71f5d6f38fc6). Could you please have a look?

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
```diff
@@ -43,14 +55,55 @@ public DefaultDefinitionsProvider(final ObjectMapper objectMapper) {
 
     this.supplier = Suppliers.memoize(() -> {
       try {
-        return objectMapper.readerFor(Definitions.class)
+        Definitions baseDefinitions = objectMapper.readerFor(Definitions.class)
```
sappenin: It looks like `readValue` is deprecated, and the try/catch block is too broad. I suggest we refactor this section to look like this:

```java
this.supplier = Suppliers.memoize(() -> {
      final URL resource = Resources.getResource(DefaultDefinitionsProvider.class, "/definitions.json");
      final Definitions baseDefinitions;
      try (InputStream inputStream = resource.openStream()) {
        baseDefinitions = objectMapper.readerFor(Definitions.class).readValue(inputStream);
      } catch (IOException e) {
        throw new IllegalStateException("Cannot read definition.json file", e);
      }

      // Generate PERMISSION_VALUES dynamically from TRANSACTION_TYPES
      final Map<String, Integer> permissionValues = generatePermissionValues(baseDefinitions);

      // Return a new Definitions object with the generated PERMISSION_VALUES
      return ImmutableDefinitions.builder()
        .from(baseDefinitions)
        .permissionValues(permissionValues)
        .build();
    });
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountPermission.java
```diff
@@ -0,0 +1,90 @@
...
+      .build();
+  }
+
+  /**
+   * The transaction type that is being delegated, represented as a string.
+   *
+   * @return A {@link String} representing the transaction type.
+   */
+  @JsonProperty("PermissionValue")
+  String permissionValue();
```
sappenin: Can we make this type a `TransactionType`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/DelegateObject.java
```diff
@@ -0,0 +1,140 @@
...
+ * {@link DelegateSet} transactions create these objects.
+ *
+ * <p>This class will be marked {@link Beta} until the featurePermissionDelegation amendment is enabled on mainnet.
+ * Its API is subject to change.</p>
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableDelegateObject.class)
+@JsonDeserialize(as = ImmutableDelegateObject.class)
+@Beta
+public interface DelegateObject extends LedgerObject {
```
sappenin, line 48: This class needs a unit test (mostly for JSON validation)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
```diff
@@ -0,0 +1,156 @@
...
+ * rather than entire transaction types.
+ *
+ * <p>Granular permission values are always greater than 65536 (UINT16_MAX)
+ * to avoid conflicts with transaction type permission values.</p>
+ *
+ * <p>This enum will be marked {@link Beta} until the featurePermissionDelegation amendment is enabled on mainnet.
+ * Its API is subject to change.</p>
+ */
+@Beta
+public enum GranularPermission {
```
sappenin, line 41: This needs a unit test.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
```diff
@@ -395,13 +405,15 @@ public enum TransactionType {
...
+   * <p>The comparison is case-insensitive to handle variations in transaction type formatting.</p>
+   *
    * @param value The {@link String} value corresponding to a {@link TransactionType}.
    *
    * @return The {@link TransactionType} with the corresponding value.
    */
   public static TransactionType forValue(String value) {
     for (TransactionType transactionType : TransactionType.values()) {
-      if (transactionType.value.equals(value)) {
+      if (transactionType.value.equalsIgnoreCase(value)) {
```
sappenin, line 551: Does Delegate introduce some occurrences of transaction types having incorrect capitalization?

(this change seems like a fine improvement, I ask more out of curiosity).

We also probably want to update `TransactionTypeTest` to include some examples of this (e.g., `ACCOUNT_SET` or whatever).

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProviderTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    // Find Invalid transaction type (should have code -1)
+    transactionTypes.forEach((txType, txCode) -> {
+      if (txCode == -1) {
+        // Invalid should not be in permission values
+        assertThat(permissionValues).doesNotContainKey(txType);
+      }
+    });
+  }
+
```
sappenin, line 113: Nit: empty lines

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/DelegateObjectJsonTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Unit tests for {@link DelegateObject}.
+ */
+public class DelegateObjectJsonTest extends AbstractJsonTest {
+
+  @Test
+  public void testDelegateObjectJsonWithPermissions() throws JsonProcessingException, JSONException {
```
sappenin: Both of the tests in this class contain "multiple permissions", despite the second test's name.

Could we adjust this slightly to test other boundary conditions, like:

1. A single permission
2. Empty list of permissions.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/DelegateSetTest.java
sappenin: Re-opening this one. The test currently called `testDelegateSetWithoutPermissions` should be renamed to `testDelegateSetWithEmptyPermissions`. Then we need a new test called `testDelegateSetWithoutPermissions` that actually omits the `.permission` call in the builder.

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefinitionsService.java
```diff
@@ -184,6 +187,8 @@ public Optional<Integer> mapFieldSpecialization(String fieldName, String value)
         return Optional.ofNullable(definitions.transactionResults().get(value));
       case "TransactionType":
         return Optional.ofNullable(definitions.transactionTypes().get(value));
+      case "PermissionValue":
```
sappenin, line 190: Per the [coverage report](https://app.codecov.io/gh/XRPLF/xrpl4j/pull/689?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=XRPLF), these lines are not covered by unit tests. In fact, this entire class is missing a unit test, which is technical debt (we usually add unit test coverage for this kind of thing if we make changes in a class).

If you want to just cover this function, I would be fine with that. But if you want to add coverage for this whole class, that would be awesome too.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountPermission.java
cybele-ripple: The current design using a string allows for both TransactionType.PAYMENT and GranularPermission.TRUSTLINE_AUTHORIZE

sappenin: Oh, good point. Still, it seems like we should design a type for this kind of thing that can accept either thing? (sort of like how `CurrencyAmount` works to allow for a polymorphic sub-type)?

sappenin: Actually, it seems like `AccountPermission` is maybe that parent-type I'm looking for. So, it seems like `GranularAccountPermission` and `TransactionTypeAccountPermission` might be the interfaces I'm imagining. Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/TransactionType.java
cybele-ripple: Delegate does not introduce any new occurrences of incorrect capitalization. In the newest commit there are tests that cover case-insensitive matching in TransactionTypeTests.java to be safe

sappenin: Excellent, thanks for clarifying. Resolving.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountPermission.java
cybele-ripple: That makes sense. Will do

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/PermissionDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+    String value = jsonParser.getValueAsString();
+
+    // Try to parse as TransactionType first
+    TransactionType transactionType = TransactionType.forValue(value);
+    if (transactionType != TransactionType.UNKNOWN) {
+      return TransactionTypePermission.of(transactionType);
+    }
+
+    // Try to parse as GranularPermission
+    Optional<GranularPermission> granularPermission = GranularPermission.forValue(value);
```
sappenin: Consider (at your preference) using Stream here, which is a clearer than dereferencing to an `Optional` and then having to call `.get()` on the Optional (a good rule of thumb is if you ever need to call `.get()` on an optional, then Streams are you friend).

```java
// Try to parse as GranularPermission. If not present, throw an exception because one of 
// either transactionType or granularPermission must be present.
return GranularPermission.forValue(value)
  .map(GranularPermissionValue::of)
  .orElseThrow(() -> new IllegalArgumentException(
    String.format("Unknown permission value: %s", value)
  ));
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
```diff
@@ -0,0 +1,101 @@
...
+      throw new IllegalStateException(String.format("Unsupported Permission Type: %s", this.getClass()));
+    }
+  }
+
+  /**
+   * Get the string value of this permission for JSON serialization.
+   *
+   * @return A {@link String} representing the permission value.
+   */
+  String value();
```
sappenin: We should remove this function from the `Permission` interface. This will force callers to use the handle/map methods to get a concrete instance. Jackson should still work given line 40 above.

This will have the nice side-effect of preventing someone from trying to instantiate an instance of `Permission`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/PermissionSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+    super(Permission.class, false);
+  }
+
+  @Override
+  public void serialize(
+    Permission permission,
+    JsonGenerator gen,
+    SerializerProvider provider
+  ) throws IOException {
+    gen.writeString(permission.value());
```
sappenin: Once `.value` is removed, we can use the mapper to write out each value

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
```diff
@@ -42,13 +56,55 @@ public DefaultDefinitionsProvider(final ObjectMapper objectMapper) {
...
       } catch (IOException e) {
         throw new IllegalStateException("Cannot read definition.json file", e);
       }
+
+      // Generate PERMISSION_VALUES dynamically from TRANSACTION_TYPES
+      final Map<String, Integer> permissionValues = generatePermissionValues(baseDefinitions);
+
+      // Return a new Definitions object with the generated PERMISSION_VALUES
+      return ImmutableDefinitions.builder()
+              .from(baseDefinitions)
```
sappenin: Minor nit: formatting looks off.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/PermissionDeserializer.java
cybele-ripple: Fixed in [this commit](https://github.com/XRPLF/xrpl4j/commit/73246d3c90a4f671a63706fd8b4be5168e877dec)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
cybele-ripple: Fixed in [this commit](https://github.com/XRPLF/xrpl4j/commit/73246d3c90a4f671a63706fd8b4be5168e877dec)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/PermissionSerializer.java
cybele-ripple: Fixed in [this commit](https://github.com/XRPLF/xrpl4j/commit/73246d3c90a4f671a63706fd8b4be5168e877dec)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
cybele-ripple: Fixed in [this commit](https://github.com/XRPLF/xrpl4j/commit/73246d3c90a4f671a63706fd8b4be5168e877dec)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,908 @@
...
+    assertThat(entryByIndex.node()).isEqualTo(delegateEntry.node());
+
+    /////////////////////////
+    // Query by index without type parameter
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<LedgerObject> entryByIndexUnTyped = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.index(
+        expectedDelegateObject.index(),
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: Resolved in commit ebab59a

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,908 @@
...
+
+    /////////////////////////
+    // Also query by index to verify both methods return the same object
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<DelegateObject> entryByIndex = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.index(
+        expectedDelegateObject.index(),
+        DelegateObject.class,
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: Resolved in commit ebab59a

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,908 @@
...
+    // Query the Delegate object via ledger_entry using DelegateLedgerEntryParams
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<DelegateObject> delegateEntry = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.delegate(
+        org.xrpl.xrpl4j.model.client.ledger.DelegateLedgerEntryParams.builder()
+          .account(delegatorKeyPair.publicKey().deriveAddress())
+          .authorize(delegateKeyPair.publicKey().deriveAddress())
+          .build(),
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: Resolved in commit ebab59a

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,908 @@
...
+    logger.info(
+      "DelegateSet removal transaction successful. Hash: {}",
+      removeResult.transactionResult().hash()
+    );
+
+    /////////////////////////
+    // Wait for the removal transaction to be validated
+    TransactionResult<DelegateSet> validatedRemoveTransaction = this.scanForResult(
+      () -> this.getValidatedTransaction(removeResult.transactionResult().hash(), DelegateSet.class)
+    );
```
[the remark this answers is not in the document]
cybele-ripple: Resolved in commit ebab59a

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,908 @@
...
+
+    SubmitResult<DelegateSet> createResult = xrplClient.submit(signedCreateDelegateSet);
+    assertThat(createResult.engineResult()).isEqualTo("tesSUCCESS");
+
+    /////////////////////////
+    // Wait for the create transaction to be validated
+    TransactionResult<DelegateSet> validatedCreateTransaction = this.scanForResult(
+      () -> this.getValidatedTransaction(createResult.transactionResult().hash(), DelegateSet.class)
+    );
```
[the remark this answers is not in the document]
cybele-ripple: Resolved in commit ebab59a

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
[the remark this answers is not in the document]
Patel-Raj11: @cybele-ripple Unrelated to this comment but there are some merge conflicts. I think we should rebase.

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefinitionsService.java
cybele-ripple: Added [here](xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/definitions/DefinitionsServiceTest.java)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/DelegateSetTest.java
cybele-ripple: DelegateSetTest.java now has both [testDelegateSetWithEmptyPermissions() ](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-6cb918c97944fc8fb494330cc484f48e211ba6768c0cbd31a861f84ec843c040R73-R87) and [testDelegateSetWithoutPermissions](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-6cb918c97944fc8fb494330cc484f48e211ba6768c0cbd31a861f84ec843c040R90-R104)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProvider.java
cybele-ripple: DefaultDefinitionsProdver.java now implements this [here](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-5c53e8e5fc784ef7a5681f599658f7445ed2c42ca04f10104f8feaec1c696877R58-R76)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/AccountPermission.java
cybele-ripple: Implemented [here](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-9b3de3f4f19e7a10d0f18bebe3a4932d4bd0ab7193c8481e584c6464bb94fa37R70-R96)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/DelegateObject.java
cybele-ripple: Added [xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/DelegateObjectTest.java](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-d1f7709be564e22a25023af4c6d9cf7454aedf422190de40d95a432093b1296a)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/definitions/DefaultDefinitionsProviderTest.java
cybele-ripple: Resolved [here](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-d0dd0a41113d300e804478fc379dc26e7b52a5ad64fc1cbdb0c9ddd01fc76aadR113)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/GranularPermission.java
cybele-ripple: Added [xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/GranularPermissionTest.java](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-cab9cd8c046212ff39564f32742fa898f790dca52f95c030d083ee9ff1456781)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/DelegateObjectJsonTest.java
cybele-ripple: 1. [Added single permission](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-c75cb87b4ff10e9d87c18b20d7f1627680dfe3c630474f13d8c9eb8f3b55e463R69-R99)
2. [Empty list of permissions](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-c75cb87b4ff10e9d87c18b20d7f1627680dfe3c630474f13d8c9eb8f3b55e463R43-R66)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,916 @@
...
+    // Query the Delegate object via ledger_entry using DelegateLedgerEntryParams
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<DelegateObject> delegateEntry = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.delegate(
+        org.xrpl.xrpl4j.model.client.ledger.DelegateLedgerEntryParams.builder()
+          .account(delegatorKeyPair.publicKey().deriveAddress())
+          .authorize(delegateKeyPair.publicKey().deriveAddress())
+          .build(),
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: delegateEntry is used on lines [707-710](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR707-R710), and [722](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR722)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,916 @@
...
+
+    /////////////////////////
+    // Also query by index to verify both methods return the same object
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<DelegateObject> entryByIndex = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.index(
+        expectedDelegateObject.index(),
+        DelegateObject.class,
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: entryByIndex is used on lines [722](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR722) and [733](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR733)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
```diff
@@ -0,0 +1,916 @@
...
+    assertThat(entryByIndex.node()).isEqualTo(delegateEntry.node());
+
+    /////////////////////////
+    // Query by index without type parameter
+    org.xrpl.xrpl4j.model.client.ledger.LedgerEntryResult<LedgerObject> entryByIndexUnTyped = xrplClient.ledgerEntry(
+      org.xrpl.xrpl4j.model.client.ledger.LedgerEntryRequestParams.index(
+        expectedDelegateObject.index(),
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
cybele-ripple: entryByIndexUnTyped is used on [line 733](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR733)

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Permission.java
sappenin: The overall naming here still feels slightly "off" to me. From what I can tell, this is what's there currently:

```
`AccountPermission` (Holds a Permission)

`Permission` (parent interface)
`GranularPermissionValue` extends Permission (holds a `GranularPermission`)
`TransactionTypePermission` (extends Permission, holds a `TransactionType`)

`GranularPermission` (Enum with actual values)
`TransactionType` (enum with actual values)
```

I think we want to aim towards more congruity. Because we can't (and likely don't want to) rename `TransactionType`, we can start from there and do the following:

1.  Rename `GranularPermission` to `GranularPermissionType`
2. Rename `GranularPermissionValue` to `GranularPermissionTypePermission`

This will yield the following naming conventions (which also happen to conform to typical Java naming idioms):

```
`AccountPermission` (Holds a Permission)

`Permission`
`GranularPermissionTypePermission` extends `Permission` (holds a `GranularPermissionType`)
`TransactionTypePermission` (extends `Permission`, holds a `TransactionType`)

`GranularPermissionType` (Enum with actual enum values)
`TransactionType` (enum with actual enum values)
```

Thoughts?

sappenin (changes requested):

sappenin (dismissed):

Patel-Raj11 (approved):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/DelegateSet.java
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * The maximum number of permissions that can be delegated in a single transaction.
+   */
+  int PERMISSIONS_MAX_LENGTH = 10;
+
+  /**
+   * Set of transaction types that cannot be delegated.
+   */
+  Set<TransactionType> NON_DELEGABLE_TRANSACTIONS = Sets.newHashSet(
```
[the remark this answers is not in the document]
cybele-ripple: Added this in the most recent commit

sappenin (approved):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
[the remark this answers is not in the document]
cybele-ripple: `LedgerEntryResult delegateEntry` is used in [lines 707-710](https://github.com/XRPLF/xrpl4j/pull/689/changes/BASE..a937da5ea75c271ed7045d77eed3759361c19e1f#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR707-R710)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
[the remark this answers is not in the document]
cybele-ripple: `entryByIndex` is used in [lines 722](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR722) and [733](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR733)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/DelegateIT.java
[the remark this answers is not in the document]
cybele-ripple: `entryByIndexUnTyped` is used in [line 733](https://github.com/XRPLF/xrpl4j/pull/689/changes#diff-8fc4b7d7f7b4f334455912d90b165899fa2bcd72ce97988d978c7fac6a591c8aR733)

PR #705, opened 2026-02-24
Document manual release process

Patel-Raj11:
This PR documents manual release process for `xrpl4j` artifacts. These steps would be used as a guide for release automation and serve as future reference, in the event we need to manually release `xrpl4j` package.


RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+This document describes the manual release process for xrpl4j.
+
+## Stable Releases
+
+1. Checkout main and pull in latest changes from remote
+   ```  
+   git checkout main
+   git pull
+   ```
+2. Create a new release branch from main (e.g. `releases/v6.0` for major version release, `releases/v6.1` for minor version release)
```
Patel-Raj11: @sappenin 
I kept this step because it's the convention that is currently being followed in `xrpl4j` release process. Since we are tagging the commits, we can skip creating the release branches and rely on git tags to know what should be the next version to release. WDYT?

If we decide to move away from release branches, then we won't be able to update the patch version in advance (as we do in step 9 currently). Would that cause any issue? Because currently in `xrpl.js` and `xrpl-py` we always update the version before doing any release.

Also, in `xrpl.js` and `xrpl-py` we push that latest released version to `main` branch, could we follow the same convention in `xrpl4j`?

sappenin: Thinking about it a bit more, let's keep the release branch pattern + versioning strategy for now - I haven't decided how we should track patches if we only tag.

RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+   ```  
+   git checkout -b releases/v6.1  
+   ```  
+
+3. Set intended release version
+   ```  
+   mvn versions:set -DnewVersion=6.1.0
```
sappenin:
```suggestion
   mvn versions:set -DnewVersion=6.1.0-SNAPSHOT
```

sappenin: Only `SNAPSHOT` should be committed into the repo. The non-SNAPSHOT versions are only interstitially existing on whatever machine deploys the artifacts.

RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+   ``` 
+4. Commit and push changes in `pom.xml` files created by step #3.
+5. Verify Integration tests that executes on `Devnet` passes in GitHub actions. Integration tests that executes on 
+   `Testnet` might fail as the amendment would generally not be available on `Testnet` yet.
+6. Create a new tag for the release and push to remote
+   ```  
+   git tag v6.1.0
+   git push origin v6.1.0
+   ```
+7. Generate and deploy artifacts to [maven central repository](https://central.sonatype.com/) (This would prompt to sign the artifacts using private key)
```
sappenin: Here we would execute the version bump, to deploy:

```
mvn versions:set -DnewVersion=6.1.0
```

RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+   ```  
+   mvn clean deploy -DskipTests -P release
+   ```
+8. Approve the release on [maven central repository](https://central.sonatype.com/)
+9. Update the version to next patch version, commit and push to remote
+   ```  
+   mvn versions:set -DnewVersion=6.1.1
```
sappenin:
```suggestion
   mvn versions:set -DnewVersion=6.1.1-SNAPSHOT
```

RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+   git commit -m "update version to 6.1.1"
```
sappenin:
```suggestion
   git commit -m "update version to 6.1.1-SNAPSHOT"
```

RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+## Stable Releases
+
+1. Checkout main and pull in latest changes from remote
+   ```  
+   git checkout main
+   git pull
+   ```
+2. Create a new release branch from main (e.g. `releases/v6.0` for major version release, `releases/v6.1` for minor version release)
+   ```  
+   git checkout -b releases/v6.1
```
sappenin: Looking at this more, I worry people will think `v6.1` is normative. Maybe we should do something like `v[major].[minor]` and when we want to bump, say something like this:

Increment the version from `v[major].[minor].[patch]-SNAPSHOT` to `v[major].[minor].[patch+1]-SNAPSHOT`. For example, increment from `v6.0.0-SNAPSHOT` to `v6.0.1-SNAPHSOT`.

RELEASE.md
Patel-Raj11: Ok, have kept the release steps that we currently follow.

RELEASE.md
Patel-Raj11: Addressed in [8dac788](https://github.com/XRPLF/xrpl4j/pull/705/commits/8dac788ad63a6e2b92e430c946b5efb142847880)

RELEASE.md
Patel-Raj11: Addressed in [8dac788](https://github.com/XRPLF/xrpl4j/pull/705/commits/8dac788ad63a6e2b92e430c946b5efb142847880)

RELEASE.md
Patel-Raj11: Addressed in [8dac788](https://github.com/XRPLF/xrpl4j/pull/705/commits/8dac788ad63a6e2b92e430c946b5efb142847880)

RELEASE.md
Patel-Raj11: I have added the generic version numbers as well as concrete example in [8dac788](https://github.com/XRPLF/xrpl4j/pull/705/commits/8dac788ad63a6e2b92e430c946b5efb142847880).

Also, I have split the steps into `major/minor` release and `patch` release. Since we don't create a separate `releases/v[major].[minor].[patch]` branch for patch releases, currently it was confusing. Could you have another look in the patch releases section?

PR #704, opened 2026-02-23
XLS-82d MPT-DEX support

cybele-ripple:
Added support for [XLS-82d](https://github.com/XRPLF/XRPL-Standards/discussions/231)

PR #713, opened 2026-03-16
Add Single Asset Vault support - XLS-65

Patel-Raj11:
Single Asset Vault - [XLS-65](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065-single-asset-vault)
Single Asset Vault had multiple PRs, so rippled/xrpld release [3.1.0](https://github.com/XRPLF/rippled/tree/3.1.0) was considered as source of truth for Transactions, ledger objects, RPC and binary codec changes.

Transactions added/updated:
1. VaultCreate
2. VaultSet
3. VaultDelete
4. VaultDeposit
5. VaultWithdraw
6. VaultClawback
7. MpTokenIssuanceCreate
8. MpTokenIssuanceSet

Ledger objects added/updated:
1. VaultObject
2. MpTokenIssuanceObject

RPC added/updated:
1. vault_info

Binary codec added/updated:
1. IssueType
2. NumberType

MPT DEX [PR](https://github.com/XRPLF/xrpl4j/pull/704) and this PR both have IssueType binary codec change and Issue object refactoring change in common. Since, Single Asset Vault is up for voting, this PR would likely get merged first. So, I have kept these suggestions in mind when making changes:
1. https://github.com/XRPLF/xrpl4j/pull/704#discussion_r2921606453
2. https://github.com/XRPLF/xrpl4j/pull/704#discussion_r2921181552
3. https://github.com/XRPLF/xrpl4j/pull/704#discussion_r2921589771
4. https://github.com/XRPLF/xrpl4j/pull/704#discussion_r2921600560


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/IssueType.java
```diff
@@ -27,7 +37,30 @@ public IssueType fromJson(JsonNode node) throws JsonProcessingException {
 
     Issue issue = objectMapper.treeToValue(node, Issue.class);
 
-    UnsignedByteArray byteArray = new CurrencyType().fromJson(issue.currency()).value();
+    if (issue.mptIssuanceId().isPresent()) {
+      // MPTokenIssuanceID binary format: issuer (20 bytes) + ACCOUNT_ONE (20 bytes) + sequence (4 bytes) = 44 bytes
```
sappenin: Remind me why rippled adds an additional `ACCOUNT_ONE` into the binary encoding? I feel like there was a good reason, but I can't remember it.

Also, we should update the comment description to properly describe the two different formats. Currently, the only descriptor is `binary` vs `hex`, which is confusing (generally binary and hex should be the same, but in this case they're not, which means they're describing two different values)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+        exponent++;
+      }
+
+      // Round up if last removed digit was >= 5
+      if (lastDigit != null && lastDigit.compareTo(BigInteger.valueOf(5)) >= 0) {
+        absMantissa = absMantissa.add(BigInteger.ONE);
+        // After rounding, mantissa may exceed MAX_INT64 again
+        if (absMantissa.compareTo(MAX_INT64) > 0) {
+          if (exponent >= MAX_EXPONENT) {
+            throw new IllegalArgumentException("Exponent overflow: value too large to represent");
```
sappenin, line 127: This line seems uncovered by the unit test.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+          }
+          lastDigit = absMantissa.mod(BigInteger.TEN);
+          absMantissa = absMantissa.divide(BigInteger.TEN);
+          exponent++;
+          if (lastDigit.compareTo(BigInteger.valueOf(5)) >= 0) {
```
sappenin, line 132: This line seems uncovered by the unit test.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/Issue.java
```diff
@@ -26,26 +27,46 @@ static ImmutableIssue.Builder builder() {
...
   Optional<JsonNode> issuer();
 
   /**
-   * Validate that {@link #issuer()} is empty if {@link #currency()} is "XRP".
+   * The MPT issuance ID. Present for MPT assets.
+   *
+   * @return An optionally present {@link JsonNode}.
+   */
+  @JsonProperty("mpt_issuance_id")
+  Optional<JsonNode> mptIssuanceId();
```
sappenin, line 49: This is technically a breaking change (though I doubt any external software is using this object, but you never know).

Can we capture somewhere a list of all breaking changes in this PR? We'll want to use that to draft release notes. We should probably have (as part of this PR) something that outlines info that will eventually make its way into a migration guide, sort of like [this one](https://github.com/XRPLF/xrpl4j/blob/main/V6_MIGRATION.md).

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
```diff
@@ -489,6 +495,30 @@ public <T extends Transaction> SingleSignedTransaction<T> addSignatureToTransact
       transactionWithSignature = PermissionedDomainDelete.builder().from((PermissionedDomainDelete) transaction)
         .transactionSignature(signature)
         .build();
+    } else if (VaultCreate.class.isAssignableFrom(transaction.getClass())) {
```
sappenin: FYI - take a look at #712 and #715. That change will remove the need to update these types here.

I suggest we merge that before merging this PR, FWIW.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/vault/VaultInfoRequestParams.java
```diff
@@ -0,0 +1,122 @@
...
+      );
+    }
+
+    if (hasVaultId && hasOwnerAndSeq) {
+      throw new IllegalArgumentException(
+        "VaultInfoRequestParams cannot specify both vaultId and owner/seq"
+      );
+    }
+
+    if (owner().isPresent() != seq().isPresent()) {
```
sappenin, line 115: I'm not seeing coverage for this chunk of code.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/client/vault/VaultInfoResultTest.java
```diff
@@ -0,0 +1,152 @@
+package org.xrpl.xrpl4j.model.client.vault;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
```
sappenin: Some unused imports.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -841,6 +845,20 @@ public String toString() {
 
   }
 
+  @Value.Immutable
+  @Wrapped
+  @JsonSerialize(as = AssetAmount.class, using = AssetAmountSerializer.class)
+  @JsonDeserialize(as = AssetAmount.class, using = AssetAmountDeserializer.class)
+  @Beta
+  abstract static class _AssetAmount extends Wrapper<String> implements Serializable {
```
sappenin: I'm questioning the existence of this class. Shouldn't we be using `CurrencyAmount` instead?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/VaultDataDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * No-args constructor.
+   */
+  public VaultDataDeserializer() {
+    super(VaultData.class);
+  }
+
+  @Override
+  public VaultData deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+    return VaultData.of(jsonParser.getText());
```
sappenin, line 44: I'm not sure `VaultData` should be a String, but maybe it should? It's going to be binary data, so if it's a String it will be Hex, right?

(maybe this is fine, but I'm trying to think if `VaultData` should actually just hold an array (e.g., List of `UnsignedByte`).

That said, perhaps there's prior art in xrpl4j for this kind of thing?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
```diff
@@ -0,0 +1,198 @@
...
+  @JsonProperty("WithdrawalPolicy")
+  WithdrawalPolicy withdrawalPolicy();
+
+  /**
+   * Power of 10 for asset-to-share conversion.
+   *
+   * @return An optionally-present {@link AssetScale}.
+   */
+  @JsonProperty("Scale")
+  Optional<AssetScale> scale();
```
sappenin: In the spec, `scale` and most of the other values have are required (but default to 0).

Looks like only `data` and `assetsMaximum` are actually optional.

See https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065-single-asset-vault#212-fields

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultCreate.java
```diff
@@ -0,0 +1,129 @@
...
+      }
+    });
+
+    scale().ifPresent(scaleValue -> asset().handle(
+      xrpIssue -> {
+        throw new IllegalArgumentException("Scale is only allowed for IOU assets.");
+      },
+      iouIssue -> {
+        long scaleVal = scaleValue.value().longValue();
+        if (scaleVal < 0 || scaleVal > 18) {
```
sappenin: I think this line needs some more coverage.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/BinarySerializationTests.java
```diff
@@ -2340,12 +2341,12 @@ void serializeAmmBid() throws JsonProcessingException {
       .sequence(UnsignedInteger.ONE)
       .signingPublicKey(
         PublicKey.fromBase16EncodedPublicKey("02B4A8F64B97151FA303F86417751B7EA5AF1D0014FCC110C234D04AF15F3F654A"))
-      .asset(Issue.builder()
+      .asset(IouIssue.builder()
         .currency("4755534400000000000000000000000000000000")
         .issuer(Address.of("rPZtDb6ZHtfYzMmzyUGvehoi2vYhrNAPhy"))
         .build()
       )
-      .asset2(Issue.builder()
```
sappenin: This is a breaking change - see my other comment about how we should document breakage?

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/AssetAmountTest.java
```diff
@@ -0,0 +1,101 @@
+package org.xrpl.xrpl4j.model.transactions;
+
+/*-
+ * ========================LICENSE_START=================================
+ * xrpl4j :: core
+ * %%
+ * Copyright (C) 2020 - 2023 XRPL Foundation and its contributors
```
sappenin: Minor nit - If you're adding new files _and_ adding a license, can you make the date correct, like this:

```suggestion
 * Copyright (C) 2020 - 2026 XRPL Foundation and its contributors
```

(Alternatively, don't add this license and it will get taken care of, eventually, by https://github.com/XRPLF/xrpl4j/issues/669)

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/IssueType.java
Patel-Raj11: > why rippled adds an additional ACCOUNT_ONE into the binary encoding?

IOU - currency (20 bytes) || issuer (20 bytes)
MPT - issuer (20 bytes) || ACCOUNT_ONE (20 bytes) || sequence (4 bytes)

I think `ACCOUNT_ONE` acts as a discriminator between IOU and MPT for deserializer. Reading first 20 bytes helps us to distinguish between XRP (all zeros) and IOU/MPT, but it won't be possible to distinguish between IOU and MPT if `ACCOUNT_ONE` is not present.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/Issue.java
Patel-Raj11: I think since `mptIssuanceId` is Optional, it won't be a breaking change. Any user using `Issue` object present in `binary.types` package would still be able to create Issue object the way they have been creating it, ever after the upgrade. Did I understand it incorrectly?

There is a breaking change in Issue interface present in `model.ledger` package this PR, and I am planning to document it.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
Patel-Raj11: I added `AssetAmount` type to represent the fields that are of `Number` serialized type (eg. `AssetsMaximum`). Number serialized type is just a decimal number represented in string format. However, none of the concrete implementations of `CurrencyAmount` wraps `String` alone.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/VaultDataDeserializer.java
Patel-Raj11: `VaultData` is a wrapper type containing `String` and it has checks to make sure that the content is valid Hex.
I followed the convention that we have in xrpl4j for such metadata (`MpTokenMetadata`, `DidData` etc).

We could make this type wrap `List<UnsignedByte>` or `UnsignedByteArray` and update the serializer and deserialize to do bytes <-> String (Hex) conversion, but I think doing that won't make much difference as we will anyway provide static convenience methods like `fromHex` for users to create an instance of `VaultData` from off-chain binary data that they would have stored in Hex format.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultCreate.java
Patel-Raj11: `AssetScale` wraps `UnsignedInteger`, so I have updated the check accordingly and now this line has full coverage.

Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
Patel-Raj11: Thanks for highlighting this. I was marking all the`soeDEFAULT` fields as optional just to be on safer side. But now have updated them to default to zero value of their type. From `ledger_entries.macro` only 
`Data` is `soeOPTIONAL` other fields are either `soeREQUIRED` or `soeDEFAULT`.

Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/client/vault/VaultInfoResultTest.java
Patel-Raj11: Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/vault/VaultInfoRequestParams.java
Patel-Raj11: Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
Patel-Raj11: Will rebase and adjust once https://github.com/XRPLF/xrpl4j/pull/712 and https://github.com/XRPLF/xrpl4j/pull/715 are merged.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/IssueType.java
Patel-Raj11: Updated the code comment in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/AssetAmountTest.java
Patel-Raj11: Addressed in [5256ce8](https://github.com/XRPLF/xrpl4j/pull/713/commits/5256ce8e79ea2fa8b9bee104fc3513cc80ea4eb1) as one off. But fixing https://github.com/XRPLF/xrpl4j/issues/669 is the right approach.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/BinarySerializationTests.java
Patel-Raj11: Added in [95a0530](https://github.com/XRPLF/xrpl4j/pull/713/commits/95a0530ce4faeba50bb65a8216e2346c0063cfa4)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+
+  @Override
+  public NumberType fromParser(BinaryParser parser) {
+    return new NumberType(parser.read(WIDTH));
+  }
+
+  // xrpld implementation: src/libxrpl/basics/Number.cpp -> doNormalize
+  @Override
+  public NumberType fromJson(JsonNode node) {
+    String value = node.isInt() || node.isLong() ? String.valueOf(node.asLong()) : node.asText();
```
cybele-ripple: Can you add a test case for when `node.isInt()` returns true and `node.isLong()` returns true? Looks like just the `node.asText()` true case is tested

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultFlags.java
cybele-ripple, line 1: Should there be `VAULT_SHARE_NON_TRANSFERABLE` in this file? Looks like it's present in `model/flags/VaultCreateFlags.java`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/IssueDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+   * No-args constructor.
+   */
+  public IssueDeserializer() {
+    super(Issue.class);
+  }
+
+  @Override
+  public Issue deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+    JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+
```
cybele-ripple, line 52: Would there be an instance where `mpt_issuance_id` and `currency` are present? If so, maybe include something like 

```
  boolean hasMpt = node.has("mpt_issuance_id");
  boolean hasCurrency = node.has("currency");

  // Validate mutual exclusivity
  if (hasMpt && hasCurrency) {
    throw new IOException(
      "Invalid Issue JSON: cannot have both 'mpt_issuance_id' and 'currency' fields"
    );
  }
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultFlags.java
Patel-Raj11: Actually it's not present in the Flags returned in Vault ledger object as per the spec and rippled code. I remember discussing it with Bronek - https://ripple-enterprise.slack.com/archives/C070YDAAWKZ/p1758720696840199

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: I think instead I should remove these checks as the type would always be string since `AssetAmount` wraps String type.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/IssueDeserializer.java
Patel-Raj11: This codepath would get exercised when parsing the transaction/ledger object from rippled. The object returned would never have both `mpt_issuance_id` and `currency` present at the same time.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Removed these checks in [4d6b600](https://github.com/XRPLF/xrpl4j/pull/713/commits/4d6b60070d3c5978b30718a41361c3999232e9c9)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SingleAssetVaultIT.java
```diff
@@ -0,0 +1,1428 @@
...
+  /**
+   * Test vault with IOU as asset. Exercises all vault transactions and verifies vault state through
+   * vault_info, ledger_entry, and account_objects.
+   */
+  @Test
+  void vaultWithIouAsset() throws Exception {
+    // Step 1: Create accounts
+    KeyPair issuerKeyPair = createRandomAccountEd25519();
+    final KeyPair vaultOwnerKeyPair = createRandomAccountEd25519();
+    final KeyPair depositorKeyPair = createRandomAccountEd25519();
```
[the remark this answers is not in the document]
Patel-Raj11: `depositorKeyPair` is being used in the test.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SingleAssetVaultIT.java
```diff
@@ -0,0 +1,1428 @@
...
+    // Get vault ID from ledger entry
+    LedgerEntryResult<VaultObject> vaultLedgerEntry = xrplClient.ledgerEntry(
+      LedgerEntryRequestParams.vault(
+        VaultLedgerEntryParams.builder()
+          .owner(vaultOwnerKeyPair.publicKey().deriveAddress())
+          .seq(vaultCreate.sequence())
+          .build(),
+        LedgerSpecifier.VALIDATED
+      )
+    );
```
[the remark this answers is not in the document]
Patel-Raj11: It's being read

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
Patel-Raj11: Added migration guide - [31ba40e](https://github.com/XRPLF/xrpl4j/pull/713/commits/31ba40e5ddab9722c39735f889f948926f86bb8e)

xrpl4j-core/src/test/resources/codec-fixtures.json
```diff
@@ -5046,6 +5046,153 @@
         "TxnSignature": "AACD31A04CAE14670FC483A1382F393AA96B49C84479B58067F049FBD772999325667A6AA2520A63756EE84F3657298815019DD56A1AECE796B08535C4009C08",
         "URI": "6469645F6578616D706C65"
       }
+    },
+    {
+      "binary": "120041811401476926B590BA3245F63C829116A0A3AF7F382D0318000000000000000000000000555344000000000001476926B590BA3245F63C829116A0A3AF7F382D",
```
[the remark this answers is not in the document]
Patel-Raj11: The codec fixtures are to test the Serialized types instead of each transaction. Having VaultCreate tests covers all the new types that are introduced.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/IssueDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+      String mptIssuanceId = node.get("mpt_issuance_id").asText();
+      return MptIssue.builder()
+        .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
+        .build();
+    } else if (node.has("currency")) {
+      String currency = node.get("currency").asText();
+      if ("XRP".equals(currency)) {
+        return XrpIssue.XRP;
+      }
+      String issuer = node.get("issuer").asText();
```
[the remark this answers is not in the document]
Patel-Raj11: This JSON to Java object deserialization happens for the objects that are received from rippled. It would never be null in case of IOU asset. It's not something that user inputs and could be missing.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/vault/VaultInfoRequestParams.java
```diff
@@ -0,0 +1,122 @@
...
+   * @return An {@link Optional} {@link UnsignedInteger} containing the sequence number.
+   */
+  Optional<UnsignedInteger> seq();
+
+  /**
+   * Validates that either {@link #vaultId()} is present, or both {@link #owner()} and {@link #seq()} are present,
+   * but not both identification methods at the same time.
+   */
+  @Value.Check
+  default void check() {
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [e2f2730](https://github.com/XRPLF/xrpl4j/pull/713/commits/e2f2730be714ee8e84cd502500044f507173965f)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/IssueType.java
```diff
@@ -37,26 +70,54 @@ public IssueType fromJson(JsonNode node) throws JsonProcessingException {
...
+        Issue.builder().currency(currencyJson).build()
+      );
     }
 
-    return objectMapper.valueToTree(builder.build());
+    AccountIdType accountId = new AccountIdType().fromParser(parser);
+
+    // MPT
+    if (accountId.value().equals(ACCOUNT_ONE)) {
+      // Convert sequence from little-endian to bug-endian
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [e2f2730](https://github.com/XRPLF/xrpl4j/pull/713/commits/e2f2730be714ee8e84cd502500044f507173965f)

xrpl4j-core/src/test/resources/codec-fixtures.json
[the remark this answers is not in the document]
Patel-Raj11: The binary codec doesn't have transaction-specific encoding logic. Moreover, if we have missed any field encoding, then tests in SingleAssetVaultIT would be rejected by rippled, but currently they all pass. The main idea of adding the codec-fixtures is to do a sanity check if the tx_blob generated by xrpl4j matches with tx_blob generated by rippled and would encoding for any serialized type is not incorrect. So, we added only those transaction where new serialized fields were implemented/modified.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SingleAssetVaultIT.java
```diff
@@ -0,0 +1,1433 @@
...
```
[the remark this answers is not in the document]
Patel-Raj11: The way it's done correctly is better and more readable.

cybele-ripple (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/IssueType.java
```diff
@@ -3,14 +3,24 @@
...
 import org.xrpl.xrpl4j.codec.binary.serdes.BinaryParser;
+import org.xrpl.xrpl4j.model.AddressConstants;
+
+import java.util.regex.Pattern;
 
 public class IssueType extends SerializedType<IssueType> {
 
   private static final ObjectMapper objectMapper = BinaryCodecObjectMapperFactory.getObjectMapper();
 
+  private static final UnsignedByteArray ACCOUNT_ONE = AddressCodec.getInstance().decodeAccountId(
```
sappenin: For later, I wonder if we should add a utility method to `Address` to get the bytes as an `UnsignedByteArray`. Then we could get rid of this constant.

Just thinking out loud there, so not something to block this review. If you find that suggestion interesting, feel free to make a GH issue.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+  public NumberType() {
+    this(UnsignedByteArray.ofSize(WIDTH));
+  }
+
+  public NumberType(UnsignedByteArray bytes) {
+    super(bytes);
+  }
+
+  @Override
+  public NumberType fromParser(BinaryParser parser) {
```
sappenin:
```suggestion
  public NumberType fromParser(final BinaryParser parser) {
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+    return new NumberType(parser.read(WIDTH));
```
sappenin, line 61:
```suggestion
    Objects.requireNonNull(parser);
    return new NumberType(parser.read(WIDTH));
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  // xrpld implementation: src/libxrpl/basics/Number.cpp -> doNormalize
+  @Override
+  public NumberType fromJson(JsonNode node) {
+    String value = node.asText();
```
sappenin, line 69: Minor nit, check null here to fail-fast.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * The potential unrealized loss.
+   *
+   * @return An {@link AssetAmount}, defaulting to "0".
+   */
+  @JsonProperty("LossUnrealized")
+  @Value.Default
+  default AssetAmount lossUnrealized() {
+    return AssetAmount.of("0");
```
sappenin: So, after thinking about this over the weekend, I think we should introduce an `Amount` as a formal part of what would otherwise be a `CurrencyAmount`. However, that's too large of a refactor for `CurrencyAmount`, so I want to introduce something new and just deprecate `CurrencyAmount` to give people a chance to upgrade over a few xrpl4j versions. So, https://github.com/XRPLF/xrpl4j/pull/769 is a proposal for adding `Amount` and then creating a new `TokenAmount` (that would essentially be a `CurrencyAmount`). Take a look at that PR and let me know what you think.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
sappenin: Resolving this in-favor of a different conversation in this PR where we're working out the Java interface of an `AssetAmount` (AssetAmount stores only a numeric magnitude, whereas the actual "type" is stored in a different field of type `Issue`). therefore, as @Patel-Raj11 points out above, `CurrencyAmount` is not the appropriate type here.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/IssueDeserializer.java
sappenin: Resolving this as completed. @cybele-ripple feel free to re-open if you disagree.

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Addressed [193bf81](https://github.com/XRPLF/xrpl4j/pull/713/commits/193bf81bcbacf827e86aca2cbe8805fd11c72965)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Addressed [193bf81](https://github.com/XRPLF/xrpl4j/pull/713/commits/193bf81bcbacf827e86aca2cbe8805fd11c72965)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/NumberType.java
Patel-Raj11: Addressed [193bf81](https://github.com/XRPLF/xrpl4j/pull/713/commits/193bf81bcbacf827e86aca2cbe8805fd11c72965)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
Patel-Raj11: @sappenin #769 approach looks good to me. I'll pull in these changes in SAV PR and make use of `Amount` instead of `AssetAmount`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * Power of 10 for asset-to-share conversion.
+   *
+   * @return An {@link AssetScale}, defaulting to 0.
+   */
+  @JsonProperty("Scale")
+  @Value.Default
+  default AssetScale scale() {
+    return AssetScale.of(UnsignedInteger.ZERO);
```
sappenin, line 195: I created https://github.com/XRPLF/xrpl4j/issues/776 to track replacing this with a constant.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * The potential unrealized loss.
+   *
+   * @return An {@link Amount}, defaulting to "0".
+   */
+  @JsonProperty("LossUnrealized")
+  @Value.Default
+  default Amount lossUnrealized() {
+    return Amount.of("0");
```
sappenin: Let's make a constant for this, so we don't create all kinds of objects on every immutable construction. 

```
static Amount ZERO = Amount.of("0");
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Amount.java
```diff
@@ -1,6 +1,7 @@
 package org.xrpl.xrpl4j.model.transactions;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
```
sappenin: This is unused.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/VaultObject.java
Patel-Raj11: Addressed [0482545](https://github.com/XRPLF/xrpl4j/pull/713/commits/048254558fb93f868441004c6d6f7ed209467c8f)

sappenin (changes requested):

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/AmountTest.java
```diff
@@ -89,6 +89,51 @@ void isNegativeReturnsTrueForNegativeDecimal() {
...
+
+  @Test
+  void isZeroReturnsFalseForSmallDecimal() {
+    assertThat(Amount.of("0.001").isZero()).isFalse();
+  }
+
+  @Test
+  void zeroConstantIsZero() {
+    assertThat(Amount.ZERO.isZero()).isTrue();
+    assertThat(Amount.ZERO.value()).isEqualTo("0");
```
sappenin, line 134:
```suggestion
    assertThat(Amount.ZERO.value()).isEqualTo("0");
    
    assertThat(Amount.ZERO).isEqualTo(Amount.of("0"));
    assertThat(Amount.ZERO).isEqualTo(Amount.of("-0"));
    assertThat(Amount.ZERO).isEqualTo(Amount.of("0.00"));
    assertThat(Amount.ZERO).isEqualTo(Amount.of("0e5"));
```

Patel-Raj11: Addressed [017bbb8](https://github.com/XRPLF/xrpl4j/pull/713/commits/017bbb85f9e84f5dad29eea43e1ac225c95debaf)

sappenin (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultCreate.java
```diff
@@ -0,0 +1,129 @@
...
+    });
+
+    scale().ifPresent(scaleValue -> asset().handle(
+      xrpIssue -> {
+        throw new IllegalArgumentException("Scale is only allowed for IOU assets.");
+      },
+      iouIssue -> {
+        long scaleVal = scaleValue.value().longValue();
+        if (scaleVal > 18) {
+          throw new IllegalArgumentException("Scale must be between 0 and 18.");
```
cybele-ripple, line 120: Do we need to include a check like `scaleVal > 18 || scaleVal < 0`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultClawback.java
```diff
@@ -0,0 +1,84 @@
...
+  @JsonProperty("Holder")
+  Address holder();
+
+  /**
+   * The amount to clawback. When Amount is 0 clawback all funds, up to the total shares the Holder owns.
+   *
+   * @return An optionally-present {@link CurrencyAmount}.
+   */
+  @JsonProperty("Amount")
+  Optional<CurrencyAmount> amount();
```
cybele-ripple, line 66: Javadoc specifies `When Amount is 0 clawback all funds` but the field is optional. Should the javadoc say `When Amount is absent, clawback all funds`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultCreate.java
Patel-Raj11: AssetScale wraps `UnsignedInteger`, shouldn't be -ve.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultClawback.java
Patel-Raj11: Yes, skipping the Amount field will clawback all the funds but the Javadoc verbatim taken from XLS spec explicitly emphasize that 0 value will clawback all the funds. Added /absent in addition to 0 - [a17cf7c](https://github.com/XRPLF/xrpl4j/pull/713/commits/a17cf7c1f6b0c4b55f2438dbe5bcc00941aaf9a5)

sappenin (approved):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultClawback.java
sappenin: Resolving this comment as fixed per the comment above from @Patel-Raj11. @cybele-ripple please re-open if you see something still unresolved.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultCreate.java
sappenin: Resolving this comment as fixed per the comment above from @Patel-Raj11. @cybele-ripple please re-open if you see something still unresolved.

PR #719, opened 2026-04-01
Add support for Lending Protocol - XLS-66

Patel-Raj11:
XLS spec - https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0066-lending-protocol

**Transactions added/updated:**

- LoanBrokerSet
- LoanBrokerDelete
- LoanBrokerCoverDeposit
- LoanBrokerCoverWithdraw
- LoanBrokerCoverClawback
- LoanSet
- LoanDelete
- LoanManage
- LoanPay

**Ledger objects added/updated:**

- LoanObject
- LoanBrokerObject

**Flags added:**

- LoanSetFlags
- LoanManageFlags
- LoanPayFlags
- LoanFlags (ledger entry flags)

**Signing added/updated:**

- TransactionSigner (added `counterpartySign()` and `counterpartyMultiSign()` for counterparty signing)

**Binary codec added/updated:**

- XrplBinaryCodec (added `encodeForMultiSigningWithSigningPubKey()` — preserves `SigningPubKey` unlike `encodeForMultiSigning`)

**Fees added/updated:**

- FeeUtils (added `computeLoanSetNetworkFees()` — accounts for extra counterparty signer in fee calculation)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/AbstractTransactionSigner.java
```diff
@@ -115,6 +116,27 @@ public Signature multiSignInner(final P privateKeyable, final Batch batchTransac
     return this.signingHelper(privateKeyable, signableBytes);
   }
 
+  @Override
+  public Signature counterpartySign(final P privateKeyable, final LoanSet transaction) {
+    Objects.requireNonNull(privateKeyable);
+    Objects.requireNonNull(transaction);
+
+    final UnsignedByteArray signableTransactionBytes = this.signatureUtils.toSignableBytes(transaction);
```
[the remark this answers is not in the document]
Patel-Raj11: When the counterparty single-signs the transaction, there is no special handling needed. We have introduced a new method called `counterpartySign` because it helps the user see that if they are sining as a counterParty they can use this and expect a `Signature` out of it, instead of a `SingleSignedTransaction<T>` (sign method returns this).

Using existing `sign` method would be incorrect since the counterparty's signature goes into `TxnSignature`  present in`CounterpartySignature` as opposed to `TxnSignature` present at the root level.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
```diff
@@ -0,0 +1,160 @@
...
+      !(hasDirectSigning && hasMultiSig),
+      "CounterpartySignature cannot have both direct signing fields and Signers array"
+    );
+
+    if (hasMultiSig && !sortedSigners()) {
+      // Normalize the order of the signers by account address (required by XRPL)
+      return ImmutableCounterpartySignature.builder()
+        .from(this)
+        .signers(signers().stream()
+          .sorted(Comparator.comparing(signature -> new BigInteger(
```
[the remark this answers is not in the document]
Patel-Raj11: `AddressCodec.getInstance().decodeAccountId(signature.signer().account()).hexValue()` will always give 20 byte value. So this concern doesn't apply here.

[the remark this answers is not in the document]
Patel-Raj11: `Account` is derived from `SigningPubKey`. If `decodeAccountId` throws an exception, then there is a more severe problem in deriving an account from `SigningPubKey` than the potential unchecked exception being thrown here. I think such verbose checks are unnecessary here.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,279 @@
+package org.xrpl.xrpl4j.model.transactions;
```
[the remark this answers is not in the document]
Patel-Raj11: `isSigningField` of CounterpartySignature is marked as `false`. So when the transaction is encoded for signing, this fields will never be included. So it's presence/absence doesn't matter. The user should not worry about it's presence or absence in the LoanSet transaction for the signing purposes.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,279 @@
...
+      "InterestRate must be between 0 and 100000 inclusive."
+    ));
+    overpaymentFee().ifPresent(rate -> Preconditions.checkArgument(
+      rate.compareTo(UnsignedInteger.valueOf(100000)) <= 0,
+      "OverpaymentFee must be between 0 and 100000 inclusive."
+    ));
+    lateInterestRate().ifPresent(rate -> Preconditions.checkArgument(
+      rate.compareTo(UnsignedInteger.valueOf(100000)) <= 0,
+      "LateInterestRate must be between 0 and 100000 inclusive."
+    ));
```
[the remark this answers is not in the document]
Patel-Raj11: `Counterparty` field is itself optional — when the broker initiates and omits it, rippled assumes the counterparty is LoanBroker.Owner. So we'd also need ledger state to validate that case, which we don't have
at model construction time and we don't do such checks that requires current ledger state. Current sets of checks are sufficient and in case of incorrect signatures, xrpld node will fail the transaction.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodec.java
```diff
@@ -208,6 +208,35 @@ public UnsignedByteArray encodeForBatchInnerMultiSigning(Batch batch, Address si
...
+   * {@link #encodeForMultiSigning(String, String)}.</p>
+   *
+   * @param json         A {@link String} containing JSON to be encoded.
+   * @param xrpAccountId A {@link String} containing the XRPL AccountId of the signer.
+   *
+   * @return hex encoded representation
+   *
+   * @throws JsonProcessingException if JSON is not valid.
+   */
+  public String encodeForMultiSigningWithSigningPubKey(
```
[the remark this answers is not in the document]
Patel-Raj11: Adding `@Beta` to a codec-level method would be misleading — it suggests the encoding itself is unstable, when really it's just a different serialization path that the protocol requires. The amendment dependency is on the transaction type (`LoanSet`), not on the encoding function. The function does exactly what it says: encodes for multi-signing while preserving `SigningPubKey`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,279 @@
...
+    );
+
+    // 8. LoanOriginationFee must not be negative and must not exceed PrincipalRequested.
+    loanOriginationFee().ifPresent(fee -> {
+      Preconditions.checkArgument(
+        !fee.isNegative(),
+        "LoanOriginationFee must not be negative."
+      );
+      Preconditions.checkArgument(
+        new BigDecimal(fee.value()).compareTo(new BigDecimal(principalRequested().value())) <= 0,
```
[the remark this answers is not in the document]
Patel-Raj11: Added a comment in [22de7b0](https://github.com/XRPLF/xrpl4j/pull/719/commits/22de7b0e638cc8e3cee26623518d23be146f5a97)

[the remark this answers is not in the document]
Patel-Raj11: Addressed in [22de7b0](https://github.com/XRPLF/xrpl4j/pull/719/commits/22de7b0e638cc8e3cee26623518d23be146f5a97)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Validates that the {@code CounterpartySignature} has either direct signing fields or multi-sig signers (but not
+   * both), and normalizes the signer order by account address if multi-signing.
+   *
+   * @return A normalized {@link CounterpartySignature}.
+   */
+  @Value.Check
+  default CounterpartySignature checkAndNormalize() {
+    boolean hasDirectSigning = signingPubKey().isPresent() && txnSignature().isPresent();
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [22de7b0](https://github.com/XRPLF/xrpl4j/pull/719/commits/22de7b0e638cc8e3cee26623518d23be146f5a97)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -852,6 +866,25 @@ public String toString() {
...
+    }
+
+    /**
+     * Indicates whether this amount is zero. Handles all string representations of zero (e.g., "0", "0.0", "0.000",
+     * "0e0", "-0").
+     *
+     * @return {@code true} if this amount is zero; {@code false} otherwise.
+     */
+    public boolean isZero() {
+      return new BigDecimal(this.value()).signum() == 0;
```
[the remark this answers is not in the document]
Patel-Raj11: Preconditions checks added in _AssetAmount itself will prevent this now. - [22de7b0](https://github.com/XRPLF/xrpl4j/pull/719/commits/22de7b0e638cc8e3cee26623518d23be146f5a97)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
```diff
@@ -0,0 +1,1237 @@
...
+      .principalRequested(AssetAmount.of("50000"))
+      .counterparty(borrowerKeyPair.publicKey().deriveAddress())
+      .paymentTotal(UnsignedInteger.valueOf(3))
+      .data(LoanData.of("AABBCC"))
+      .signingPublicKey(loanBrokerKeyPair.publicKey())
+      .build();
+
+    // Dual-signing: borrower signs the same transaction payload
+    // Since CounterpartySignature has isSigningField=false, both parties sign the same bytes
+    SingleSignedTransaction<LoanSet> borrowerSigned = signatureService.sign(
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [e71f59f](https://github.com/XRPLF/xrpl4j/pull/719/commits/e71f59f6a7380dd65191d49606ca8996c2861be0)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -184,6 +184,47 @@ protected static XrpCurrencyAmount computeBatchFee(
...
+   * @return A {@link ComputedNetworkFees} with low, medium and high fee levels scaled for the LoanSet transaction.
+   */
+  public static ComputedNetworkFees computeLoanSetNetworkFees(
+    final FeeResult feeResult,
+    final UnsignedInteger numFirstPartySigners,
+    final UnsignedInteger numCounterpartySigners
+  ) {
+    Objects.requireNonNull(feeResult);
+    Objects.requireNonNull(numFirstPartySigners);
+    Objects.requireNonNull(numCounterpartySigners);
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [e71f59f](https://github.com/XRPLF/xrpl4j/pull/719/commits/e71f59f6a7380dd65191d49606ca8996c2861be0)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Validates LoanSet data verification preconditions.
+   */
+  @Value.Check
+  default void check() {
+    // 1. LoanBrokerID must not be zero.
+    Preconditions.checkArgument(
+      !loanBrokerId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Since `0000000000000000000000000000000000000000000000000000000000000000` is used throughout this PR, create a constant.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerSet.java
```diff
@@ -0,0 +1,173 @@
...
+  default void check() {
+    // VaultID must not be all zeros
+    Preconditions.checkArgument(
+      !vaultId().value().equals("0000000000000000000000000000000000000000000000000000000000000000"),
+      "VaultID must not be zero."
+    );
+
+    // LoanBrokerID, if present, must not be all zeros
+    loanBrokerId().ifPresent(id -> Preconditions.checkArgument(
+      !id.value().equals("0000000000000000000000000000000000000000000000000000000000000000"),
```
cybele-ripple: Since `0000000000000000000000000000000000000000000000000000000000000000` is used throughout this PR, create a constant.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanDelete.java
```diff
@@ -0,0 +1,63 @@
...
+  Hash256 loanId();
+
+  /**
+   * Validates LoanDelete data verification preconditions.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Since `0000000000000000000000000000000000000000000000000000000000000000` is used throughout this PR, create a constant.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerDelete.java
```diff
@@ -0,0 +1,63 @@
...
+  Hash256 loanBrokerId();
+
+  /**
+   * Validates LoanBrokerDelete data verification preconditions.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanBrokerId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanManage.java
```diff
@@ -0,0 +1,72 @@
...
+  Hash256 loanId();
+
+  /**
+   * Validates LoanManage data verification preconditions per the Lending Protocol spec section 3.10.4.1.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanPay.java
```diff
@@ -0,0 +1,85 @@
...
+  CurrencyAmount amount();
+
+  /**
+   * Validates LoanPay data verification preconditions per the Lending Protocol spec section 3.11.4.1.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverClawback.java
```diff
@@ -0,0 +1,115 @@
...
+  default void check() {
+    // 1. At least one of LoanBrokerID or Amount must be specified.
+    Preconditions.checkArgument(
+      loanBrokerId().isPresent() || amount().isPresent(),
+      "At least one of LoanBrokerID or Amount must be specified."
+    );
+
+    // 2. LoanBrokerID, if present, must not be zero.
+    loanBrokerId().ifPresent(id -> Preconditions.checkArgument(
+      !id.value().equals("0000000000000000000000000000000000000000000000000000000000000000"),
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverDeposit.java
```diff
@@ -0,0 +1,76 @@
...
+  CurrencyAmount amount();
+
+  /**
+   * Validates LoanBrokerCoverDeposit data verification preconditions.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanBrokerId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverWithdraw.java
```diff
@@ -0,0 +1,101 @@
...
+  Optional<UnsignedInteger> destinationTag();
+
+  /**
+   * Validates LoanBrokerCoverWithdraw data verification preconditions.
+   */
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(
+      !loanBrokerId().value().equals(
+        "0000000000000000000000000000000000000000000000000000000000000000"
```
cybele-ripple: Set `0000000000000000000000000000000000000000000000000000000000000000` as a constant

cybele-ripple:
General note: There are multiple places where `0000000000000000000000000000000000000000000000000000000000000000` is used (`LoanSet.java`, `LoanBrokerSet.java`, `LoanDelete.java`, `LoanBrokerDelete.java`, `LoanManage.java`, `LoanPay.java`, `LoanBrokerCoverClawback.java`, `LoanBrokerCoverDeposit.java`, `LoanBrokerCoverWithdraw.java`). We should put this value in a constant and then import this constant where needed throughout this PR

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerSet.java
```diff
@@ -0,0 +1,173 @@
...
+   * Arbitrary metadata in hex format. The field is limited to 256 bytes.
+   *
+   * @return An optionally-present {@link LoanBrokerData}.
+   */
+  @JsonProperty("Data")
+  Optional<LoanBrokerData> data();
+
+  /**
+   * The 1/10th basis point fee charged by the Lending Protocol Owner. Valid values are between 0 and 10000
+   * inclusive (1% - 10%).
```
cybele-ripple: Since valid values are between 0 and 10000 inclusive, then should the percentage start at 0% instead of 1%?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -975,6 +1026,142 @@ public int hashCode() {
...
+      return false;
+    }
+
+    @Override
+    public int hashCode() {
+      return value().toUpperCase(Locale.ENGLISH).hashCode();
+    }
+
+  }
+
```
cybele-ripple: LoadData and LoanBrokerData appear nearly identical (except for the error message). Can we streamline is code?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/metadata/MetaLoanBrokerObject.java
```diff
@@ -0,0 +1,177 @@
...
+@Beta
+public interface MetaLoanBrokerObject extends MetaLedgerObject {
+
+  /**
+   * Ledger object flags.
+   *
+   * @return An optionally-present {@link UnsignedInteger}.
+   */
+  @JsonProperty("Flags")
+  Optional<UnsignedInteger> flags();
```
cybele-ripple: In the [`MetaLoanObject.java`](https://github.com/XRPLF/xrpl4j/pull/719/changes#diff-356b81e39ab25f68d9d727b3f30c3712911929f09f4abfc332bf94ce14bdd4cbR29-R35) file, the ledger object flags use `LoanFlags` instead of `UnsignedInteger`. Is this an issue?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverClawback.java
```diff
@@ -0,0 +1,115 @@
...
+        !(amt instanceof XrpCurrencyAmount),
+        "Amount must not be XRP."
+      );
+
+      // 6. LoanBrokerID absent + MPT amount is invalid.
+      if (!loanBrokerId().isPresent() && amt instanceof MptCurrencyAmount) {
+        throw new IllegalArgumentException(
+          "LoanBrokerID must be specified when Amount is an MPT."
+        );
+      }
```
cybele-ripple, line 99: nit: count skips 5 (does from 4 to 6 instead)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/LoanBrokerObjectTest.java
cybele-ripple, line 1: Create a constant for `79E25403E9FC010A277D80410EED5494FDD033A09FD4C1432335A1734A1D099D` at the beginning of the file since this value is used multiple times here

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/LoanBrokerCoverClawbackJsonTest.java
cybele-ripple, line 1: Consider creating constants for `ED1F863E4E0957C6965B7B0563D56C77E4F68D571E8026251834F0ADBB0411D6FD` and `79E25403E9FC010A277D80410EED5494FDD033A09FD4C1432335A1734A1D099D`

cybele-ripple: Looks like these values appear in other files as well. Maybe create this constant in some place where multiple files can easily reuse this value?

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/ledger/LoanBrokerObjectTest.java
Patel-Raj11: I think moving this to a class level constant would be appropriate if these were unit tests and had specific meaning across the tests. However, these are JSON tests and the value has no particular meaning. I think it's better to leave it here.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/json/LoanBrokerCoverClawbackJsonTest.java
Patel-Raj11: These have no particular significance and are readable in the current form along side the JSON string.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverClawback.java
Patel-Raj11: Addressed [c033617](https://github.com/XRPLF/xrpl4j/pull/719/commits/c033617eb8adfcc2b65867dc1fae9efd204051a1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/metadata/MetaLoanBrokerObject.java
Patel-Raj11: It would always be 0, but good to have `Flags` type. Addressed here [c033617](https://github.com/XRPLF/xrpl4j/pull/719/commits/c033617eb8adfcc2b65867dc1fae9efd204051a1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
Patel-Raj11: Good point! Created an abstract parent class with common methods in it. Addressed here [c033617](https://github.com/XRPLF/xrpl4j/pull/719/commits/c033617eb8adfcc2b65867dc1fae9efd204051a1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerSet.java
Patel-Raj11: Addressed here - [c033617](https://github.com/XRPLF/xrpl4j/pull/719/commits/c033617eb8adfcc2b65867dc1fae9efd204051a1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanBrokerCoverWithdraw.java
Patel-Raj11: Addressed for every occurence [c033617](https://github.com/XRPLF/xrpl4j/pull/719/commits/c033617eb8adfcc2b65867dc1fae9efd204051a1)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanManage.java
```diff
@@ -0,0 +1,70 @@
...
+      "LoanID must not be zero."
+    );
+
+    // tfLoanDefault, tfLoanImpair, tfLoanUnimpair are mutually exclusive.
+    int flagCount = (flags().tfLoanDefault() ? 1 : 0) +
+      (flags().tfLoanImpair() ? 1 : 0) +
+      (flags().tfLoanUnimpair() ? 1 : 0);
+    Preconditions.checkArgument(
+      flagCount <= 1,
+      "Only one of tfLoanDefault, tfLoanImpair, or tfLoanUnimpair may be set."
```
cybele-ripple: Should this be `flagCount == 1`? The string on line 67 implies that a flagCount of 0 is not expected

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanPay.java
```diff
@@ -0,0 +1,83 @@
...
+      "Amount must be greater than zero."
+    );
+
+    // tfLoanLatePayment, tfLoanFullPayment, tfLoanOverpayment are mutually exclusive.
+    int flagCount = (flags().tfLoanLatePayment() ? 1 : 0) +
+      (flags().tfLoanFullPayment() ? 1 : 0) +
+      (flags().tfLoanOverpayment() ? 1 : 0);
+    Preconditions.checkArgument(
+      flagCount <= 1,
+      "Only one of tfLoanLatePayment, tfLoanFullPayment, or tfLoanOverpayment may be set."
```
cybele-ripple: Same issue: should this be `flagCount==1`? Either this check should be updated or the string on line 80 should be updated to account for a possible `flagCount` of 0

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanBrokerObject.java
```diff
@@ -0,0 +1,208 @@
...
+  @JsonProperty("ManagementFeeRate")
+  Optional<UnsignedInteger> managementFeeRate();
+
+  /**
+   * The total asset amount the protocol owes the Vault, including interest.
+   *
+   * @return An optionally-present {@link Amount}.
+   */
+  @JsonProperty("DebtTotal")
+  Optional<Amount> debtTotal();
```
sappenin: In the spec, the only two truly optional fields are `ManagementFeeRate ` and `Data`. All other fields, including this field, are marked as `required=yes` with a default value of 0. So, it seems like we should change all of the non-optional fields to be like this:

```java
@JsonProperty("DebtMaximum")
default Amount debtTotal(){
  return Amount.ZERO;
}
```

The one exclusion to this (that I can see in the spec) is `DebtMaximum`, which has the comment "_The default value of 0 means there is no limit to the debt._"

Because of that one comment, `DebtMaximum` seems like the only truly `Optional` field, and should be left alone.

Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
```diff
@@ -0,0 +1,296 @@
...
+  @JsonProperty("TotalValueOutstanding")
+  Optional<Amount> totalValueOutstanding();
+
+  /**
+   * The principal amount that the Borrower still owes.
+   *
+   * @return An optionally-present {@link Amount}.
+   */
+  @JsonProperty("PrincipalOutstanding")
+  Optional<Amount> principalOutstanding();
```
sappenin: In the spec, most of these fields (except `LoanScale`) are non-optional. Seems like these should just be `default` values with `Amount.ZERO` as the default.

Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
```diff
@@ -0,0 +1,166 @@
...
+      .build();
+  }
+
+  /**
+   * The public key used by the counterparty to sign the transaction.
+   *
+   * @return An optionally-present {@link String}.
+   */
+  @JsonProperty("SigningPubKey")
+  Optional<String> signingPubKey();
```
sappenin: I think we should use `PublicKey` here, like we do in `BatchSigner`. That way, developers are guided around how exactly to assemble bytes (no `String`), and serde does all the work.

Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * The counterparty's transaction signature.
+   *
+   * @return An optionally-present {@link String}.
+   */
+  @JsonProperty("TxnSignature")
+  Optional<String> txnSignature();
```
sappenin: Can we use `Signature` here?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanManage.java
sappenin: Let's make sure we have unit test coverage for whatever the final decision is.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -892,49 +913,43 @@ public String toString() {
...
+   * Provides shared validation, equality, and string handling for metadata types.
+   * Subclasses must override {@link #maxBytes()} to specify their size limit.
    */
-  @Value.Immutable
-  @Wrapped
-  @JsonSerialize(as = VaultData.class, using = VaultDataSerializer.class)
-  @JsonDeserialize(as = VaultData.class, using = VaultDataDeserializer.class)
-  @Beta
-  abstract static class _VaultData extends Wrapper<String> implements Serializable {
+  abstract static class Metadata extends Wrapper<String> implements Serializable {
```
sappenin, line 920: We need a unit test that covers `Metadata` directly (not implicitly, such as `VaultDataTest` or otherwise)

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
Patel-Raj11: Addressed [0f523c5](https://github.com/XRPLF/xrpl4j/pull/719/commits/0f523c5107d0053d0f09d0f35b1562a4f59a5615)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
Patel-Raj11: Addressed [0f523c5](https://github.com/XRPLF/xrpl4j/pull/719/commits/0f523c5107d0053d0f09d0f35b1562a4f59a5615)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/CounterpartySignature.java
Patel-Raj11: Yes makes sense to use `PublicKey` and `Signature`
[0f523c5](https://github.com/XRPLF/xrpl4j/pull/719/commits/0f523c5107d0053d0f09d0f35b1562a4f59a5615)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanBrokerObject.java
Patel-Raj11: LoanBroker object has either `soeDEFAULT` or `soeREQUIRED` fields, as per [ledger_entries.macro](https://github.com/XRPLF/rippled/blob/182d844996ff56201a63318542c646da05c2593d/include/xrpl/protocol/detail/ledger_entries.macro#L503). So, I defaulted to Zero value for most of the `soeDEFAULT` fields. 

Even though `Data` as marked as `soeDEFAULT`, it makes sense to keep it optional as it doesn't make sense to have return an empty string as `Metadata` if present cannot be empty.

I agree to keep `DebtMaximum` Optional in Java for the reason you mentioned.

Addressed [0f523c5](https://github.com/XRPLF/xrpl4j/pull/719/commits/0f523c5107d0053d0f09d0f35b1562a4f59a5615)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
Patel-Raj11: Loan object has either soeDEFAULT or soeREQUIRED fields, as per [ledger_entries.macro](https://github.com/XRPLF/rippled/blob/182d844996ff56201a63318542c646da05c2593d/include/xrpl/protocol/detail/ledger_entries.macro#L527). So I have given default zero value for each type accordingly.

LoanScale is also `soeDEFAULT`, so I have defaulted it's value to 0. Does this sound right for `LoanScale`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanPay.java
Patel-Raj11: Technically the transaction passes at rippled even if none of the flags are set. So I have updated the error message to depict the same in [0f523c5](https://github.com/XRPLF/xrpl4j/pull/719/commits/0f523c5107d0053d0f09d0f35b1562a4f59a5615). Also, there are unit tests present to test this.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanManage.java
Patel-Raj11: See the comment in `LoanPay`. It applies here as well.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtils.java
```diff
@@ -169,6 +170,34 @@ public UnsignedByteArray toSignableInnerBytes(final Batch batch) {
...
+   * @return An {@link UnsignedByteArray}.
+   */
+  @Beta
+  public UnsignedByteArray toCounterpartyMultiSignableBytes(final LoanSet transaction, final Address signerAddress) {
+    Objects.requireNonNull(transaction);
+    Objects.requireNonNull(signerAddress);
+
+    try {
+      final String unsignedJson = objectMapper.writeValueAsString(transaction);
+      final String unsignedBinaryHex = binaryCodec.encodeForMultiSigningWithSigningPubKey(
```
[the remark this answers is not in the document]
Patel-Raj11: when both parties multi-sign a LoanSet, the first-party's SigningPubKey is legitimately empty. The method correctly preserves whatever value is there (including empty), and that's the intended behavior.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/signing/AbstractTransactionSigner.java
```diff
@@ -112,7 +112,43 @@ public Signature multiSignInner(final P privateKeyable, final Batch batchTransac
...
-    return this.signingHelper(privateKeyable, signableBytes);
+    return this.signatureHelper(privateKeyable, signableBytes);
+  }
+
+  @Override
+  public Signature counterpartySign(final P privateKeyable, final LoanSet transaction) {
+    Objects.requireNonNull(privateKeyable);
+    Objects.requireNonNull(transaction);
+
+    return signatureHelper(privateKeyable, transaction);
```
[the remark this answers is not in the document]
Patel-Raj11: Yes, that is correct and intended behaviour. The separate method exists to make it clear for the user that one exists for counterparty sign and the existing `sign` method, returns a `SingleSignedTransaction` which is what we don't want.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodec.java
```diff
@@ -208,6 +208,35 @@ public UnsignedByteArray encodeForBatchInnerMultiSigning(Batch batch, Address si
...
+    String json, String xrpAccountId
+  ) throws JsonProcessingException {
+    JsonNode node = BINARY_CODEC_OBJECT_MAPPER.readTree(json);
+    if (!node.isObject()) {
+      throw new IllegalArgumentException("JSON object required for signing");
+    }
+    // NOTE: Unlike encodeForMultiSigning, we do NOT clear SigningPubKey here.
+    String suffix = new AccountIdType().fromJson(new TextNode(xrpAccountId)).toHex();
+    return TRX_MULTI_SIGNATURE_PREFIX + encode(removeNonSigningFields(node)) + suffix;
```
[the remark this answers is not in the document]
Patel-Raj11: when both parties multi-sign a LoanSet, the first-party's SigningPubKey is legitimately empty. The method correctly preserves whatever value is there (including empty), and that's the intended behavior.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/XrplBinaryCodec.java
```diff
@@ -208,6 +208,35 @@ public UnsignedByteArray encodeForBatchInnerMultiSigning(Batch batch, Address si
...
+   */
+  public String encodeForMultiSigningWithSigningPubKey(
+    String json, String xrpAccountId
+  ) throws JsonProcessingException {
+    JsonNode node = BINARY_CODEC_OBJECT_MAPPER.readTree(json);
+    if (!node.isObject()) {
+      throw new IllegalArgumentException("JSON object required for signing");
+    }
+    // NOTE: Unlike encodeForMultiSigning, we do NOT clear SigningPubKey here.
+    String suffix = new AccountIdType().fromJson(new TextNode(xrpAccountId)).toHex();
```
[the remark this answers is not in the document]
Patel-Raj11: when both parties multi-sign a LoanSet, the first-party's SigningPubKey is legitimately empty. The method correctly preserves whatever value is there (including empty), and that's the intended behavior. JS implementation does not validate this.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,278 @@
...
+  @JsonProperty("Counterparty")
+  Optional<Address> counterparty();
+
+  /**
+   * The signature of the counterparty over the transaction.
+   *
+   * @return An optionally-present {@link CounterpartySignature}.
+   */
+  @JsonProperty("CounterpartySignature")
+  Optional<CounterpartySignature> counterpartySignature();
```
sappenin, line 86: Per @Patel-Raj11 this will become non-optional.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * The Loan Broker ID associated with the loan.
+   *
+   * @return A {@link Hash256}.
+   */
+  @JsonProperty("LoanBrokerID")
+  Hash256 loanBrokerId();
+
+  /**
+   * The address of the counterparty of the Loan.
```
sappenin:
```suggestion
   * The address of the counterparty of the Loan.  This field may be absent when Loanbroker is the counterparty.
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
```diff
@@ -0,0 +1,341 @@
...
+   * @return An {@link Amount}.
+   */
+  @JsonProperty("TotalValueOutstanding")
+  @Value.Default
+  default Amount totalValueOutstanding() {
+    return Amount.ZERO;
+  }
+
+  /**
+   * The principal amount that the Borrower still owes.
```
sappenin, line 393: Let's update this (and other fields) javadoc to explain why this field is not a Java Optional:

```
/**
 * The principal amount that the Borrower still owes.
 *
 * <p>On-ledger, this field is marked {@code soeDEFAULT} to save space, which means it
 * is omitted from JSON responses when its value is {@code 0} (again, to save space),
 * even though the spec defines it as a required field. To handle this, {@code @Default}
 * is used so that Immutables populates the correct {@code 0} value when the field is
 * absent from a response.
 *
 * <p>Note: this field must <em>not</em> be made {@link Optional}, since the spec
 * defines it as required and a missing value semantically means {@code 0}, not "absent".
 */
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
```diff
@@ -0,0 +1,341 @@
...
+  default UnsignedInteger paymentRemaining() {
+    return UnsignedInteger.ZERO;
+  }
+
+  /**
+   * The total outstanding value of the Loan, including all fees and interest.
+   *
+   * @return An {@link Amount}.
+   */
+  @JsonProperty("TotalValueOutstanding")
```
sappenin, line 386: /**
 * The total outstanding value of the Loan, including all fees and interest.
 *
 * <p>On-ledger, this field is marked {@code soeDEFAULT} to save space, which means it
 * is omitted from JSON responses when its value is {@code 0} (again, to save space),
 * even though the spec defines it as a required field. To handle this, {@code @Default}
 * is used so that Immutables populates the correct {@code 0} value when the field is
 * absent from a response.
 *
 * <p>Note: this field must <em>not</em> be made {@link Optional}, since the spec
 * defines it as required and a missing value semantically means {@code 0}, not "absent".
 */

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  /**
+   * The remaining Management Fee owed to the LoanBroker.
+   *
+   * @return An {@link Amount}.
+   */
+  @JsonProperty("ManagementFeeOutstanding")
+  @Value.Default
+  default Amount managementFeeOutstanding() {
```
sappenin, line 428: /**
 * The remaining Management Fee owed to the LoanBroker.
 *
 * <p>On-ledger, this field is marked {@code soeDEFAULT} to save space, which means it
 * is omitted from JSON responses when its value is {@code 0} (again, to save space),
 * even though the spec defines it as a required field. To handle this, {@code @Default}
 * is used so that Immutables populates the correct {@code 0} value when the field is
 * absent from a response.
 *
 * <p>Note: this field must <em>not</em> be made {@link Optional}, since the spec
 * defines it as required and a missing value semantically means {@code 0}, not "absent".
 */

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
```diff
@@ -0,0 +1,341 @@
...
+  @JsonProperty("PeriodicPayment")
+  Amount periodicPayment();
+
+  /**
+   * The scale factor that ensures all computed amounts are rounded to the same number of decimal places. It
+   * is determined based on the total loan value at creation time.
+   *
+   * @return An {@link Integer}.
+   */
+  @JsonProperty("LoanScale")
```
sappenin, line 455: /**
 * The scale factor that ensures all computed amounts are rounded to the same number of decimal places. It
   * is determined based on the total loan value at creation time.
 *
 * <p>On-ledger, this field is marked {@code soeDEFAULT} to save space, which means it
 * is omitted from JSON responses when its value is {@code 0} (again, to save space),
 * even though the spec defines it as a required field. To handle this, {@code @Default}
 * is used so that Immutables populates the correct {@code 0} value when the field is
 * absent from a response.
 *
 * <p>Note: this field must <em>not</em> be made {@link Optional}, since the spec
 * defines it as required and a missing value semantically means {@code 0}, not "absent".
 */

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/LoanObject.java
Patel-Raj11: Addressed all JavaDoc comments in [65158ab](https://github.com/XRPLF/xrpl4j/pull/719/commits/65158abd91e6b1e1c043c754b857de3102d0b40b)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
Patel-Raj11: Addressed in [65158ab](https://github.com/XRPLF/xrpl4j/pull/719/commits/65158abd91e6b1e1c043c754b857de3102d0b40b)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
```diff
@@ -0,0 +1,1229 @@
...
+  // //////////////////////
+  // LoanSet counterparty signing tests (XRP vault, broker initiates, borrower is counterparty)
+  // //////////////////////
+
+  // //////////////////////
+  // Test 1: Broker single-signs, Borrower (counterparty) single-signs
+  // //////////////////////
+
+  @Test
+  void loanSetWithSingleSignBrokerAndSingleSignCounterparty()
```
sappenin, line 91: I wonder if we should have some tests that validate Counterparty RegularKey scenarios?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
Patel-Raj11: After going though this again, I found out that this field is [optional](https://github.com/XRPLF/rippled/blob/93ac1aa7aa23f801b0840a87804a6d755c944623/include/xrpl/protocol/detail/transactions.macro#L1026) as we should be able to create a transaction, sign it (leaving `CounterpartySignature` empty) and hand it over to counterparty to sign it and fill in `CounterpartySignature`.

sappenin: Ah, that makes sense. Mind adding something to the Javadoc around that?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
Patel-Raj11: Added [9a3f67f](https://github.com/XRPLF/xrpl4j/pull/719/commits/9a3f67fdd8a24562f0acae17873b8813bc4fccbc)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/LoanSet.java
Patel-Raj11: Added [9a3f67f](https://github.com/XRPLF/xrpl4j/pull/719/commits/9a3f67fdd8a24562f0acae17873b8813bc4fccbc)

sappenin (approved):

cybele-ripple (approved):

PR #769, opened 2026-04-27
Add  Amount/TokenAmount

sappenin:
The Single Asset Vault (SAV) PR requires the concept of a tokenized `Amount` that does not have any other token-identifying information in it (i.e., just a scalar). Because the scalar amounts for the three token types (IOU, MPT, XRP) are slightly different, we re-purpose `CurrencyAmount` into a similar heirarchy of `Amount` (provides a numeric amount only) and `TokenAmount` (provides currency/id information). `TokenAmount` contains an `Amount`.


sappenin:
Closing this in favor of a smaller change via #774

PR #771, opened 2026-04-27
Refactor Issue to a polymorphic interface (XRP / IOU / MPT)

Patel-Raj11:
Extracts out `Issue` refactor from #713 

 ## Summary
  - Refactors `Issue` from a single `@Value.Immutable` (currency + optional issuer) into                                                                                        
    a polymorphic interface with three concrete subtypes: `XrpIssue`, `IouIssue`, `MptIssue`.                                                                                   
  - Adds `Issue.handle(...)` and `Issue.map(...)` for type-safe dispatch over the three                                                                                         
    asset shapes.                                                                                                                                                               
  - Extends the binary codec (`codec/binary/types/Issue` + `IssueType`) to round-trip MPT                                                                                       
    issues using the rippled wire format `issuer(20B) || ACCOUNT_ONE(20B) || sequence(4B LE)`.                                                                                  
  - Adds `UnsignedByteArray.reverse()` for the BE↔LE sequence conversion required by MPT                                                                                        
    issue encoding.                                                                                                                                                             
  - Adds `IssueDeserializer`, which selects the correct subtype based on JSON fields                                                                                            
    (`currency` / `currency`+`issuer` / `mpt_issuance_id`) — JSON wire format is unchanged.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/XrpIssue.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/XrpIssue.java
@@ -1,0 +1,50 @@
+package org.xrpl.xrpl4j.model.ledger;
+
+/*-
+ * ========================LICENSE_START=================================
+ * xrpl4j :: model
+ * %%
+ * Copyright (C) 2020 - 2022 XRPL Foundation and its contributors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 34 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/ledger/XrpIssue.java
```diff
@@ -0,0 +1,50 @@
+package org.xrpl.xrpl4j.model.ledger;
+
+/*-
+ * ========================LICENSE_START=================================
+ * xrpl4j :: model
+ * %%
+ * Copyright (C) 2020 - 2022 XRPL Foundation and its contributors
```
sappenin: Let's bump this to `2026` in each file that's new

Patel-Raj11: Addressed [aa1f329](https://github.com/XRPLF/xrpl4j/pull/771/commits/aa1f32954d07e97ca9cd97093d18ec90d2915692)

sappenin (approved):

PR #783, opened 2026-05-05
Improve error handling in SerializedType.getTypeByName for unknown wire types

### LINKED ISSUE #775: SerializedType.getTypeByName throws raw  NullPointerException on unknown wire types instead of meaningful error
**Current behavior**

When SerializedType.typeMap does not contain an entry for a wire type (e.g.,  "Number"  or  "Int32" ),  getTypeByName  dereferences the null  Supplier  and throws a raw NullPointerException :

```
public static SerializedType<?> getTypeByName(String name) {
    try {
        return typeMap.get(name).get();  // NPE when type is unknown
    } catch (NullPointerException e) {
        throw e;
    }
}
```

This makes debugging difficult because the stack trace does not indicate which  wire type was missing.

**Expected behavior**

Fail fast with a clear, actionable error message that identifies the unknown type and suggests the SDK may need updating.

**Suggested change**

```
public static SerializedType<?> getTypeByName(String name) {
    Supplier<SerializedType<?>> factory = typeMap.get(name);
    if (factory == null) {
        throw new IllegalArgumentException(
            "Unknown serialized type '" + name + "'; xrpl4j may be out of date with rippled's sfields.macro");
    }
    return factory.get();
}
```

**Why this matters**

As the XRPL protocol evolves and new wire types are introduced (e.g.,  NUMBER ,  INT32  for upcoming amendments), integrators currently hit opaque NPEs at the codec boundary. A meaningful exception would immediately surface the root cause and reduce debugging time.

Patel-Raj11:
### Summary

Fixes #775 

Replaced the no-op `catch (NullPointerException e) { throw e; }` in SerializedType.getTypeByName() with explicit validation that throws a descriptive `IllegalArgumentException` when an unknown serialized type is encountered.

### Test plan

Added unit tests to verify known types resolves correctly, unknown types throw a descriptive  `IllegalArgumentException`.
All other existing tests should pass.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/SerializedType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/SerializedType.java
@@ -76,12 +76,16 @@ public abstract class SerializedType<T extends SerializedType<T>> {
    *
    * @return A {@link SerializedType} for the supplied {@code name}.
    */
-  public static SerializedType<?> getTypeByName(String name) {
-    try {
-      return typeMap.get(name).get();
-    } catch (NullPointerException e) {
-      throw e;
+  public static SerializedType<?> getTypeByName(final String name) {
+    Objects.requireNonNull(name);
+
+    if (!typeMap.containsKey(name)) {
+      throw new IllegalArgumentException(
+        String.format("Unknown serialized type '%s'. This likely means xrpl4j is out of date and does not yet " +
+          "support a new field type introduced by a rippled amendment.", name)
+      );
     }
+    return typeMap.get(name).get();
   }
 
   /**
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/SerializedType.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/SerializedType.java
@@ -31,6 +31,7 @@ import org.xrpl.xrpl4j.codec.binary.serdes.BinaryParser;
 
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/codec/binary/types/SerializedType.java
```diff
@@ -76,12 +76,16 @@ public SerializedType(UnsignedByteArray bytes) {
    *
    * @return A {@link SerializedType} for the supplied {@code name}.
    */
-  public static SerializedType<?> getTypeByName(String name) {
-    try {
-      return typeMap.get(name).get();
-    } catch (NullPointerException e) {
-      throw e;
+  public static SerializedType<?> getTypeByName(final String name) {
```
sappenin, line 80: Let's prefer a stream-based syntax, like this:

```java
return Optional.ofNullable(typeMap.get(name))
  .map(Supplier::get)
  .orElseThrow(() -> new IllegalArgumentException(
    String.format("Unknown serialized type '%s'. This likely means xrpl4j is out of date and does not yet " +
      "support a new field type introduced by a rippled amendment.", name)
  ));
```

This replaces the `containsKey`/`get` pair with `Optional.ofNullable` + `orElseThrow`, avoiding the double map lookup.

Patel-Raj11: Makes sense. Addressed in [2b3f6cc](https://github.com/XRPLF/xrpl4j/pull/783/commits/2b3f6ccbe5d3101a888ef7cf2d8df029439e8f51)

sappenin (approved):

Issue #749, opened 2026-04-17
JsonRpcClient.send() NPE when response has no result field


ckeshava:
## Description

`response.get(RESULT)` can return null if the server response doesn't contain a `"result"` key. The null is then passed to `result.toString()` on line 164, causing NPE. The `checkForError()` call on line 162 does check `response.has(RESULT)`, but the error check logic doesn't prevent the subsequent null dereference.

## Location

[JsonRpcClient.java:161-164](https://github.com/XRPLF/xrpl4j/blob/e48a6a6ae2457b0e330390610effe8e881894c4b/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/JsonRpcClient.java#L161-L164)

```java
JsonNode result = response.get(RESULT);  // Can be null
checkForError(response);
return objectMapper.readValue(result.toString(), resultType);  // NPE if result is null
```

## Impact / Severity

**CRITICAL** — Any server response missing `"result"` crashes the client with an unhelpful NullPointerException.

## Adversarial Review Notes

Confirmed by code inspection. The `checkForError()` method checks for `response.has(RESULT)` but only throws if the result has `status == "error"`. A response with neither `result` nor an error status will pass through `checkForError()` and hit the NPE.

## Unit Test Patch

No direct unit test patch — this bug is in the `xrpl4j-client` module which requires network mocking. The bug is confirmed by code inspection of the `send()` default method.

sappenin:
Per the comment by @Patel-Raj11 in #749, this issue is not `Critical` because `xprld` RPC contract guarantees a `result` will exist if the response is not an error (See [documentation](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/response-formatting)). Bottom line, this condition will never occur.

However, in the interest of good Java hygiene and in keeping with the xrpl4j rule that `Null is a Bug`, we should tighten up the contract in the code.

PR #785, opened 2026-05-11
NPE no result issue (issue 749)

cybele-ripple:


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/JsonRpcClient.java
```diff
@@ -160,6 +160,10 @@ default <T extends XrplResult> T send(
     JsonNode response = postRpcRequest(request);
     JsonNode result = response.get(RESULT);
     checkForError(response);
+    if (result == null || result.isNull()) {
```
Patel-Raj11: As per the [documentation](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/response-formatting) result is a required field in the response object. IMO, this condition will never occur where you receive a response from rippled with 200 OK and result field missing. With this reasoning in mind we can avoid this check entirely and close #749

sappenin: I have low conviction around keeping this change (kind of lean towards keeping it, but documenting what @Patel-Raj11 found above to ensure no AI-bot classifies this missing check as `critical`, which it absolutely is not. 

If we keep it, it seems like the throw should just be in `checkForError` instead?

We might even want to create a small helper function called `getResultSafe(response)` which calls `checkForError` inside so that we can always guarantee that `result` is non-null.

sappenin (changes requested):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/JsonRpcClient.java
cybele-ripple: Went with the getResultSafe helper approach [in this commit](https://github.com/XRPLF/xrpl4j/pull/785/changes/601b8f54e1281e17b325255c0bb08b5c1dcb2349)

sappenin (dismissed):

Patel-Raj11 (dismissed):

xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/JsonRpcClient.java
```diff
@@ -158,15 +158,34 @@ default <T extends XrplResult> T send(
...
+   *
+   * @throws JsonRpcClientErrorException If rippled returns an error message or if the response does not contain a
+   *                                     {@code result} field.
+   */
+  default JsonNode getResultSafe(JsonNode response) throws JsonRpcClientErrorException {
+    checkForError(response);
+    JsonNode result = response.get(RESULT);
+    if (result == null || result.isNull()) {
+      throw new JsonRpcClientErrorException(
+        "Response did not contain a 'result' field: " + response);
```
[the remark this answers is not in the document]
cybele-ripple: The exception message now lists only the response's top-level field names via Iterators.toString(response.fieldNames()) instead of embedding the full response, so no values are leaked into logs.

Patel-Raj11 (approved):

sappenin (approved):

Issue #752, opened 2026-04-17
CurrencyAmountDeserializer NPE on missing required fields


ckeshava:
## Description

The deserializer accesses `node.get("value")`, `node.get("currency")`, `node.get("issuer")`, and `node.get("mpt_issuance_id")` without null checks. If any required field is missing from the JSON, `.asText()` on null throws NPE.

## Location

[CurrencyAmountDeserializer.java:57-66](https://github.com/XRPLF/xrpl4j/blob/e48a6a6ae2457b0e330390610effe8e881894c4b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java#L57-L66)

```java
String currency = node.get("currency").asText();  // NPE if "currency" missing
String value = node.get("value").asText();        // NPE if "value" missing
String issuer = node.get("issuer").asText();      // NPE if "issuer" missing
```

## Impact / Severity

**CRITICAL** — Malformed server responses crash the client. DoS vector.

## Adversarial Review Notes

Confirmed by code inspection and unit test. While the `node.has("mpt_issuance_id")` check guards the MPT branch, inside it `node.get("value")` is still unguarded. The IOU branch has no guards at all.

## Unit Test Patch

```diff
+  @Test
+  void bug09_currencyAmountDeserializerNpeOnMissingFields() {
+    String json = "{\"some_field\":\"test\"}";
+    // BUG: Throws NullPointerException instead of a meaningful deserialization error
+    assertThatThrownBy(() -> objectMapper.readValue(json, CurrencyAmount.class))
+      .isInstanceOf(MismatchedInputException.class);
+  }
```

PR #786, opened 2026-05-12
CurrencyAmountDeserializer NPE (Issue 752)

cybele-ripple:
- Replaced node.get() with node.path() in `CurrencyAmountDeserializer` so that missing JSON fields return a MissingNode instead of Java `null`, preventing NPEs
- Added tests to reflect this change

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -56,6 +56,9 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
       if (node.has("mpt_issuance_id")) {
         String mptIssuanceId = node.path("mpt_issuance_id").asText();
         String value = node.path("value").asText();
+        if (mptIssuanceId.isEmpty() || value.isEmpty()) {
+          throw new IOException("Missing required field(s) in MptCurrencyAmount: mpt_issuance_id and value are required.");
+        }
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
@@ -64,6 +67,9 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
         String currency = node.path("currency").asText();
         String value = node.path("value").asText();
         String issuer = node.path("issuer").asText();
+        if (currency.isEmpty() || value.isEmpty() || issuer.isEmpty()) {
+          throw new IOException("Missing required field(s) in IssuedCurrencyAmount: currency, value, and issuer are all required.");
+        }
 
         return IssuedCurrencyAmount.builder()
           .value(value)
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -56,9 +56,6 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
       if (node.has("mpt_issuance_id")) {
         String mptIssuanceId = node.path("mpt_issuance_id").asText();
         String value = node.path("value").asText();
-        if (mptIssuanceId.isEmpty() || value.isEmpty()) {
-          throw new IOException("MptCurrencyAmount missing required field(s): mpt_issuance_id, value");
-        }
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
@@ -67,9 +64,6 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
         String currency = node.path("currency").asText();
         String value = node.path("value").asText();
         String issuer = node.path("issuer").asText();
-        if (currency.isEmpty() || value.isEmpty() || issuer.isEmpty()) {
-          throw new IOException("IssuedCurrencyAmount missing required field(s): currency, value, issuer");
-        }
 
         return IssuedCurrencyAmount.builder()
           .value(value)
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -57,7 +57,7 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
         String mptIssuanceId = node.path("mpt_issuance_id").asText();
         String value = node.path("value").asText();
         if (mptIssuanceId.isEmpty() || value.isEmpty()) {
-          throw new IOException("Missing required field(s) in MptCurrencyAmount: mpt_issuance_id and value are required.");
+          throw new IOException("MptCurrencyAmount missing required field(s): mpt_issuance_id, value");
         }
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
@@ -68,7 +68,7 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
         String value = node.path("value").asText();
         String issuer = node.path("issuer").asText();
         if (currency.isEmpty() || value.isEmpty() || issuer.isEmpty()) {
-          throw new IOException("Missing required field(s) in IssuedCurrencyAmount: currency, value, and issuer are all required.");
+          throw new IOException("IssuedCurrencyAmount missing required field(s): currency, value, issuer");
         }
 
         return IssuedCurrencyAmount.builder()
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -54,16 +54,16 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
 
     if (node.isContainerNode()) {
       if (node.has("mpt_issuance_id")) {
-        String mptIssuanceId = textOrEmpty(node.path("mpt_issuance_id"));
-        String value = textOrEmpty(node.path("value"));
+        String mptIssuanceId = node.get("mpt_issuance_id").asText();
+        String value = node.get("value").asText();
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
           .build();
       } else {
-        String currency = textOrEmpty(node.path("currency"));
-        String value = textOrEmpty(node.path("value"));
-        String issuer = textOrEmpty(node.path("issuer"));
+        String currency = node.get("currency").asText();
+        String value = node.get("value").asText();
+        String issuer = node.get("issuer").asText();
 
         return IssuedCurrencyAmount.builder()
           .value(value)
@@ -74,9 +74,5 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
     } else {
       return XrpCurrencyAmount.ofDrops(node.asLong());
     }
-  }
-
-  private static String textOrEmpty(JsonNode node) {
-    return node.isNull() ? "" : node.asText();
   }
 }
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -54,16 +54,16 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
 
     if (node.isContainerNode()) {
       if (node.has("mpt_issuance_id")) {
-        String mptIssuanceId = node.get("mpt_issuance_id").asText();
-        String value = node.get("value").asText();
+        String mptIssuanceId = node.path("mpt_issuance_id").asText();
+        String value = node.path("value").asText();
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
           .build();
       } else {
-        String currency = node.get("currency").asText();
-        String value = node.get("value").asText();
-        String issuer = node.get("issuer").asText();
+        String currency = node.path("currency").asText();
+        String value = node.path("value").asText();
+        String issuer = node.path("issuer").asText();
 
         return IssuedCurrencyAmount.builder()
           .value(value)
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
@@ -54,16 +54,16 @@ public class CurrencyAmountDeserializer extends StdDeserializer<CurrencyAmount> 
 
     if (node.isContainerNode()) {
       if (node.has("mpt_issuance_id")) {
-        String mptIssuanceId = node.path("mpt_issuance_id").asText();
-        String value = node.path("value").asText();
+        String mptIssuanceId = textOrEmpty(node.path("mpt_issuance_id"));
+        String value = textOrEmpty(node.path("value"));
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
           .build();
       } else {
-        String currency = node.path("currency").asText();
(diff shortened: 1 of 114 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,16 @@ public CurrencyAmount deserialize(
...
+        String value = node.path("value").asText();
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
           .build();
       } else {
-        String currency = node.get("currency").asText();
-        String value = node.get("value").asText();
-        String issuer = node.get("issuer").asText();
+        String currency = node.path("currency").asText();
```
Patel-Raj11: The JSON hitting this deserializer comes from rippled's JSON-RPC responses, which always return well-formed CurrencyAmount objects. The three valid shapes are:               
                                                                                                                                                                                
  - XRP: "1000000" (a string/number, not a container)                                                                                                                 
  - IOU: {"currency":"USD","value":"100","issuer":"r..."} (always has all 3 fields)                                                                                             
  - MPT: {"mpt_issuance_id":"...","value":"100"} (always has both fields)                                                                                                       
                                                                                                                                                                                
Rippled's serialization guarantees these fields are present. The only way to trigger the NPE is by manually feeding malformed JSON into the deserializer, which isn't a realistic scenario — this is an internal deserialization boundary, not a user-facing API that accepts arbitrary input.

@cybele-ripple If the above reasoning makes sense then, can we close #752  as not a bug? WDYT?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,16 @@ public CurrencyAmount deserialize(
...
+        String value = node.path("value").asText();
```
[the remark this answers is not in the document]
cybele-ripple: Added suggested change in [this commit](https://github.com/XRPLF/xrpl4j/pull/786/changes/5b5c1b9c426032e5a4155877ec0ab759b5dfc6a4)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,16 @@ public CurrencyAmount deserialize(
 
     if (node.isContainerNode()) {
       if (node.has("mpt_issuance_id")) {
-        String mptIssuanceId = node.get("mpt_issuance_id").asText();
-        String value = node.get("value").asText();
+        String mptIssuanceId = node.path("mpt_issuance_id").asText();
+        String value = node.path("value").asText();
```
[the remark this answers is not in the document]
cybele-ripple: Added suggested change in [this commit](https://github.com/XRPLF/xrpl4j/pull/786/changes/5b5c1b9c426032e5a4155877ec0ab759b5dfc6a4)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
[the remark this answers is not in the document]
cybele-ripple: Added suggested change in [this commit](https://github.com/XRPLF/xrpl4j/pull/786/changes/5b5c1b9c426032e5a4155877ec0ab759b5dfc6a4)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,22 @@ public CurrencyAmount deserialize(
 
     if (node.isContainerNode()) {
       if (node.has("mpt_issuance_id")) {
-        String mptIssuanceId = node.get("mpt_issuance_id").asText();
-        String value = node.get("value").asText();
+        String mptIssuanceId = node.path("mpt_issuance_id").asText();
```
sappenin: When I execute the unit test in this PR against the change-set (and against the original code), the `get` vs `path` changes appear to have no effect. 

The only failure I see with the old code (and new unit tests) is the check on line 59.

See my code comment in the unit test suggestion. This is because Jackson is handling NPE and mapping it to a `JsonMappingException`.

Thinking more here, I wonder if this isn't "just fine" behavior in the deserialized.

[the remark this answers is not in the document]
sappenin: Can we confirm this in unit tests?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,22 @@ public CurrencyAmount deserialize(
...
+        String value = node.path("value").asText();
+        if (mptIssuanceId.isEmpty() || value.isEmpty()) {
```
sappenin: If we keep these changes, we should update this section to the following:

```java
        // The idiomatic Jackson approach for a deserializer is to use MismatchedInputException (what Jackson itself
        // throws for missing required fields). The advantage is that MismatchedInputException carries JSON location
        // info (line/column), making error messages more actionable.
        if (mptIssuanceId.isEmpty()) {
          throw MismatchedInputException.from(
            jsonParser, MptCurrencyAmount.class, "MptCurrencyAmount missing required field(s): mpt_issuance_id"
          );
        }

        if (value.isEmpty()) {
          throw MismatchedInputException.from(
            jsonParser, MptCurrencyAmount.class, "MptCurrencyAmount missing required field(s): value"
          );
        }
```

Also, we should separate the two value checks. The current message includes both fields when only one is missing, which is ambiguous at best, and wrong at worst.

sappenin: That said, In this project we prefer to put content/boundary checks in the Immutables layer via @Value.Check rather than in deserializers, so that the invariants are enforced everywhere an object is constructed and not just through the Jackson path.

  Two suggestions:

  1. Move the value emptiness check into Wrappers._MpTokenIssuanceId:
     ```java
     @Value.Check
     void check() {
      Preconditions.checkArgument(this.value().length() == 24, "MpTokenIssuanceId must be 24 characters (12 bytes) long.");
      }
     ```  
  2. Add an additional check in `MptCurrencyAmount` to check that `value` is not empty.

  Once those checks are in place, the deserializer can drop both if-blocks entirely and let construction throw naturally. This also means the validation applies to any code that builds an MptCurrencyAmount directly, not just deserialized JSON.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/CurrencyAmountTest.java
```diff
@@ -438,6 +439,80 @@ void toAmountRoundtripsBackToMptCurrencyAmount() {
...
+    // Prior to fix, node.get("value") returned Java null, causing NPE on .asText().
+    // After fix, missing required fields throw IOException (wrapped as JsonProcessingException).
+    String json = "{\"amount\": {\"mpt_issuance_id\": \"00000143A58DCB491FD36A15A7D3172E6A9F088A5478BA41\"}}";
+    assertThatThrownBy(() -> objectMapper.readValue(json, CurrencyAmountWrapper.class))
+      .isInstanceOf(JsonProcessingException.class)
+      .isNotInstanceOf(NullPointerException.class);
+  }
+
+  @Test
+  void deserializeCurrencyAmountWithEmptyObjectDoesNotThrowNpe() {
```
sappenin: I think we need to refactor these tests a bit. They don't fail even with the old code, but that's because Jackson is mapping the NPE to something else. This is a better way to structure this test:

```java
@Test
  void deserializeMptCurrencyAmountWithMissingValueDoesNotThrowNpe() {
    // node.get("value") returns Java null when the field is absent; .asText() on null throws NPE.
    // Jackson wraps that NPE in a generic JsonMappingException, not MismatchedInputException.
    // Asserting MismatchedInputException here ensures the deserializer uses node.path() (or an
    // explicit guard) so the isEmpty() check fires and throws the right exception type.
    String json = "{\"amount\": {\"mpt_issuance_id\": \"00000143A58DCB491FD36A15A7D3172E6A9F088A5478BA41\"}}";
    assertThatThrownBy(() -> objectMapper.readValue(json, CurrencyAmountWrapper.class))
      .isInstanceOf(MismatchedInputException.class);
  }
```

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
cybele-ripple: Validation is now in `@Value.Check` on `_MpTokenIssuanceId`, `MptCurrencyAmount`, and `IssuedCurrencyAmount` rather than in the deserializer. This means the invariants are enforced at every construction site, not just through the Jackson path. Changes in this PR are related to the issue so I still think it's worth it to keep the issue and this PR paired.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
cybele-ripple: Yes jackson maps the NPE to JsonMappingException either way, so `get` vs `path` made no difference through the Jackson path. 

The latest commit has `node.path()` to ensure absent fields produce an empty string `""` (rather than relying on NPE-wrapping).  `@Value.Check` on the model types now have a clean `IllegalArgumentException`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
cybele-ripple: Went with the @Value.Check so that invariants are enforced everywhere

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/model/transactions/CurrencyAmountTest.java
cybele-ripple: Refactored test to assert `MismatchedInputException` instead of `JsonProcessingException`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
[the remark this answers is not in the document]
cybele-ripple: Now confirmed in [this test](https://github.com/XRPLF/xrpl4j/pull/786/changes#diff-d67ae5d6b6411ad3fda6e0e8cbe55acdc16e7a1cda5d6a0af23c7eab4d99a71bR491-R499) in CurrencyAmountTest

sappenin: > The latest commit has node.path() to ensure absent fields produce an empty string "" (rather than relying on NPE-wrapping). 

Hmmm... I think if Java expects a field to be populated (per the Java contract, no `Optional`) and the JSON is missing a field, then an `NPE` (mapped to `JsonMappingException`) is the correct exception. In this library, `Null is a bug`, so throwing `IllegalArgumentException` feels like it subtly shifts the expectations of the library. 

All that to say, I think we should keep using `.get`, but am I missing something obvious?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
```diff
@@ -54,16 +54,16 @@ public CurrencyAmount deserialize(
...
         return MptCurrencyAmount.builder()
           .mptIssuanceId(MpTokenIssuanceId.of(mptIssuanceId))
           .value(value)
           .build();
       } else {
-        String currency = node.get("currency").asText();
-        String value = node.get("value").asText();
-        String issuer = node.get("issuer").asText();
+        String currency = textOrEmpty(node.path("currency"));
+        String value = textOrEmpty(node.path("value"));
```
sappenin: I'm not sure I agree with this change (open to changing my mind), and I realize this sort of contradicts my previous PR feedback. Apologies for the extra work, but I'm seeing things more clearly now that you've implemented things in this way.

So, if it happens that something like the `currency` node is not present, then it will be `null`. An `NPE` will be thrown, mapped to a `JsonMappingException`. 

That feels correct to me. The problem is actually a `JsonMappingException` -- something was wrong with the JSON, and our Java contract explicitly enforces that `currency` can never be null (if it is null, then something went wrong).

This PR subtly shifts the meaning of this scenario. Instead of the NPE above, `currency` get mapped to an empty string, and is then caught in the `@Check` method of `IssuedCurrencyAmount`. However, this will manifest as an `IllegalArgumentException`, which isn't actually helpful because it masks the actual problem (which is: bad JSON!). 

I think instead we want to see the JSON exception so that developers know where to look.

All that reduces to: I think we should use `.get()`.

cybele-ripple: No worries! Reverted to `node.get()`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
Patel-Raj11: I think this makes sense. Should we add not-empty conditions on `value` field as well for both `IssuedCurrencyAmount` and `MptCurrencyAmount`?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
```diff
@@ -97,6 +98,11 @@ static ImmutableIssuedCurrencyAmount.Builder builder() {
    */
   Address issuer();
 
+  @Value.Check
+  default void check() {
+    Preconditions.checkArgument(!currency().isEmpty(), "IssuedCurrencyAmount currency must not be empty.");
```
sappenin, line 103: This check feels reasonable, but given that we're typing the `value` as a `String`, I think we should probably remove this check. 

That might sound counterintuitive, or even nonsensical. However, rippled always surprises me with the possible values it returns. It frequently does things it's not supposed to, and I've had enough xrpl4j hot-fix releases that relax a precondition, that I'm now quite hesitant to aggressively check precondtions.

My approx rule of thumb is

cybele-ripple: It seems safer to keep the value check to defend against empty string currency values. Removing it could cause confusing downstream debugging if someone passes in an empty string. Planning to keep in the check unless it would have an effect on the code.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
cybele-ripple: I think so. Added not-empty conditions [for IssuedCurrencyAmount.java](https://github.com/XRPLF/xrpl4j/pull/786/changes#diff-41a3ac0475efe4f2a3e3d168bc6f01630532105a818641e3fbaa60da97d98b83R103-R104) and [MptCurrencyAmount.java](https://github.com/XRPLF/xrpl4j/pull/786/changes#diff-1fa30a8f875e9e287a08fe38fe3c7e838e881c47d63250d33b5a365a5d2bd4f1R47-R51). I also added associated unit tests in [IssuedCurrencyAmountTest](https://github.com/XRPLF/xrpl4j/pull/786/changes#diff-0b33833514781db92536f5b8496b8d264dc878ca52e14ec3ed489f37830b20c5R134-R144) and [MptCurrencyAmountTest](https://github.com/XRPLF/xrpl4j/pull/786/changes#diff-450ebfe3737ab9c41ff40c34ba043c6edbb02f68b60058d48bd53f3ac55adf9aR118-R128)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java
cybele-ripple: I agree and the [current version of CurrencyAmountDeserializer on main uses node.get()](https://github.com/XRPLF/xrpl4j/blob/main/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/CurrencyAmountDeserializer.java#L57-L66)

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/IssuedCurrencyAmount.java
sappenin: Ok, as long as you make the fix if it breaks, I'm supportive. 😇

sappenin (approved):

PR #787, opened 2026-05-12
Fixing integration test flakiness

cybele-ripple:
These test changes improve the reliability of integration tests against clio clusters. Previously, tests would confirm a transaction reached a validated ledger state and then immediately read derived state from a clio node. However, clio nodes replicate ledger data independently so a given node's cache can lag behind consensus. Reads made before replication completes return stale data, causing intermittent failures.

With these changes, a read is only made when the account sequence number increases meaning that the current clio node's replication state has caught up with the with the ledger.

### DIFF
--- a/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
+++ b/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
@@ -802,10 +802,12 @@ public abstract class AbstractIT {
 
       assertThat(createTxIntermediateResult.engineResult()).isEqualTo("tesSUCCESS");
 
-      // Wait for the tx to be applied. A plain scanForResult on getValidatedTransaction is not sufficient
-      // here because when this code executes against a Clio endpoint operating in a cluster, each endpoint
-      // may return stale data even after reporting the tx as validated. The predicate retries the Clio call
-      // for up to 30s until at least one node returns the tx with tesSUCCESS in its metadata.
+      // Wait for the account sequence to advance, which confirms the tx was applied. A plain `scanForResult` on
+      // `getValidatedTransaction` is not sufficient here because when this code executes against a Clio endpoint
+      //  operating in a cluster, each endpoint may return stale `account_info` with the old sequence even after
+      // reporting the `tx` as validated. This solution works because the predicate supplied to
+      // `getValidatedAccountInfo` means that function will keep retrying the Clio call for 30s until it
+      // encounters at least one node with the update value.
       this.scanForResult(
         () -> this.getValidatedTransaction(
           createTxIntermediateResult.transactionResult().hash(), CredentialCreate.class),
--- a/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
+++ b/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
@@ -802,10 +802,14 @@ public abstract class AbstractIT {
 
       assertThat(createTxIntermediateResult.engineResult()).isEqualTo("tesSUCCESS");
 
-      // Then wait until the transaction gets committed to a validated ledger
+      // Wait for the account sequence to advance, which confirms the tx was applied on this Clio node.
+      // A plain scanForResult on getValidatedTransaction is not sufficient because Clio nodes in a cluster
+      // may return stale account_info with the old sequence even after reporting the tx as validated.
+      final UnsignedInteger expectedIssuerSequence = issuerAccountInfo.accountData().sequence()
+        .plus(UnsignedInteger.ONE);
       this.scanForResult(
-        () ->
-          this.getValidatedTransaction(createTxIntermediateResult.transactionResult().hash(), CredentialCreate.class)
+        () -> this.getValidatedAccountInfo(issuerKeyPair.publicKey().deriveAddress()),
+        info -> info.accountData().sequence().equals(expectedIssuerSequence)
       );
     }
   }
@@ -880,10 +884,12 @@ public abstract class AbstractIT {
 
       assertThat(acceptTxIntermediateResult.engineResult()).isEqualTo("tesSUCCESS");
 
-      // Then wait until the transaction gets committed to a validated ledger
+      // Wait for the account sequence to advance, which confirms the tx was applied on this Clio node.
+      final UnsignedInteger expectedSubjectSequence = subjectAccountInfo.accountData().sequence()
+        .plus(UnsignedInteger.ONE);
       this.scanForResult(
-        () ->
-          this.getValidatedTransaction(acceptTxIntermediateResult.transactionResult().hash(), CredentialAccept.class)
+        () -> this.getValidatedAccountInfo(subjectKeyPair.publicKey().deriveAddress()),
+        info -> info.accountData().sequence().equals(expectedSubjectSequence)
       );
     }
   }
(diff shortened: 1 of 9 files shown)


kuan121 (dismissed):

ckeshava (dismissed):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -802,10 +802,14 @@ protected void createCredentials(KeyPair issuerKeyPair, KeyPair subjectKeyPair,
 
       assertThat(createTxIntermediateResult.engineResult()).isEqualTo("tesSUCCESS");
 
-      // Then wait until the transaction gets committed to a validated ledger
+      // Wait for the account sequence to advance, which confirms the tx was applied on this Clio node.
```
sappenin, line 805: This comment isn't quite right because this class (AbstractIT) runs for all GH jobs/environments. Let's reword this subtly to qualify it, something like this:

```
Wait for the account sequence to advance, which confirms the tx was applied. A plain `scanForResult` on 
`getValidatedTransaction` is not sufficient here because when this code executes against a Clio endpoint
 operating in a cluster, each endpoint may return stale `account_info` with the old sequence even after 
reporting the `tx` as validated. This solution works because the predicate supplied to `getValidatedAccountInfo` means that function will keep retrying the Clio call for 30s until it encounters at least one node with the update value.
```

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -802,10 +802,14 @@ protected void createCredentials(KeyPair issuerKeyPair, KeyPair subjectKeyPair,
...
+      // A plain scanForResult on getValidatedTransaction is not sufficient because Clio nodes in a cluster
+      // may return stale account_info with the old sequence even after reporting the tx as validated.
+      final UnsignedInteger expectedIssuerSequence = issuerAccountInfo.accountData().sequence()
+        .plus(UnsignedInteger.ONE);
       this.scanForResult(
-        () ->
-          this.getValidatedTransaction(createTxIntermediateResult.transactionResult().hash(), CredentialCreate.class)
+        () -> this.getValidatedAccountInfo(issuerKeyPair.publicKey().deriveAddress()),
+        info -> info.accountData().sequence().equals(expectedIssuerSequence)
```
sappenin, line 812: See my comment above for how I'm currently imagining the overall solution in this PR is working. Assuming my understanding is correct, I'm wondering if there's a simpler solution for this PR (one that doesn't involve tracking an `expectedIssuerSequence`). 

Basically, by simply adding a predicate to the existing code, the old check will eventually succeed. Here's the simpler fix (confirmed locally hitting Clio, though this might not be the same as hitting Clio from CI):

```
this.scanForResult(
  () -> this.getValidatedTransaction(createTxIntermediateResult.transactionResult().hash(), CredentialCreate.class),
  result -> result.metadata().map(meta -> meta.transactionResult().equals("tesSUCCESS")).orElse(false)
);
```

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
```diff
@@ -880,10 +884,12 @@ protected void acceptCredentials(KeyPair issuerKeyPair, KeyPair subjectKeyPair,
...
       assertThat(acceptTxIntermediateResult.engineResult()).isEqualTo("tesSUCCESS");
 
-      // Then wait until the transaction gets committed to a validated ledger
+      // Wait for the account sequence to advance, which confirms the tx was applied on this Clio node.
+      final UnsignedInteger expectedSubjectSequence = subjectAccountInfo.accountData().sequence()
+        .plus(UnsignedInteger.ONE);
       this.scanForResult(
-        () ->
-          this.getValidatedTransaction(acceptTxIntermediateResult.transactionResult().hash(), CredentialAccept.class)
+        () -> this.getValidatedAccountInfo(subjectKeyPair.publicKey().deriveAddress()),
```
sappenin, line 893: See my comment above. If we can get rid of `expectedSubjectSequence` I think we should.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/CredentialIT.java
```diff
@@ -147,7 +149,8 @@ public void testCreateAcceptDeleteCredential() throws JsonRpcClientErrorExceptio
     );
 
     subjectAccountInfo = this.scanForResult(
-      () -> this.getValidatedAccountInfo(subjectKeyPair.publicKey().deriveAddress())
+      () -> this.getValidatedAccountInfo(subjectKeyPair.publicKey().deriveAddress()),
```
sappenin: Note: I'm not able to make calls to the Clio server using Java 21 on my local machine. Using Java 26 does seem to work though, so I'm going to make a PR to make Github actions use that. Then I think this PR's Clio action will start working.

See https://github.com/XRPLF/xrpl4j/pull/788

sappenin (changes requested):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
cybele-ripple: Updated to use this exact working

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
cybele-ripple: Added in the predicate [here](https://github.com/XRPLF/xrpl4j/pull/787/changes#diff-ae4185e6a768c34e1d955a6f6736ca9b65aa1582a4bae18e3fc963c887232b43R812-R814)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
cybele-ripple: Applied the same predicate pattern [here](https://github.com/XRPLF/xrpl4j/pull/787/changes#diff-ae4185e6a768c34e1d955a6f6736ca9b65aa1582a4bae18e3fc963c887232b43R890-R892)

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/CredentialIT.java
```diff
@@ -115,13 +116,14 @@ public void testCreateAcceptDeleteCredential() throws JsonRpcClientErrorExceptio
     AccountInfoResult subjectAccountInfo = this.scanForResult(
       () -> this.getValidatedAccountInfo(subjectKeyPair.publicKey().deriveAddress())
     );
+    final UnsignedInteger subjectSeqBeforeAccept = subjectAccountInfo.accountData().sequence();
```
sappenin: @cybele-ripple Just double-checking to see if this tracking number is needed, or if the simpler "just add predicate" works for this one.

sappenin:
FYI: Latest run of this PR is still flaky on Clio.

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/CredentialIT.java
cybele-ripple: Inlined the sequence on the builder and switched to the `tesSUCCESS` predicate pattern

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractIT.java
sappenin: Oh, I think maybe I wasn't clear in my comment. I was proposing that we remove the `expectedIssuerSequence` entirely. See my code suggestion, but if we simply add a predicate, I think the tests will do what you're wanting (and we don't need to add a new tracking sentinel everywhere).

sappenin (approved):

Issue #750, opened 2026-04-17
AffectedNodeDeserializer NPE on missing LedgerEntryType


ckeshava:
## Description

Accesses `nodeFieldAndValue.getValue().get("LedgerEntryType").asText()` without null checks. If `LedgerEntryType` is missing from the node value, `.get()` returns null and `.asText()` throws NPE.

## Location

[AffectedNodeDeserializer.java:59-62](https://github.com/XRPLF/xrpl4j/blob/e48a6a6ae2457b0e330390610effe8e881894c4b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java#L59-L62)

```java
MetaLedgerEntryType ledgerEntryType = MetaLedgerEntryType.of(
  nodeFieldAndValue.getValue().get("LedgerEntryType").asText()
);
```

## Impact / Severity

**CRITICAL** — DoS via malformed transaction metadata. A `CreatedNode` or `ModifiedNode` missing its `LedgerEntryType` crashes the client.

## Adversarial Review Notes

Confirmed. No null check on the result of `.get("LedgerEntryType")`.

## Unit Test Patch

```diff
+  @Test
+  void bug11_affectedNodeDeserializerNpeOnMissingLedgerEntryType() {
+    String json = "{\"CreatedNode\":{\"SomeOtherField\":\"value\"}}";
+    // BUG: Throws NullPointerException instead of a meaningful deserialization error
+    assertThatThrownBy(() -> objectMapper.readValue(json, AffectedNode.class))
+      .isInstanceOf(MismatchedInputException.class);
+  }
```

PR #797, opened 2026-05-27
AffectedNodeDeserializer NPE on missing LedgerEntryType (Issue #750)

cybele-ripple:
Updated exception use case so that `MismatchedInputException` is thrown instead of a `NullPointerException` when `LedgerEntryType` is absent from a `CreatedNode`, `ModifiedNode`, or `DeletedNode` payload. There was previously no associated unit tests for `AffectedNodeDeserializerTest.java` so `AffectedNodeDeserializer.java` was added.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
@@ -27,6 +27,7 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import org.xrpl.xrpl4j.model.transactions.metadata.AffectedNode;
 import org.xrpl.xrpl4j.model.transactions.metadata.CreatedNode;
 import org.xrpl.xrpl4j.model.transactions.metadata.DeletedNode;
@@ -56,9 +57,13 @@ public class AffectedNodeDeserializer extends StdDeserializer<AffectedNode> {
     Map.Entry<String, JsonNode> nodeFieldAndValue = jsonNode.fields().next();
     String affectedNodeType = nodeFieldAndValue.getKey();
 
-    MetaLedgerEntryType ledgerEntryType = MetaLedgerEntryType.of(
-      nodeFieldAndValue.getValue().get("LedgerEntryType").asText()
-    );
+    JsonNode ledgerEntryTypeNode = nodeFieldAndValue.getValue().get("LedgerEntryType");
+    if (ledgerEntryTypeNode == null || ledgerEntryTypeNode.isNull()) {
+      throw MismatchedInputException.from(
+        jsonParser, AffectedNode.class, "AffectedNode is missing required 'LedgerEntryType' field"
+      );
+    }
+    MetaLedgerEntryType ledgerEntryType = MetaLedgerEntryType.of(ledgerEntryTypeNode.asText());
     Class<? extends MetaLedgerObject> ledgerObjectClass = ledgerEntryType.ledgerObjectType();
 
     switch (affectedNodeType) {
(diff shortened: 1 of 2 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
```diff
@@ -56,9 +57,13 @@ public AffectedNode deserialize(JsonParser jsonParser, DeserializationContext ct
     Map.Entry<String, JsonNode> nodeFieldAndValue = jsonNode.fields().next();
```
sappenin, line 57: This PR is heading in the right direction, but it has me thinking of two things:

1. If we keep this PR, we should be more agressive about checking for potentially null values (e.g., `odeFieldAndValue.getValue().get("LedgerEntryType");`)
2. If we do encounter a null value anywhere, it seems better to just log this as a warning and continue/skip the invalid input. PR #807 is my attempt to implement that strategy.

@cybele-ripple @Patel-Raj11 Take a look and both, and let me know which one you prefer. Currently, I think #807 is a better approach, but am open to hearing alternate opinions.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
```diff
@@ -56,9 +57,13 @@ public AffectedNode deserialize(JsonParser jsonParser, DeserializationContext ct
...
     String affectedNodeType = nodeFieldAndValue.getKey();
 
-    MetaLedgerEntryType ledgerEntryType = MetaLedgerEntryType.of(
-      nodeFieldAndValue.getValue().get("LedgerEntryType").asText()
-    );
+    JsonNode ledgerEntryTypeNode = nodeFieldAndValue.getValue().get("LedgerEntryType");
```
sappenin, line 60: We should guard against a null `nodeFieldAndValue.getValue()`

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
sappenin: @cybele-ripple Any thoughts around my comment above?

cybele-ripple: I am concerned about silently dropping a malformed `AffectedNode` if we go with #807

sappenin: That sounds fine; I'm happy to go with this approach.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
cybele-ripple: Fixed [here](https://github.com/XRPLF/xrpl4j/blob/951eeaf929b2779dd03b5754a9a15523c77b0948/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java#L62-L64). `nodeFieldAndValue.getValue()` is now guarded before it's used to look up `LedgerEntryType`, and now throws `MismatchedInputException` when the entry's value is null.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeDeserializer.java
```diff
@@ -55,26 +56,36 @@ public AffectedNode deserialize(JsonParser jsonParser, DeserializationContext ct
...
+        jsonParser, AffectedNode.class, "AffectedNode entry value is missing or null"
+      );
+    }
 
-    MetaLedgerEntryType ledgerEntryType = MetaLedgerEntryType.of(
-      nodeFieldAndValue.getValue().get("LedgerEntryType").asText()
-    );
+    JsonNode ledgerEntryTypeNode = nodeValue.get("LedgerEntryType");
+    if (ledgerEntryTypeNode == null || ledgerEntryTypeNode.isNull()) {
+      throw MismatchedInputException.from(
```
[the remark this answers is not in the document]
sappenin: @cybele-ripple This one sound good to you?

sappenin (dismissed):

PR #798, opened 2026-05-27
Fix Math.abs(Long.MIN_VALUE) overflow in XrpCurrencyAmount.ofDrops(long)

cybele-ripple:
Within the `ofDrops` function, when the value of drops is negative, this value is normalized through `UnsignedLong.valueOf(Math.abs(drops))`. When the value of drops is equal to `Long.MIN_VALUE` (`-9,223,372,036,854,775,808`), then `Math.abs(drops)` should return `9,223,372,036,854,775,808`. However, `Long.MAX_VALUE` is equal to `9,223,372,036,854,775,807` so any value larger than this is outside the range of a Long value. Additionally, this value is outside the range of the total number of drops in existence (`100,000,000,000,000,000`).

Other large values outside of the max amount of drops allowed are caught by `@Value.Check` so we only need to fix this special edge case.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
@@ -210,12 +210,13 @@ public class Wrappers {
      * @return An {@link XrpCurrencyAmount} of {@code drops}.
      */
     public static XrpCurrencyAmount ofDrops(final long drops) {
-      if (drops == Long.MIN_VALUE) {
-        throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Wrappers.java
```diff
@@ -210,6 +210,11 @@ abstract static class _XrpCurrencyAmount extends Wrapper<UnsignedLong> implement
      * @return An {@link XrpCurrencyAmount} of {@code drops}.
      */
     public static XrpCurrencyAmount ofDrops(final long drops) {
+      if (drops == Long.MIN_VALUE) {
```
sappenin: Minor nit: Move this check inside of the `if(drops < 0)` block so that in the happy-path, it's not checked?

sappenin (dismissed):

sappenin (approved):

PR #801, opened 2026-05-28
Fix secp256k1 deriveScalar accumu0lating bytes across iterations (Issu…

cybele-ripple:
`HashingUtils.addUInt32()` appends 4 bytes in-place to its UnsignedByteArray argument. Because `seedCopy` was created once outside the loop, each iteration kept appending its counter onto the bytes from the previous iteration. This PR updates the implementation to match the [rippled implementation](https://github.com/XRPLF/rippled/blob/70d5c624e8cf732a362335642b2f5125ce4b43c1/src/libxrpl/protocol/SecretKey.cpp#L103-L114), which hashes on each iteration using the original seed bytes as the base.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
@@ -475,8 +475,8 @@ public interface Seed extends javax.security.auth.Destroyable {
         Objects.requireNonNull(discriminator);
 
         BigInteger key = null;
-        UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
         for (long i = 0; i <= 0xFFFFFFFFL; i++) {
+          UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
           discriminator.map(d -> HashingUtils.addUInt32(seedCopy, d));
           HashingUtils.addUInt32(seedCopy, (int) i);
           UnsignedByteArray hash = HashingUtils.sha512Half(seedCopy);
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
@@ -476,6 +476,8 @@ public interface Seed extends javax.security.auth.Destroyable {
 
         BigInteger key = null;
         for (long i = 0; i <= 0xFFFFFFFFL; i++) {
+          // Rarely loops (secp256k1 order ≈ 2^256, so iteration is needed with negligible probability),
+          // but seedCopy must be recreated each iteration to avoid accumulating bytes across retries.
           UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
           discriminator.map(d -> HashingUtils.addUInt32(seedCopy, d));
           HashingUtils.addUInt32(seedCopy, (int) i);
--- a/xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/SeedTest.java
+++ b/xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/SeedTest.java
@@ -364,4 +364,27 @@ public class SeedTest {
       "rByLcEZ7iwTBAK8FfjtpFuT7fCzt4kF4r2"
     );
   }
+
+  @Test
+  void deriveScalarDoesNotAccumulateBytesAcrossIterations() {
+    // addUInt32 mutates its UnsignedByteArray argument in-place by appending 4 bytes.
+    // Before the fix, seedCopy was created once outside the loop so every iteration
+    // appended onto the previous iteration's bytes, corrupting the hash input.
+    // This test confirms addUInt32 grows the array and that the fix (re-creating
+    // seedCopy on each iteration) prevents that accumulation.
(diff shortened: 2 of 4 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
```diff
@@ -475,8 +475,8 @@ private static BigInteger deriveScalar(final UnsignedByteArray seed, final Optio
         Objects.requireNonNull(discriminator);
 
         BigInteger key = null;
-        UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
         for (long i = 0; i <= 0xFFFFFFFFL; i++) {
+          UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
```
sappenin, line 481: It was somewhat non-obvious why this change is required (mostly because the for-loop implies that it will potentially do many loops). In reality, this seems to generally only loop once, and only rarely loop more than once. So, how about adding a small disclaimer into the code so that future developers don't make the same mistake I did?

```suggestion
          // Rarely loops (secp256k1 order ≈ 2^256, so iteration is needed with negligible probability),
         // but seedCopy must be recreated each iteration to avoid accumulating bytes across retries.
          UnsignedByteArray seedCopy = UnsignedByteArray.of(seed.toByteArray());
```

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/SeedTest.java
```diff
@@ -364,4 +364,27 @@ public void generateWalletFromSecp256k1Seed() {
       "rByLcEZ7iwTBAK8FfjtpFuT7fCzt4kF4r2"
     );
   }
+
+  @Test
+  void deriveScalarDoesNotAccumulateBytesAcrossIterations() {
```
sappenin, line 377: This test doesn't fail when I revert the changes in `Seed.java` above. Claude is suggesting to me this test instead, tell me what you think (it does fail if I revert the above change):

```java
@Test
  void deriveScalarDoesNotAccumulateBytesAcrossIterations() {
    // Force deriveScalar's loop to iterate twice by making the first sha512Half call return a value
    // >= N (an invalid scalar). The second call's argument reveals whether seedCopy was re-created
    // inside the loop (correct: seed + [i=1] = 20 bytes) or accumulated outside it
    // (broken: seed + [i=0] + [i=1] = 24 bytes).
    //
    // Rarely loops (secp256k1 order ≈ 2^256, so iteration is needed with negligible probability),
    // but seedCopy must be recreated each iteration to avoid accumulating bytes across retries.

    // All-0xFF as a BigInteger is >> N, so the loop rejects it and advances to i=1.
    byte[] aboveN = new byte[32];
    Arrays.fill(aboveN, (byte) 0xFF);
    UnsignedByteArray invalidScalar = UnsignedByteArray.of(aboveN);

    // BigInteger(1) is a valid scalar (> 0 and < N).
    byte[] belowN = new byte[32];
    belowN[31] = 1;
    UnsignedByteArray validScalar = UnsignedByteArray.of(belowN);

    ArgumentCaptor<UnsignedByteArray> captor = ArgumentCaptor.forClass(UnsignedByteArray.class);

    Entropy entropy = Entropy.of(BaseEncoding.base16().decode("CC4E55BC556DD561CBE990E3D4EF7069"));
    Seed ecSeed = Seed.secp256k1SeedFromEntropy(entropy);

    try (MockedStatic<HashingUtils> mocked = mockStatic(HashingUtils.class)) {
      mocked.when(() -> HashingUtils.addUInt32(any(), any())).thenCallRealMethod();
      // Call 1: deriveScalar(seed), i=0 → rejected (>= N), forces loop to i=1
      // Call 2: deriveScalar(seed), i=1 → accepted
      // Call 3: deriveScalar(publicGen, accountIndex=0), i=0 → accepted
      mocked.when(() -> HashingUtils.sha512Half(any(UnsignedByteArray.class)))
        .thenReturn(invalidScalar)
        .thenReturn(validScalar)
        .thenReturn(validScalar);

      ecSeed.deriveKeyPair();

      mocked.verify(() -> HashingUtils.sha512Half(captor.capture()), times(3));
    }

    List<UnsignedByteArray> args = captor.getAllValues();
    // The seed entropy is 16 bytes. With correct code, the second sha512Half call receives
    // seed + [0,0,0,1] = 20 bytes. With broken code it would receive seed + [0,0,0,0] + [0,0,0,1] = 24 bytes.
    assertThat(args.get(1).length()).isEqualTo(16 + 4);
  }
```

For the above test to work, we might need to add the following dependency to the pom:

```xml
  <dependency>
      <groupId>org.mockito</groupId>
      <artifactId>mockito-inline</artifactId>
      <scope>test</scope>
  </dependency>
```

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/crypto/keys/Seed.java
cybele-ripple: Good call. Just added your suggested comment

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/keys/SeedTest.java
cybele-ripple: Added `mockito-inline` and a test that fails before this change and passes after the change

sappenin (approved):

PR #807, opened 2026-06-12
Alternate approach to #797: warn-and-skip instead of throw on null input (Fixed #750)

### LINKED ISSUE #750: AffectedNodeDeserializer NPE on missing LedgerEntryType

sappenin:
The current code and the approach proposed in #797 both throw on invalid input. This PR takes a different stance: log a warning and skip the offending entry rather than failing hard. The rationale: rippled should never produce a null JSON value, but in practice it doesn't always behave the way Java expects — particularly around null inputs.

Changes:
AffectedNodeDeserializer: when LedgerEntryType is absent from a CreatedNode, ModifiedNode, or DeletedNode payload, the deserializer now warns and skips rather than propagating an exception. 

Exception behavior fix: MismatchedInputException is now thrown (instead of NullPointerException) when LedgerEntryType is missing — correcting the existing error type regardless of the null-input path
AffectedNodeDeserializerTest: added unit tests, which were previously absent entirely


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/jackson/modules/AffectedNodeListDeserializer.java
```diff
@@ -0,0 +1,56 @@
...
+    super(List.class);
+  }
+
+  @Override
+  public List<AffectedNode> deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+    List<AffectedNode> result = new ArrayList<>();
+    while (jsonParser.nextToken() != JsonToken.END_ARRAY) {
+      AffectedNode node = elementDeserializer.deserialize(jsonParser, ctxt);
+      if (node != null) {
+        result.add(node);
```
sappenin, line 51: TODO: An else block should warn if `node == null`

sappenin:
Closing; will go with #797

PR #809, opened 2026-06-30
Dynamic mpt 7439 remove clear flags

cybele-ripple:
This PR covers the changes made in [this rippled PR](https://github.com/XRPLF/rippled/pull/7439) including:
-  removed all CLEAR_* flags; remaining SET_* bits compacted to 0x01–0x20 (VALID_MASK = 0x3F).
- renamed the six capability flags CanMutate* → CanEnable* (bit values unchanged)
- dropped the now-impossible validations (set+clear conflict; non-zero TransferFee + ClearCanTransfer).
- updated tests to reflect these changes


xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/MpTokenIT.java
```diff
@@ -699,6 +699,118 @@ void createDynamicIssuanceThenToggleFlagViaMutableFlags()
...
+        .tmfMptCanEnableCanTransfer(true)
+        .build()
+      )
+      .build();
+
+    SingleSignedTransaction<MpTokenIssuanceCreate> signedCreate = signatureService.sign(
+      issuerKeyPair.privateKey(), issuanceCreate
+    );
+    SubmitResult<MpTokenIssuanceCreate> createResult = xrplClient.submit(signedCreate);
+    Assumptions.assumeTrue(
```
Patel-Raj11, line 740: nit: We should remove this assertion, as it would skip the entire test rather then failing it.

Patel-Raj11 (dismissed):

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/MpTokenIssuanceCreateMutableFlags.java
```diff
@@ -5,47 +5,49 @@
  * {@link org.xrpl.xrpl4j.model.transactions.MpTokenIssuanceCreate} transactions.
  *
  * <p>These flags (prefixed with {@code tmf}) declare which fields or flags of the created
- * {@code MPTokenIssuance} may be mutated after issuance via {@code MPTokenIssuanceSet}.
+ * {@code MPTokenIssuance} may be mutated after issuance via {@code MPTokenIssuanceSet}. The
+ * {@code CanEnable*} flags declare that the corresponding capability may later be enabled (once) by the
```
sappenin: Clarifyin intent here - once enabled, they can't be disabled correct? (that's my take-away from the use of the word `once`).

sappenin: FWIW, I sort of like the language below in [MpTokenSetMutableFlags](https://github.com/XRPLF/xrpl4j/pull/809/changes#diff-c504aab05be3414698ef58686ada879ff7d6499a4dddb64a788603ce309a120dR62) where it says, "Only capabilities that were declared mutable at creation time may be enabled, and a capability becomes immutable once enabled."

Totally fine to leave this alone, but I just found that to be more direct and clear.

sappenin (dismissed):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/MpTokenIT.java
cybele-ripple: Agreed but he DynamicMPT amendment is not yet enabled on devnet, so removing this guard causes the test to fail with `temDISABLED`.

sappenin (dismissed):

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/MpTokenIT.java
Patel-Raj11: I see. We can leave this for now, but ideally DynamicMPT is a new amendment and if we created a new IT file rather then having the tests in MpTokenIT, then we would have ability to disable all the tests via `@DisabledIf(value = "shouldNotRun", disabledReason = "DynamicMptIT only runs on local rippled node.")` and re-enable once the rippled rc gets enabled on Devnet. Currently, I think we don't know if these ITs are running on standalone or are they unintentionally being skipped in the CI.

Patel-Raj11 (dismissed):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/MpTokenIssuanceCreateMutableFlags.java
cybele-ripple: Updated the javadoc to use that phrasing: "Only capabilities that were declared mutable at creation time may be enabled, and a capability becomes immutable once enabled."

sappenin (approved):

PR #815, opened 2026-07-30
docs: add .ai-review/instructions.md for repo-specific review context

arshjafri-ripple:
## Summary

Adds a `.ai-review/instructions.md` file that the AI code reviewer reads before reviewing a pull request. It gives the reviewer context about how this codebase actually works, so its comments reflect xrpl4j's real conventions instead of generic Java advice. This is a review-guidance file only: it changes no code and no runtime behavior, and it does not affect CI or block merges.

## Why

The reviewer is general-purpose. Without repo context it tends to miss the pitfalls specific to a financial SDK and, worse, flag intentional idioms as if they were bugs. This file encodes what a new senior reviewer would need to know before commenting on xrpl4j, for example:

- Adding a transaction type touches three places, and nothing fails at compile time if one is missed: a missing `Transaction.typeMap` entry surfaces as a runtime NPE, and a missing enum constant silently deserializes as `UnknownTransaction`.
- The two ObjectMapper factories have opposite strictness on purpose: the model mapper tolerates unknown fields from rippled, while the binary codec mapper must fail on them, because a silently dropped field would change signed bytes. "Harmonizing" them either way is a bug.
- `definitions.json` is a vendored snapshot with no in-repo generator, so regeneration fidelity is what needs review, not the diff shape.
- Flag constants have sharp edges: an unsuffixed `0x80000000` sign-extends and still compiles, and a wrong constant in an `of(boolean...)` ternary ships a validly signed transaction with the wrong bit set.
- Signing rules (per-payload encoders, numeric signer ordering, low-S normalization, the empty SigningPubKey convention) where a mistake fails only on-ledger.

It also tells the reviewer which patterns are deliberate and should not be flagged: the standalone genesis seed and test keystore, fixture-driven codec tests ported from the JS implementation, Awaitility polling that ignores transient exceptions, and `@Beta` as the amendment lifecycle marker.

## What's in it

A short, high-signal list grouped by area:

- Amounts and numbers
- Binary codec
- definitions.json
- Transaction models
- Flags
- Immutables and Jackson
- Keys
- Signing
- Client
- Test conventions (idioms not to flag)
- Contributor conventions (JDK 1.8 floor, migration guides, @Beta)

The file is intentionally small so it fits the reviewer's context budget.

## How it was built and checked

- Derived from a scan of the codebase plus a review of what human reviewers have commented on in recent xrpl4j pull requests.
- Every rule was validated against the current code on `main`: each cited class, method, and constant was opened and confirmed.
- A separate verification pass then re-checked each rule for accuracy and for false-positive safety (would a reviewer applying it literally flag correct code?), with an independent second review confirming each change. Several rules were narrowed using counter-examples from this repo's own code, for example the fee helpers that use the explicitly rounded `BigDecimal.divide` overload.

## Notes for reviewers

- This is guidance, not policy. You can edit it through the normal PR process, and the reviewer will pick up the change.
- No package behavior changes, so no migration-guide entry is included.
- This follows the same approach as the equivalent files for the other XRPL SDKs (xrpl.js, xrpl-py, xrpl-rust), so review guidance stays consistent across them.


sappenin (dismissed):

.ai-review/instructions.md
```diff
@@ -0,0 +1,115 @@
...
+## Client
+- Clients never sign. `submit()` binary-encodes to `tx_blob` (flag raw-JSON blobs);
+  `submitMultisigned` sends `tx_json` by design. `JsonRpcClientErrorException` throws on
+  `result.status == "error"`; tec/tem/ter results return in `SubmitResult` — don't flag callers
+  inspecting the result instead of catching.
+- Every client-module mapper comes from `ObjectMapperFactory.create()` — flag a bare
+  `new ObjectMapper()` outside tests.
+
+## Test conventions (do NOT flag)
+- `snoPBrXtMeMyMHUVTgbuqAfg1SUTb` is the standalone genesis seed; keystore password "password"
```
cybele-ripple, line 104: Would having this instruction be an issue for possibly letting through unintended hardcoded information?

arshjafri-ripple: Good point. But since it's an exact-match allowlist, it only covers those two exact strings and doesn't work as a general "ignore hardcoded creds" rule. Both are listed literally: snoPBrXtMeMyMHUVTgbuqAfg1SUTb is XRPL's public standalone-mode genesis seed, so it holds no real value and never touches mainnet, and the keystore password is a documented test fixture. Any new hardcoded secret elsewhere wouldn't match either entry, so it should still get flagged.

.ai-review/instructions.md
```diff
@@ -0,0 +1,115 @@
...
+  LockedAmount) are base-10 — a new base-10 u64 field must be added there or misparses as hex.
+- `BinaryParser` has no per-field bounds checks: a `fromParser` under-reading its width (or the VL
+  sizeHint) silently misaligns every later field; the ObjectEndMarker is CALLER-emitted on both
+  paths — new container code must replicate it.
+- Base58 is the XRPL alphabet (`rpshnaf39w…`, zero byte = `r`), double-SHA256 checksum — never
+  substitute a Bitcoin base58 library; the broad catch→false in `isValid*` is intentional.
+
+## definitions.json
+- A vendored snapshot (rippled `--get_definitions` / xrpl.js) — **no
+  in-repo generator**; regenerate whole per `resources/README.md`, never hand-edit. A NEW
```
sappenin: > no in-repo generator

This is technically true, but the README has instructions for how to generate, so an AI agent could be capable of generating. AI blurs the line of this statement, FWIW.

.coderabbit.yaml
```diff
@@ -0,0 +1,57 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+
+# Short steer toward useful feedback for open-source contributors; skip style nits (max 250 chars).
+tone_instructions: "Be concise and encouraging to open-source contributors. Flag correctness, security, and API-contract issues; skip style and formatting nits."
```
sappenin: > skip style and formatting nits.

Are we sure we want to include this part? Style and formatting are important to get correct to prevent PR diff noise; checkstyle checks this anyway, so maybe this is fine at the AI level?

arshjafri-ripple: Reasoning was that checkstyle already enforces this deterministically in the Maven build, so bot style comments could be duplicating a blocking check and add noise. But I agree that "style and formatting" is too broad, narrowed it to "formatting and whitespace nits already covered by checkstyle."

sappenin (dismissed):

.ai-review/instructions.md
arshjafri-ripple: Reworded in [eeb52e7](https://github.com/XRPLF/xrpl4j/pull/815/commits/eeb52e7fcf71d6626698eb35142293677120c1ef).

cybele-ripple (dismissed):

arshjafri-ripple:
@sappenin could you run the CI checks for this?

sappenin (dismissed):

sappenin (approved):

cybele-ripple (approved):

arshjafri-ripple:
@sappenin could you run the CI checks again? Thanks!

sappenin (approved):

PR #818, opened 2026-08-07
temporarily use rippleci/xrpld:3.3.0 for CI

kuan121:
### Why
This is to unblock pull requests opened from forked xrpl4j that doesn't support running CI jobs against private xrpld docker container. We will revert the default image from `rippleci/xrpld:3.3.0` to `rippleci/xrpld:develop` after 3.3.0 changes are merged into the `xrpld` `develop` branch

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -19,7 +19,7 @@ jobs:
         id: resolve
         env:
           IMAGE_CONFIG_FILE: .github/xrpld-image.env
-          DEFAULT_IMAGE: rippleci/xrpld:develop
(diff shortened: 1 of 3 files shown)


.github/workflows/workflow.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         id: resolve
         env:
           IMAGE_CONFIG_FILE: .github/xrpld-image.env
-          DEFAULT_IMAGE: rippleci/xrpld:develop
+          DEFAULT_IMAGE: rippleci/xrpld:3.3.0
```
sappenin: We want to run CI against `latest`. Is there something about that version that breaks things?

sappenin (approved):

PR #821, opened 2026-08-19
build: split unit vs. aggregate coverage gates in Codecov

sappenin:
- Splits the single Codecov upload into two flagged reports: unittests (per-module unit coverage from xrpl4j-core/xrpl4j-client, gates PRs) and aggregate (unit + IT coverage via the xrpl4j-coverage module, informational only) — so integration tests can't mask real unit-test gaps.
- Adds named codecov.yml status checks (unittests-gate for both project/patch, aggregate-visibility informational-only) with disable_search: true on both upload steps so neither can fall back to auto-discovering unintended XML files.
- Feeds ConfidentialTransfersIT coverage into the aggregate from the linux x86-64 leg of build_confidential_native_its (the one deterministic native/local-rippled leg), which previously uploaded no coverage at all.
- Drops xrpl4j-mpt-crypto from the unit upload — it ships no Java sources (only downloaded native resources), so referencing its nonexistent jacoco.xml with fail_ci_if_error: true would have broken CI on merge.

  Test plan                                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                            
  - [ ] build_jdk_temurin_8 completes and both Codecov upload steps succeed                                                                                                                                                                                                                                                                                                                                                                 
  - [ ] build_confidential_native_its (linux-x86-64 leg) uploads an aggregate report including ConfidentialTransfersIT                                                                                                                                                                                                                                                                                                                      
  - [ ] Codecov PR comment shows both unittests and aggregate flags with the expected file sets                                                                                                                                                                                                                                                                                                                                             
  - [ ] After merge, update branch protection: required check renames from codecov/project → codecov/project/unittests-gate (and codecov/patch → codecov/patch/unittests-gate, if required) — do this only after one post-merge build on main uploads with the new flags


.github/workflows/workflow.yml
```diff
@@ -83,11 +83,24 @@ jobs:
           token: ${{ secrets.GITLAB_REGISTRY_TOKEN }}
```
kuan121, line 83: Can we rebased the feature branch so that the PR has only the codecov change?

sappenin:
Closing this in-favor of https://github.com/XRPLF/xrpl4j/pull/835

PR #828, opened 2026-08-27
Fix json serialization of UInt192Type

nkramer44:
UInt192Type does not override `toJson`, so when deserializing binary to JSON, `UIntType`'s `toJson` method gets invoked, which looks like

```
@Override
  public JsonNode toJson() {
    return new TextNode(UnsignedLong.valueOf(toHex(), 16).toString());
  }
```

UInt192s are too large to fit in an UnsignedLong, so this method blows up. The fix is to just overload `toJson` and write `toHex()` without wrapping it in any kind of number class


sappenin (dismissed):

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/codec/binary/BinarySerializationTests.java
```diff
@@ -2536,6 +2547,92 @@ void serializeAmmVote() throws JsonProcessingException {
...
+  @Test
+  void serializeMpTokenIssuanceDestroy() throws JsonProcessingException {
+    MpTokenIssuanceDestroy mpTokenIssuanceDestroy = MpTokenIssuanceDestroy.builder()
+      .account(Address.of("rMBzp8CgpE441cp5PVyA9rpVV7oT8hP3ys"))
+      .fee(XrpCurrencyAmount.ofDrops(10))
+      .sequence(UnsignedInteger.ONE)
+      .signingPublicKey(
+        PublicKey.fromBase16EncodedPublicKey("ED9861C4CB029C0DA737B823D7D3459A70F227958D5C0C111CC7CF947FC5A93347")
+      )
+       .mpTokenIssuanceId(MpTokenIssuanceId.of("113411307C0D97D1EC6B2989138679ACDCB75C37CA30F0A6"))
```
cybele-ripple: very minor nit: looks like the indentation here is slightly off

cybele-ripple (dismissed):

cybele-ripple (approved):

Issue #829, opened 2026-08-27
Transaction Fee Calculation Issues


kuan121:
## Summary

`FeeUtils` has grown one public `computeXyz` method per amendment. They don't compose, most are incomplete against rippled, and no combination of them prices a transaction that uses more than one feature at a time. Some fee logic also lives outside `FeeUtils` entirely, in `EscrowFinish.computeFee`.

| Added | Method | Via |
|---|---|---|
| 2022-07-18 | `computeMultisigNetworkFees(feeResult, signerList)` | original `FeeUtils` #280 |
| 2026-02-19 | `computeBatchFee(feeResult, numBatchSigners)` | Batch #654 |
| 2026-05-20 | `computeLoanSetNetworkFees(feeResult, numFirstPartySigners, numCounterpartySigners)` | Lending #719 |
| 2026-08-12 | `computeSponsorshipTransferNetworkFees(feeResult, numSponseeSigners, numSponsorSigners)` | Sponsorship #720 |
| 2026-08-25 | `computeConfidentialMptNetworkFees(feeResult, numMultisigners)` | Confidential MPT #810 |

Every amendment added a signature count under a new name — `numBatchSigners`, `numFirstPartySigners`, `numMultisigners`, `numSponseeSigners`, `numSponsorSigners`, `numCounterpartySigners`. In rippled they are all terms in `base × (1 + signatures)`.

## The complete fee model (rippled)

81 transaction types in xrpl4j; 13 have special rules. Every rule below comes from rippled's `calculateBaseFee` overrides, not the XLS specs, which do not always document transaction cost.

| Shape | Types | Rule |
|---|---|---|
| Generic | ~65 | `base × (1 + signers + sponsorSigners)` |
| Additive | 5 confidential MPT | generic `+ 9 × base` ([`kConfidentialFeeMultiplier`][conf-const]) — so **10** when single-signed |
| | `EscrowFinish` w/ fulfillment | generic `+ base × (32 + bytes/16)` |
| | `LoanSet` | generic `+ counterpartySignatures × base` |
| Replacement | `AccountDelete`, `AMMCreate` | `ownerReserve` — flat, signer terms ignored |
| Multiplicative | `LoanPay` | `feeIncrements × generic`, 1–20 |
| Conditional zero | `SetRegularKey` | `0` if master-key-signed and `!lsfPasswordSpent` |
| Recursive | `Batch` | `base × (2 + batchSigners + signers + sponsorSigners) + Σ inner fees` |
[...]

Issue #648, opened 2026-01-22
Consider making Fees "type-safe"


sappenin:
Currently, transaction fees are typed as a `XrpCurrencyAmount`. However, for certain transactions -- e.g., Batch or multisig -- supplying a normal fee would be incorrect. However, it's easy to look at `FeeUtils` and miss this. Because fees might actually work in a local development server, this could lead to bugs that are actually trivial to "avoid" by using the Java type system to aid the developer.

In addition, in the future, if transaction fees could ever be denominated in non-XRP tokens, then having a wrapper type for fees will reduce the change-set if something like non-XRP fees ever gets enabled.

One idea here could be:

1. Introduce an interface like `TransactionFee`. 
2. In `Transaction.java`, we could change the `fee()` function to potentially be generic, like this:

```java
  @JsonProperty("Fee")
  <F extends TransactionFee> fee();
```

3. In concrete transactions (e.g., `Batch.java`) we could override the fee-type to be something specific, like a `BatchTransactionFee`. 

Finally, in `FeeUtils` we could make inputs and outputs more type-safe, to prevent errors in fee calculation (e.g., see `FeeUtils#computeBatchFee`, which requires more information than a regular fee). If a developer misses this function, and errantly tries to use the other `FeeUtils` functions, it would be nice if the compiler could catch this and guide the developer to the correct methods and types.

PR #830, opened 2026-08-27
Transaction fee calculation refactoring

kuan121:
## Why

Address #829, which documents the failures this fixes, the complete rippled fee model for all 81 transaction types, and worked examples for every fee shape.

## What changed

A single fee engine mirroring rippled's `Transactor::calculateBaseFee` and its per-type overrides, reached through one method:

```java
ComputedNetworkFees computeFee(FeeParams params);
```

Everything derivable comes off the transaction — its type and therefore any surcharge, an `EscrowFinish`'s fulfillment, a Batch's inner transactions and the accounts that must sign it. What remains is what cannot be derived, because the fee is signed over and the signatures do not exist yet, plus two values `FeeResult` does not carry.

### Type-scoped entry points

`FeeParams` exposes one entry point per fee shape, each offering only the inputs that apply to it, so the builder's method list is the checklist for that transaction type:

```java
FeeParams.of(feeResult, transaction)                        // generically-priced types
FeeParams.forBatch(feeResult, batch)                        // + ownerReserve(); requires collected BatchSigners
FeeParams.forLoanSet(feeResult, loanSet)                    // + counterpartySignatureCount()
FeeParams.forLoanPay(feeResult, loanPay)                    // + loanPaymentFeeIncrements()
FeeParams.forOwnerReserve(feeResult, transaction, reserve)  // AccountDelete, AMMCreate — returns a finished FeeParams
```

Passing a transaction to the wrong entry point fails immediately and names the right one:

```
A Batch has pricing inputs of its own. Use FeeParams.forBatch(feeResult, batch), which exposes a signature
count per batch signer and an owner reserve for AccountDelete/AMMCreate inner transactions.
```

The flat `FeeParams.builder()` remains for code that must accept any `Transaction`.

### The inputs

```java
FeeResult   feeResult();
Transaction transaction();

UnsignedInteger signersCount()                     = 0;   // entries in this account's Signers array
UnsignedInteger sponsorSignersCount()              = 0;   // entries in SponsorSignature.Signers
Optional<UnsignedInteger> counterpartySignatureCount();   // LoanSet; priced as 1 when absent
Optional<UnsignedInteger> loanPaymentFeeIncrements();     // LoanPay, 1-20; priced as 1 when absent
Optional<XrpCurrencyAmount> ownerReserve();               // AccountDelete, AMMCreate; not on FeeResult
```

The two type-specific counts are `Optional` rather than defaulted deliberately. Under a default, setting `counterpartySignatureCount(ONE)` on a `Payment` was indistinguishable from leaving it unset, so validation could not reject it and the caller got no signal that the field did not apply. Absence is now distinguishable from an explicit value, so it is rejected.

Validation refuses anything that would otherwise be silently ignored: a field that does not apply to the transaction type, a signature count outside its range (0–32 for the account and sponsor counts, 1–32 for `counterpartySignatureCount` and each per-batch-signer count, all capped at the signer-list limit), a missing or superfluous `ownerReserve`, and a `Batch` whose `BatchSigners` have not yet been collected. The scoped builders validate eagerly, per call, so a mistake is reported at the line that made it.

### Itemized fee explanations

`ComputedNetworkFees` gains an auxiliary `feeBreakdown()`: one `FeeTerm` per reason the transaction owes something, each tagged `SPECIFIED`, `ASSUMED` or `DERIVED`, where the breakdown's totals *are* the fee.

```java
System.out.println(fees.feeBreakdown().get().summary());
// 2 x base    [derived]   Batch outer: base fee + batch processing fee
// 1 x base    [derived]   batch signer rBob...: 1 collected signature(s)
// 1 x base    [derived]   inner Payment (rBob...): base fee
// total: 4 x base fee
```

A Batch must carry its `BatchSigners` to be priced, so a batch signature count is never forecast. That is an ordering constraint rather than a design one: `serializeBatch` excludes the outer `Fee` from what the participants sign, so the flow is collect the batch signatures, price the Batch, attach the fee, then have the outer account sign last.

### `Transaction.fee()` now defaults to zero drops

`fee()` was a required attribute. It now defaults to `XrpCurrencyAmount.ofDrops(0)`, so a transaction can be built before its fee is known — and zero is the only fee a Batch inner transaction may legally carry, so inners no longer need `.fee(XrpCurrencyAmount.ofDrops(0))` as ceremony.

Because that removes a build-time check, every `TransactionSigner` path that signs an outer transaction's own `Fee` field — `sign()`, `multiSign()`, `sponsorSign()`, `counterpartySign()` — now rejects a transaction whose `fee()` is still zero. `signInner()` and `multiSignInner()` are exempt: they sign the `serializeBatch` digest, which excludes the outer `Fee`, so a Batch's participants may legitimately sign before its fee is computed.

### `Batch.requiredSigners()`

New, deriving the accounts that must sign from the inner transactions exactly as `Batch::preflightSigValidated` does — each inner's initiator (its `Delegate` or `Account`), its `Sponsor` when the inner carries a `SponsorSignature`, and (forward-compat, currently gated off by rippled's `kDisabledTxTypes`) a `LoanSet`'s `Counterparty`. It lives on `Batch` rather than in `FeeUtils` because it is a property of a Batch rather than of fees — anyone assembling one needs to know who to circulate it to. It also replaces a second, drifted copy of the same derivation inside `checkBatchSigners`, which had omitted sponsors.

## API changes

| Method | Change |
|---|---|
| `computeFee(FeeParams)` | **new** — the single fee entry point |
| `computeNetworkFees(FeeResult)` | **deprecated** — returns only the ledger's base fee levels, so it silently under-charges any transaction that does not cost exactly one base fee, including any multi-signed or sponsored one |
| `computeMultisigNetworkFees` | **deprecated** — counts signer-*list* entries rather than the signatures the transaction will carry, so it over-charges when a quorum is met by fewer signers than the list holds, and has no sponsor term |
| `EscrowFinish.computeFee` | **deprecated** — omits the per-signature terms rippled adds the fulfillment surcharge on top of; shipped since v5.1.0, so kept along with its tests |
| `computeBatchFee` (both overloads) | **removed** |
| `computeLoanSetNetworkFees` | **removed** |
| `computeConfidentialMptNetworkFees` | **removed** |
| `computeSponsorshipTransferNetworkFees` | **removed** |

New public types: `FeeParams` (with `GenericBuilder`, `BatchBuilder`, `LoanSetBuilder`, `LoanPayBuilder`), `FeeBreakdown`, `FeeTerm` and `FeeTerm.Provenance`.

The four removals shipped in `v6.1.0-rc.1`/`rc.2` but never in a stable release — `v6.0.0` contains only `computeNetworkFees(FeeResult)` and `computeMultisigNetworkFees` — so **no stable-release user is forced to change anything**, and this is the last point at which removing them is free rather than a deprecation cycle. The deprecations above are opt-in timing rather than a break.

## The call-site shape

Because a fee is derived from the transaction's contents and is itself signed over, it must be computed from a built transaction and then applied to a copy. With `fee()` now defaulting to zero, no placeholder is needed:

```java
final ImmutableBatch unpricedBatch = Batch.builder()
  .account(alice)
  .sequence(aliceSequence)
  .signingPublicKey(alicePublicKey)
  .addRawTransactions(
    RawTransactionWrapper.of(paymentFromBob),
    RawTransactionWrapper.of(paymentFromCarol))
  .build();

final Batch batch = unpricedBatch.withFee(
  FeeUtils.computeFee(FeeParams.forBatch(feeResult, unpricedBatch).build()).recommendedFee()
);
```

Declaring the unpriced local as the concrete `ImmutableBatch` — what `Batch.builder().build()` already returns — makes `withFee` directly reachable, so no `ImmutableBatch.copyOf(...)` is required.

## Testing

- `FeeUtilsTest` — `computeFee` against real transactions rather than mocks: every row of the fee model in #829, each validation rejection, and Batch inners for confidential MPT, `EscrowFinish` (with and without a fulfillment), `LoanSet` and owner-reserve types.
- `FeeParamsScopedApiTest` — the five entry points, their steering errors, eager validation, strict mode, and the breakdown.
- `FeeBreakdownTest` / `FeeTermTest` — totals, flat-only breakdowns, `summary()` rendering, and provenance tagging.
- `AbstractTransactionSignerTest` — the zero-fee guard on each signing path.
- `BatchTest` — `requiredSigners()`, including delegated and sponsored inners.
- 20 direct `FeeUtils.computeFee` call sites migrated across `BatchTransactionIT`, `LendingProtocolIT`, `AbstractConfidentialMptIT` and `SponsorshipIT`, plus the shared `feeFor(...)` helper.

Fee values were also differentially verified: a 42-scenario matrix comparing computed drops at all three fee levels before and after the refactor — single and multi-sign, sponsored, four `EscrowFinish` sizes, confidential MPT, LoanSet and LoanPay variants, owner reserve, and ten Batch shapes — is identical except for the two inputs now correctly rejected as inapplicable.

## Migration

`V7_MIGRATION.md` gains section 3, *Transaction Fee Model*, covering the `fee()` relaxation and its signing guard, the removals and deprecations with replacements, and the new `FeeParams` / `FeeTerm` / `FeeBreakdown` API.

## Deliberately out of scope

`SetRegularKey`'s free case is charged the generic fee — a harmless overpay, since detecting it needs the AccountRoot's `lsfPasswordSpent` flag. `LoanPay` increments are supplied rather than derived, since deriving them needs the `Loan`, `LoanBroker` and `Vault` ledger objects. `LedgerStateFix` uses rippled's owner-reserve rule but is not modeled by xrpl4j, so it resolves to `UNKNOWN` and is rejected rather than mispriced.

The output half of #648 is untouched. This PR makes the fee calculation impossible to call wrongly; that issue is about making it impossible to skip, which needs a `TransactionFee` wrapper so a hand-computed `XrpCurrencyAmount` cannot be set on a Batch at all. Worth noting it is smaller than it reads — only the ~11 types with a non-generic fee shape need it, and it can start with `Batch` alone.


kuan121:
@coderabbitai configuration

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -120,200 +153,240 @@ public static ComputedNetworkFees computeNetworkFees(final FeeResult feeResult)
...
+   * @param escrowFinish The {@link EscrowFinish} being priced.
    *
-   * <p>Per the Lending Protocol specification, the total fee for a standalone {@code LoanSet}
-   * transaction (not part of a Batch) is {@code (1 + |tx.Signers| + |signatures|) × base_fee}, where
-   * {@code |signatures| = max(1, |tx.CounterpartySignature.Signers|)}. The minimum fee is always
-   * {@code 2 × base_fee}, even without a {@code tx.Signers} list.</p>
+   * @return A number of extra base fees.
+   */
+  private static long fulfillmentUnits(final EscrowFinish escrowFinish) {
+    return escrowFinish.fulfillment()
```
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/changes/aa903840e47e25f5e4773179cc49ac515ff2e64b

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -120,200 +153,240 @@ public static ComputedNetworkFees computeNetworkFees(final FeeResult feeResult)
...
+          "Only PreimageSha256Fulfillment is supported, but the fulfillment was a %s.",
+          fulfillment.getClass().getSimpleName()
+        );
+        final long fulfillmentByteSize = Base64.getUrlDecoder()
+          .decode(((PreimageSha256Fulfillment) fulfillment).getEncodedPreimage()).length;
+        // See rippled's EscrowFinish::calculateBaseFee: extraFee = base * (32 + (fulfillment size / 16)).
+        return 32L + (fulfillmentByteSize / 16L);
+      })
+      .orElse(0L);
+  }
```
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/commits/aa903840e47e25f5e4773179cc49ac515ff2e64b

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,312 @@
...
+    this.signaturesPerBatchSigner().forEach((address, count) -> {
+      Preconditions.checkArgument(
+        requiredSigners.contains(address),
+        "%s is not required to sign this Batch, so a signature count for it would be ignored. The accounts that " +
+          "must sign are %s. Note that a required signer is not always an inner transaction's Account: a delegated " +
+          "inner is signed by its Delegate, and a sponsored inner also requires its Sponsor.",
+        address, requiredSigners
+      );
+      checkSignatureCount(count, "signaturesPerBatchSigner[" + address + "]");
+    });
```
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/commits/48d532324679836eaea27296c82a9558c011e06b

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Batch.java
```diff
@@ -139,6 +141,50 @@ default BatchFlags flags() {
   @JsonProperty("BatchSigners")
   List<BatchSignerWrapper> batchSigners();
 
+  /**
+   * The set of accounts that must appear in {@link #batchSigners()}, derived from the inner transactions.
+   *
+   * <p>rippled builds the identical set in {@code Batch::preflightSigValidated} and rejects any {@code BatchSigners}
+   * array that does not match it exactly. Each inner contributes:
```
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/commits/48d532324679836eaea27296c82a9558c011e06b

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -120,200 +151,239 @@ public static ComputedNetworkFees computeNetworkFees(final FeeResult feeResult)
...
+    long ownerReserveInners = 0L;
+
+    for (final RawTransactionWrapper wrapper : batch.rawTransactions()) {
+      final Transaction inner = wrapper.rawTransaction();
+      if (FeeParams.OWNER_RESERVE_TRANSACTION_TYPES.contains(inner.transactionType())) {
+        ownerReserveInners++;
+      } else {
+        // An inner transaction never carries signatures or fee sponsorship, so it costs one base fee plus whatever
+        // surcharge its type attracts.
+        feeUnits += 1L + surchargeUnits(feeParams, inner, true);
```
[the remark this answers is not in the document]
kuan121: Valid in principle, but `ttLOAN_PAY` is in rippled's `kDisabledTxTypes`, so a `LoanPay` can't be a Batch inner today (rejected `temINVALID_INNER_BATCH`). Pricing a multi-increment inner would also need a per-inner increments input — the current `loanPaymentFeeIncrements` is a single outer field forced to `1` for a Batch — which is a speculative API addition for an unreachable case. Deferring until `LoanPay` is permitted as a Batch inner; for the single-increment default the current fee is already correct.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,320 @@
...
+   * The number of signatures the transaction's sponsor will supply in its {@code SponsorSignature.Signers} array.
+   *
+   * <p>Zero when the transaction is unsponsored, and also when the sponsor signs with a single key: rippled counts
+   * only {@code SponsorSignature.Signers} entries, and a lone sponsor signature is carried in
+   * {@code SponsorSignature.TxnSignature} instead.
+   *
+   * @return An {@link UnsignedInteger} number of signatures, defaulting to zero.
+   */
+  @Value.Default
+  default UnsignedInteger sponsorSignersCount() {
```
sappenin: What if instead of `@Default` zero, we used `Optional` instead?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,320 @@
...
+   * The number of signatures the transaction's own account will supply in its {@code Signers} array.
+   *
+   * <p>Zero for a single-signed transaction: rippled counts only {@code Signers} entries, and a lone signature is
+   * carried in {@code TxnSignature} instead. For a multi-signature this is the full size of the {@code Signers}
+   * array (e.g. 4 for a 4-of-N signer list), not one fewer — rippled charges {@code base × (1 + signersCount)}.
+   *
+   * @return An {@link UnsignedInteger} number of signatures, defaulting to zero.
+   */
+  @Value.Default
+  default UnsignedInteger signersCount() {
```
sappenin: What if instead of `@Default` zero, we used `Optional` instead?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,320 @@
...
+   * The number of fee increments a {@code LoanPay} transaction will be charged, being one increment per
+   * {@code kLoanPaymentsPerFeeIncrement} payments the transaction is estimated to make.
+   *
+   * <p>Deriving this requires the {@code Loan}, {@code LoanBroker} and {@code Vault} ledger objects, so it is supplied
+   * rather than computed. One — a single payment — is the default.
+   *
+   * @return An {@link UnsignedInteger} number of fee increments, defaulting to one.
+   */
+  @Value.Default
+  default UnsignedInteger loanPaymentFeeIncrements() {
```
sappenin: Seems like we can be smarter here - like, if the transaction is not related to Lending Protocol, then this doesn't matter. But if it is, then maybe this should be set to something overt?

One thought here is to make this `Optional` and then do some verification in an `@Check` method?

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/AbstractConfidentialMptIT.java
```diff
@@ -467,15 +493,17 @@ protected SingleSignedTransaction<ConfidentialMptClawback> buildClawback(
 
     final ConfidentialMptClawback clawback = ConfidentialMptClawback.builder()
       .account(issuance.issuer.publicKey().deriveAddress())
-      .fee(confidentialFee())
+      .fee(XrpCurrencyAmount.ofDrops(0))
```
sappenin: This pattern is awkward for me. Seems like maybe the `fee` should just be optional, and then later it should be checked perhaps (if it's required?). Or, if you prefer it to be 0, then make that a default on the `Transaction`?

I suppose I'm still a bit unclear when this is supposed to get set...

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,320 @@
...
+   *
+   * <p>Unlike {@link #signersCount()} and {@link #sponsorSignersCount()} — which are zero when the lone signature
+   * rides in {@code TxnSignature} — a single counterparty signature is itself charged: rippled counts
+   * {@code CounterpartySignature.TxnSignature} as one. One is therefore the correct value for a counterparty signing
+   * with a single key, and is the default.
+   *
+   * @return An {@link UnsignedInteger} number of signatures, defaulting to one.
+   */
+  @Value.Default
+  default UnsignedInteger counterpartySignatureCount() {
```
sappenin: See above. I think this should be `Optional`.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,320 @@
...
+   * <p>Keys must be members of
+   * {@link org.xrpl.xrpl4j.model.transactions.Batch#requiredSigners() Batch#requiredSigners()} — the accounts that
+   * must sign, derived from the inner
+   * transactions. Note that a required signer is not always an inner's {@code Account}: a delegated inner is signed by
+   * its {@code Delegate}, and a sponsored inner also requires its {@code Sponsor}.
+   *
+   * @return A {@link Map} from a Batch participant's {@link Address} to the number of signatures it will supply,
+   *   defaulting to empty.
+   */
+  Map<Address, UnsignedInteger> signaturesPerBatchSigner();
```
sappenin: Do we need this method? It seems confusing, and unclear that a consumer of xrpl4j will know how to populate this well. 

Put another way, why not just require that `Batch#batchSigners()` is filled out in order to get a fee?

Another thing making me think this is not needed in xrpl4j:

> This is needed only when pricing a Batch <em>before</em> its {@code BatchSigners} exist — a wallet displaying a fee up front, for instance. 

...it's likely that a wallet doing this will not use xrpl4j, but will use a TS or WASM library instead for its fee computations. If a backend developer wanted to do this, they could easily simulate this using fake public keys in the right spots (if they knew the number of signers).

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
```diff
@@ -103,9 +105,9 @@ void loanSetWithSingleSignBrokerAndSingleSignCounterparty()
     final AccountInfoResult brokerAccountInfo = this.scanForResult(
       () -> this.getValidatedAccountInfo(brokerKeyPair.publicKey().deriveAddress())
     );
-    final LoanSet unsignedLoanSet = LoanSet.builder()
+    final LoanSet unpricedLoanSet = LoanSet.builder()
       .account(brokerKeyPair.publicKey().deriveAddress())
-      .fee(FeeUtils.computeLoanSetNetworkFees(feeResult, UnsignedInteger.ZERO, UnsignedInteger.ZERO).recommendedFee())
+      .fee(XrpCurrencyAmount.ofDrops(0))
```
sappenin: I need to think a bit more about this one. It's pretty non-obvious that, in order to compute fees, a developer should set them to `0` and then recompute them.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeBreakdown.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * An itemized account of how a transaction's fee was computed by {@link FeeUtils#computeFee(FeeParams)}: one
+ * {@link FeeTerm} per reason the transaction owes something, each tagged with where its value came from.
+ *
+ * <p>The intended use is iterative: compute a fee, look at (or log) {@link #summary()}, and check whether any
+ * {@link FeeTerm.Provenance#ASSUMED} line describes an assumption that is wrong for the transaction at hand — e.g., a
+ * Batch participant assumed to sign with a single key when it will actually multi-sign. Each assumed line names the
+ * {@link FeeParams} input that overrides it. When every line is {@code specified} or {@code derived}, the fee is
+ * exact.
```
sappenin, line 39: I'm thinking we should make a logging `warning` whenever a `FeeBreakdown` is produced with an assumed value. That way the developer can hopefully notice that they didn't specify a value, and double-check that they're not getting any assumed rates. I'll make a suggestion down in FeeUtils.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -185,98 +188,247 @@ private static ComputedNetworkFees computeBaseFees(final FeeResult feeResult) {
...
+      FeeTerm.of(
+        "own multi-signature: " + signersCount + " Signers entries", signersCount, FeeTerm.Provenance.SPECIFIED
+      ) :
+      FeeTerm.of(
+        "assumed single-signed: a lone TxnSignature is free (set signersCount if this account will multi-sign)",
+        0L, FeeTerm.Provenance.ASSUMED
+      ));
+
+    final long sponsorSignersCount = feeParams.sponsorSignersCount().longValue();
+    if (sponsorSignersCount > 0L) {
```
sappenin: Can be replaced with single expression in functional style:

```java
return Optional.of(specifiedCount.map(unsignedInteger -> FeeTerm.of(
        "LoanSet counterparty multi-signature: " + unsignedInteger + " signature(s)",
        unsignedInteger.longValue(), FeeTerm.Provenance.SPECIFIED
      )).orElseGet(() -> FeeTerm.of(
        "LoanSet counterparty signature: assumed a single key, which is itself charged (set " +
          "counterpartySignatureCount if the counterparty will multi-sign)",
        1L, FeeTerm.Provenance.ASSUMED
      )));
```

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
```diff
@@ -185,98 +188,247 @@ private static ComputedNetworkFees computeBaseFees(final FeeResult feeResult) {
...
    */
   public static ComputedNetworkFees computeFee(final FeeParams feeParams) {
     Objects.requireNonNull(feeParams);
 
+    final FeeBreakdown feeBreakdown = computeFeeBreakdown(feeParams);
     final ComputedNetworkFees baseFees = computeBaseFees(feeParams.feeResult());
+
+    final long feeUnits = feeBreakdown.totalFeeUnits();
+    final Optional<XrpCurrencyAmount> flatAmount = feeBreakdown.totalFlatAmount();
+
```
sappenin, line 225: Per my comment above, if we add a private helper function like below, then we can call it here on line 22:

```java
warnIfAssumed(feeParams, feeBreakdown);
```

The function would be:

```java
  /**
   * Logs a warning naming every {@link FeeTerm.Provenance#ASSUMED} term in {@code feeBreakdown}, so that a developer
   * who did not intend to rely on a default (e.g., a Batch participant assumed to single-sign) has a chance to notice
   * before trusting the computed fee.
   *
   * @param feeParams    The {@link FeeParams} that produced {@code feeBreakdown}, identifying the transaction in the
   *                     log message.
   * @param feeBreakdown The {@link FeeBreakdown} to inspect.
   */
  private static void warnIfAssumed(final FeeParams feeParams, final FeeBreakdown feeBreakdown) {
    if (!LOGGER.isWarnEnabled()) {
      return;
    }
    final boolean hasAssumedTerm = feeBreakdown.terms().stream()
      .anyMatch(term -> term.provenance() == FeeTerm.Provenance.ASSUMED);
    if (hasAssumedTerm) {
      LOGGER.warn(
        "computeFee assumed one or more unspecified inputs while pricing a {} transaction; the computed fee may be "
          + "wrong if any assumption does not hold. Review the [assumed] lines below and supply the named FeeParams "
          + "input if so:\n{}",
        feeParams.transaction().transactionType().value(),
        feeBreakdown.summary()
      );
    }
  }
```

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
sappenin: Setting the fee like this is no longer necessary.

V7_MIGRATION.md
```diff
@@ -14,6 +14,10 @@ Version 7.0.0 introduces several breaking changes:
    `SignatureUtils.addMultiSignaturesToTransaction()` have been removed. Transaction signing now uses Immutables-generated
    `withTransactionSignature()` and `withSigners()` methods on `Transaction`, and validation has moved to `@Check`
    methods on `SingleSignedTransaction` and `MultiSignedTransaction`.
+3. **Transaction Fee Model Refactor** — `Transaction.fee()` is no longer a required field; it now defaults to zero
+   drops. A new fee-computation API (`FeeParams`, `FeeTerm`, `FeeBreakdown`, and an expanded `FeeUtils`) replaces the
+   old per-transaction-type `computeFee()` helpers for computing an accurate fee. To guard against the new default,
+   All `TransactionSigner` outer-transaction signing methods (`sign()`, `multiSign()`, `sponsorSign()`, etc.) now reject a transaction whose `fee()` is still zero.
```
sappenin: small typo: `All` should be `all`.

V7_MIGRATION.md
```diff
@@ -178,7 +182,83 @@ for every transaction subclass:
 - `Transaction withSigners(Iterable<? extends SignerWrapper> signers)` — returns a copy of the transaction with the
   specified signers applied.
 
-### 3. `MetaMpTokenIssuanceObject.mpTokenMetadata()` type change
+### 3. Transaction Fee Model
+
+`Transaction.fee()` and the fee-computation helpers have been reworked to support the new fee rules introduced by
+Batch, LoanSet/LoanPay, sponsored transactions, and confidential MPT transactions — none of which the old
```
sappenin: Remove dash, use a comma or paranthentical.

sappenin (changes requested):

V7_MIGRATION.md
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/commits/1a59ed3ddc4aa46c2140b227d7f9b60635ef7f3e

xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/LendingProtocolIT.java
kuan121: Addressed in https://github.com/XRPLF/xrpl4j/pull/830/changes/a356159af38203d817e340e44c79348dccc9cf3f

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
kuan121: Good point. I made the change in https://github.com/XRPLF/xrpl4j/pull/830/changes/47ec94503808708da02899f0899daa7e99417c9a

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
kuan121: Good point. I made the change in https://github.com/XRPLF/xrpl4j/pull/830/changes/47ec94503808708da02899f0899daa7e99417c9a

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
kuan121: Good point. I made the change in https://github.com/XRPLF/xrpl4j/pull/830/changes/a61e10140ddfee000969f626fa73562bd5888153

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
kuan121: Good point. I made the changes in https://github.com/XRPLF/xrpl4j/pull/830/changes/4d32b35734ed04f7397806bfe865ce4078c64a23. One difference I want to point it out is we log only when the assumption cost additional fee, i.e.,  `term -> term.provenance() == FeeTerm.Provenance.ASSUMED && term.feeUnits() > 0L`

sappenin: > One difference I want to point it out...

hmm... At first I agreed with that, but thinking more deeply: the intent of the warning is to tell a developer, "hey, you forgot to set a field when you assembled your inputs, so the library assumed something for you." 

That message shouldn't go away if the assumed fee is zero (I think) because the developer may not simply want to know about assumed fees that cost them money (e.g., the developer may want to make sure they don't underprice their fee by accident).

All that to say, I think we should remove the `&& term.feeUnits() > 0L` part of the determination.

Thoughts?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
kuan121: After giving it some more thought, I agree with you. This commit (https://github.com/XRPLF/xrpl4j/pull/830/changes/1c8ab20ae6ad0b043966d954e3de963a0f4a5e43) removes `signaturesPerBatchSigner`, `signaturesFor(...)`, and `requireExplicitSignatureCounts()`. `FeeParams.check()` now also refuses to price a Batch until its `BatchSigners` have been collected, so all signature counts are derived directly from the transaction.

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeUtils.java
kuan121: That's fair. I removed it in https://github.com/XRPLF/xrpl4j/pull/830/changes/0a6c0c5976c5f251e86d44be1fa8aeeac16a1822

sappenin (approved):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/client/fees/FeeParams.java
```diff
@@ -0,0 +1,741 @@
...
+   */
+  default void checkBatchIsSigned(final TransactionType transactionType) {
+    if (transactionType != TransactionType.BATCH) {
+      return;
+    }
+
+    final Batch batch = (Batch) this.transaction();
+    final Set<Address> requiredSigners = batch.requiredSigners();
+    Preconditions.checkArgument(
+      !batch.batchSigners().isEmpty() || requiredSigners.isEmpty(),
```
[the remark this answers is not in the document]
kuan121: This is already enforced one level down in `Batch.java`. `Batch.checkBatchSigners()` [Check 4](https://github.com/XRPLF/xrpl4j/blob/transaction-fee-calculation-refactoring/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/Batch.java#L358) rejects any Batch whose non-empty `BatchSigners` doesn't cover every account in `requiredSigners()`.

PR #832, opened 2026-09-03
Changes to single asset vault

cybele-ripple:
## Overview
Support LendingProtocolV1_1
rippled: https://github.com/XRPLF/rippled/pulls?page=1&q=is%3Apr+label%3A%22feature%3A+LP1.1%22
Specs:
https://github.com/XRPLF/XRPL-Standards/pull/587
https://github.com/XRPLF/XRPL-Standards/pull/582
## High Level Overview of Change

Adds support for the pending XLS-65 (Single Asset Vault) updates to the vault transaction and ledger-object models:

- **`VaultDelete.MemoData`** — a new optional field letting the vault owner record why a vault is being deleted. Hex-encoded, limited to 256 bytes (reuses the existing `VaultData` wrapper). Requires the `LendingProtocolV1_1` amendment.
- **`VaultFlags`** (ledger object) — adds `lsfVaultDepositBlocked` (`0x00020000`) and `lsfVaultOwnerCanBlockDeposit` (`0x00040000`).
- **`VaultCreateFlags`** — adds `tfVaultOwnerCanBlockDeposit` (`0x00040000`), which may only be set at vault creation.
- **`VaultSetFlags`** (new) — `tfVaultDepositBlock` (`0x00010000`) and `tfVaultDepositUnblock` (`0x00020000`), letting the vault owner block and unblock deposits. `VaultSet#flags()` now returns this type instead of the generic `TransactionFlags`.
- **`VaultDepositFlags`** (new) — `tfVaultDonate` (`0x00010000`), letting the vault owner donate assets into a vault without receiving shares in return. `VaultDeposit#flags()` now returns this type instead of the generic `TransactionFlags`.

Both new flag classes reject invalid state up front: `VaultSetFlags` refuses `tfVaultDepositBlock` combined with `tfVaultDepositUnblock` (rippled returns `temMALFORMED`), and `VaultSet` carries a matching `@Value.Check` so the same combination is caught when flags are built from a raw value via `VaultSetFlags.of(long)`.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

**On the breaking change:** `VaultSet#flags()` and `VaultDeposit#flags()` narrow their return type from `TransactionFlags` to `VaultSetFlags` / `VaultDepositFlags`. Callers passing a bare `TransactionFlags` to either builder will no longer compile. Both transaction types are `@Beta` (SingleAssetVault is not enabled on mainnet), and the repo has precedent for this narrowing (`Payment`→`PaymentFlags`, `OfferCreate`→`OfferCreateFlags`, `VaultCreate`→`VaultCreateFlags`).

To keep the narrowing from removing capability, both new flag classes expose `INNER_BATCH_TXN` / `tfInnerBatchTxn()` / `Builder#tfInnerBatchTxn(boolean)`, matching `SponsorshipSetFlags` and the other 17 `TransactionFlags` subclasses. Without it, `VaultSet` and `VaultDeposit` could no longer be marked as Batch inner transactions, since `RawTransactionWrapper` gates inner-Batch membership on `transactionFlags().tfInnerBatchTxn()`.

### Did you update HISTORY.md?

- [ ] Yes
- [x] No, this change does not impact library users

This repository contains no `HISTORY.md` or `CHANGELOG.md`, so there is nothing to update. The change *does* affect library users (see the breaking change above) — if release notes are tracked elsewhere, the `flags()` narrowing should be called out there.

## Test Plan

Unit tests only; all new code paths are covered so codecov patch coverage passes.

**New test classes**
- `VaultSetFlagsTests` — flag construction, derivation from a raw value, `empty()`, and JSON round-trip across every valid combination of `tfVaultDepositBlock` / `tfVaultDepositUnblock` / `tfInnerBatchTxn`, plus an explicit assertion that the mutually exclusive block+unblock pair throws.
- `VaultDepositFlagsTests` — the same matrix for `tfVaultDonate` / `tfInnerBatchTxn`.
- `VaultSetTest` / `VaultDepositTest` — assert `flags()` returns the new *types* (not merely equal values) and cover the `@Value.Check` on the block/unblock combination. This matters because `Flags#equals` compares only `getValue()`, so a type-blind assertion would pass even if the narrowing had never happened.

**Extended existing tests**
- `VaultDeleteJsonTest` — JSON round-trip of a `VaultDelete` carrying `MemoData`.
- `VaultSetJsonTest` / `VaultDepositJsonTest` — round-trips with the block, unblock, and donate flags set.
- `VaultObjectTest` — round-trip of a `Vault` ledger entry with `Flags: 458752`, asserting all three `lsf*` accessors after deserialization.
- `VaultCreateTest` / `VaultCreateJsonTest` — extended for `tfVaultOwnerCanBlockDeposit`.
- `TransactionTest` — covariance of the reflective `transactionFlags()` for both new flag types.
- `SignatureUtilsTest` — signs a `VaultDelete` carrying `MemoData` and `VaultSet` / `VaultDeposit` with non-empty typed flags. This is the only path in CI that exercises the binary codec for the new field; `MemoData` encodes as a top-level VL Blob (field header `7D`, ordered after `VaultID` and before `Account`) and is included in `encodeForSigning`.

### DIFF
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
@@ -1,0 +1,103 @@
+package org.xrpl.xrpl4j.model.flags;
+
+import com.google.common.annotations.Beta;
+import org.xrpl.xrpl4j.model.transactions.VaultDeposit;
+
+/**
+ * A set of static {@link TransactionFlags} which can be set on {@link VaultDeposit} transactions.
+ *
+ * <p>This class will be marked {@link Beta} until the SingleAssetVault amendment is enabled on mainnet. Its API is
+ * subject to change.</p>
+ */
+@Beta
+public class VaultDepositFlags extends TransactionFlags {
+
+  /**
+   * Constant {@link VaultDepositFlags} for the {@code tfVaultDonate} flag.
+   */
+  public static final VaultDepositFlags VAULT_DONATE = new VaultDepositFlags(0x00010000L);
+
+  private VaultDepositFlags(long value) {
+    super(value);
+  }
+
+  private VaultDepositFlags() {
+  }
+
+  /**
+   * Create a new {@link Builder}.
+   *
+   * @return A new {@link Builder}.
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * Construct {@link VaultDepositFlags} with a given value.
+   *
+   * @param value The long-number encoded flags value of this {@link VaultDepositFlags}.
+   *
+   * @return New {@link VaultDepositFlags}.
+   */
+  public static VaultDepositFlags of(long value) {
+    return new VaultDepositFlags(value);
+  }
+
+  private static VaultDepositFlags of(boolean tfFullyCanonicalSig, boolean tfVaultDonate) {
+    long value = Flags.of(
+      tfFullyCanonicalSig ? TransactionFlags.FULLY_CANONICAL_SIG : UNSET,
+      tfVaultDonate ? VAULT_DONATE : UNSET
+    ).getValue();
+    return new VaultDepositFlags(value);
+  }
+
+  /**
+   * Construct an empty instance of {@link VaultDepositFlags}. Transactions with empty flags will not be serialized
+   * with a {@code Flags} field.
+   *
+   * @return An empty {@link VaultDepositFlags}.
+   */
+  public static VaultDepositFlags empty() {
+    return new VaultDepositFlags();
+  }
+
+  /**
+   * If enabled, the deposited assets are donated to the vault, meaning the depositor receives no shares in return.
+   * Only the vault owner may donate to a vault.
+   *
+   * @return {@code true} if {@code tfVaultDonate} is set, otherwise {@code false}.
+   */
+  public boolean tfVaultDonate() {
+    return this.isSet(VaultDepositFlags.VAULT_DONATE);
+  }
+
+  /**
+   * A builder class for {@link VaultDepositFlags} flags.
+   */
+  public static class Builder {
+
+    private boolean tfVaultDonate = false;
+
+    /**
+     * Set {@code tfVaultDonate} to the given value.
+     *
+     * @param tfVaultDonate A boolean value.
+     *
+     * @return The same {@link Builder}.
+     */
+    public Builder tfVaultDonate(boolean tfVaultDonate) {
+      this.tfVaultDonate = tfVaultDonate;
+      return this;
+    }
+
+    /**
+     * Build a new {@link VaultDepositFlags} from the current boolean values.
+     *
+     * @return A new {@link VaultDepositFlags}.
+     */
+    public VaultDepositFlags build() {
+      return VaultDepositFlags.of(true, tfVaultDonate);
+    }
+  }
+}
--- a/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
+++ b/xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
@@ -8,6 +8,8 @@ import org.xrpl.xrpl4j.model.transactions.VaultDeposit;
  *
  * <p>This class will be marked {@link Beta} until the SingleAssetVault amendment is enabled on mainnet. Its API is
  * subject to change.</p>
+ *
+ * @see "https://github.com/XRPLF/XRPL-Standards/pull/469"
  */
 @Beta
 public class VaultDepositFlags extends TransactionFlags {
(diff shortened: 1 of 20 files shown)


xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultDeposit.java
```diff
@@ -29,14 +29,14 @@ static ImmutableVaultDeposit.Builder builder() {
...
+   * Set of {@link VaultDepositFlags}s for this {@link VaultDeposit}.
    *
-   * @return Always {@link TransactionFlags#EMPTY}.
+   * @return A {@link VaultDepositFlags}, which defaults to {@link VaultDepositFlags#empty()}.
    */
   @JsonProperty("Flags")
   @Value.Default
-  default TransactionFlags flags() {
-    return TransactionFlags.EMPTY;
+  default VaultDepositFlags flags() {
```
sappenin, line 38: Please update `V7_MIGRATION.md` with these breaking changes.

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtilsTest.java
```diff
@@ -1437,6 +1440,7 @@ void addSignatureToVaultSet() {
     VaultSet transaction = VaultSet.builder()
       .account(sourcePublicKey.deriveAddress())
       .vaultId(Hash256.of("0123456789012345678901234567890123456789012345678901234567891234"))
+      .flags(VaultSetFlags.builder().tfVaultDepositBlock(true).build())
```
sappenin, line 1443: Does a `VaultSet` work when flags are unset?

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * A set of static {@link TransactionFlags} which can be set on {@link VaultDeposit} transactions.
+ *
+ * <p>This class will be marked {@link Beta} until the SingleAssetVault amendment is enabled on mainnet. Its API is
+ * subject to change.</p>
+ *
+ * @see "https://github.com/XRPLF/XRPL-Standards/pull/469"
+ */
+@Beta
+public class VaultDepositFlags extends TransactionFlags {
```
sappenin, line 15: Do we need to do anything in `SingleAssetVaultIT.java` for these flag additions? Probably worth a quick test that include all flags on a deposit and set, just to make sure these actually work.

sappenin (changes requested):

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/transactions/VaultDeposit.java
cybele-ripple: Added a new "VaultDeposit and VaultSet flags() type change" [section](https://github.com/XRPLF/xrpl4j/blob/53d66bdc8da9bc9be2953b96f5cd2ca00ed1c88e/V7_MIGRATION.md?plain=1#L198-L213)

xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtilsTest.java
cybele-ripple: [Added addSignatureToVaultSetWithoutFlags()](https://github.com/XRPLF/xrpl4j/blob/53d66bdc8da9bc9be2953b96f5cd2ca00ed1c88e/xrpl4j-core/src/test/java/org/xrpl/xrpl4j/crypto/signing/SignatureUtilsTest.java#L1452-L1463) to `SignatureUtilsTest.java` which builds and signs a `VaultSet` without calling .flags() and confirmed that this falls back to `VaultSetFlags.empty()`

xrpl4j-core/src/main/java/org/xrpl/xrpl4j/model/flags/VaultDepositFlags.java
cybele-ripple: Added [vaultSetDepositBlockAndUnblock()](https://github.com/XRPLF/xrpl4j/blob/53d66bdc8da9bc9be2953b96f5cd2ca00ed1c88e/xrpl4j-integration-tests/src/test/java/org/xrpl/xrpl4j/tests/SingleAssetVaultIT.java#L1380-L1516) in SingleAssetVaultIT.java, which creates a vault with `VAULT_OWNER_CAN_BLOCK_DEPOSIT`, blocks deposits via `VaultSet.tfVaultDepositBlock, verifies a subsequent deposit is rejected, then unblocks via tfVaultDepositUnblock and verifies deposits succeed again.

sappenin (dismissed):

xrpl4j-core/src/main/resources/definitions.json
```diff
@@ -928,6 +928,26 @@
         "type": "UInt32"
       }
     ],
+    [
```
kuan121: The LEVersion field is missing

PR #835, opened 2026-09-14
build: split unit vs. aggregate coverage gates in Codecov

sappenin:
## Summary
- Splits the single Codecov upload into two flagged reports: `unittests` (per-module unit coverage from `xrpl4j-core`/`xrpl4j-client`, gates PRs) and `aggregate` (unit + IT coverage via a new `xrpl4j-coverage` module, informational only) — so integration tests can't mask real unit-test gaps.
- Adds `codecov.yml` status checks (`unittests-gate` for both project/patch, `aggregate-visibility` informational-only) with `disable_search: true` on both upload steps so neither can fall back to auto-discovering unintended XML files.
- Feeds Confidential MPT IT coverage into the aggregate from the linux x86-64 leg of `build_confidential_native_its` (the one deterministic native/local-rippled leg), which previously uploaded no coverage at all.
- Drops `xrpl4j-mpt-crypto` from the unit upload — it ships no Java sources (only downloaded native resources), so referencing its nonexistent `jacoco.xml` with `fail_ci_if_error: true` would have broken CI on merge.

Cherry-picked cleanly onto current `main` from two commits originally developed on a branch that also carried the (already-merged) Confidential Transfer for MPT work, so this PR is scoped to just the coverage-gate change.

## Test plan
- [x] `mvn -pl xrpl4j-core,xrpl4j-client,xrpl4j-integration-tests,xrpl4j-coverage -am clean compile test-compile -DskipTests -Dmaven.javadoc.skip=true` succeeds
- [ ] CI run confirms `unittests-gate` and `aggregate-visibility` status checks appear on this PR as expected


pom.xml
```diff
@@ -298,6 +298,8 @@
     <module>xrpl4j-integration-tests</module>
     <module>xrpl4j-client</module>
     <module>xrpl4j-bom</module>
+    <!-- Must remain last: aggregates coverage after every other module has produced its exec files. -->
+    <module>xrpl4j-coverage</module>
```
cybele-ripple, line 302: small nit: would this mean that the module will log "skipping" for all maven executions? might be cleaner to include:
```
<executions>
  <!-- Disable inherited executions - this module has no sources/tests of its own -->
  <execution>
    <id>default-prepare-agent</id>
    <phase>none</phase>
  </execution>
  <execution>
    <id>report</id>
    <phase>none</phase>
  </execution>
  <execution>
    <id>pre-integration-test</id>
    <phase>none</phase>
  </execution>
  <execution>
    <id>post-integration-test</id>
    <phase>none</phase>
  </execution>

  <execution>
    <id>aggregate-coverage</id>
    <phase>verify</phase>
    <goals>
      >     </goals>
  </execution>
</executions>
```
to cut down on extraneous logging.

cybele-ripple (dismissed):

pom.xml
sappenin: Good idea. Fixed via 29eaca420000543f0c75cd83746491423d57560c

Issue #731, opened 2026-04-17
No HTTPS enforcement in JsonRpcClient construction


ckeshava:
## Description

The `HttpUrl` is passed directly to Feign without validating the scheme is HTTPS. A developer could accidentally configure a plaintext HTTP connection, exposing transaction data and potentially private keys to man-in-the-middle attacks.

## Location

[JsonRpcClient.java:104](https://github.com/XRPLF/xrpl4j/blob/e48a6a6ae2457b0e330390610effe8e881894c4b/xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/JsonRpcClient.java#L104)

## Impact / Severity

**MEDIUM** — MITM vulnerability if HTTP is used. For a financial library, at minimum a warning should be logged for non-HTTPS connections.

## Adversarial Review Notes

Confirmed. No scheme validation exists in either `construct()` overload.

## Unit Test Patch

No direct unit test — requires constructing an HttpUrl which is in the xrpl4j-client module.

Issue #607, opened 2025-03-07
Consider retrofit as another HttpClient for xrpl4j


sappenin:
The HTTP client in xrpl4j is provisional (and marked `@Beta` to reflect that). As we explore a canonical client for xrpl4, we should consider [retrofit](https://square.github.io/retrofit/) as an alternative to feign.

See https://github.com/XRPLF/xrpl4j/issues/603 for more insight (it's possible that issue would be resolved with a simple version bump of Feign).

Issue #603, opened 2025-02-24
Failed to Initialize XrplClient due to java.lang.reflect.InvocationTargetException on Android 12 and Below


sallewsa:
Hi, I am using Xrpl4j version 4.0.4 with JDK 11. In my code, I initialize XrplClient using `client = new XrplClient(url.toHttpUrl())`. However, the initialization fails with the following error on Android 12 and below.

```java
java.lang.IllegalStateException: java.lang.reflect.InvocationTargetException
    at feign.DefaultMethodHandler.<init>(DefaultMethodHandler.java:46)
    at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:147)
    at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:56)
    at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:48)
    at feign.Feign$Builder.target(Feign.java:196)
    at feign.Feign$Builder.target(Feign.java:192)
    at org.xrpl.xrpl4j.client.JsonRpcClient.construct(JsonRpcClient.java:104)
    at org.xrpl.xrpl4j.client.JsonRpcClient.construct(JsonRpcClient.java:82)
    at org.xrpl.xrpl4j.client.XrplClient.<init>(XrplClient.java:131)
Caused by: java.lang.reflect.InvocationTargetException
    at java.lang.reflect.Method.invoke(Native Method)
    at feign.DefaultMethodHandler.androidLookup(DefaultMethodHandler.java:76)
    at feign.DefaultMethodHandler.readLookup(DefaultMethodHandler.java:57)
    at feign.DefaultMethodHandler.<init>(DefaultMethodHandler.java:42)
Caused by: java.lang.NoSuchMethodException: java.lang.invoke.MethodHandles$Lookup.<init> [class java.lang.Class]
    at java.lang.Class.getConstructor0(Class.java:2332)
    at java.lang.Class.getDeclaredConstructor(Class.java:2170)
```

sappenin:
Hi @sallewsa -- My knowledge of Android is pretty limited these days, but hopefully between the two of use we might figure out how to fix this one. Some thoughts:

* xrpl4j requires Java 1.8 or above, and I see you're using Java11. However, is there something you need to do in Android to foce Java8 compatibility (e.g., does [this comment](https://github.com/square/retrofit/issues/3281#issuecomment-574063984) inspire anything)?
[...]

Issue #334, opened 2022-12-12
Consider an empty target for Feign client used for XrplClient


theotherian:
Currently, `XrplClient` requires specifying the url of the XRPL cluster to communicate with. This is a requirement of `JsonRpcClient` which is based on Feign.

The construction of the Feign client can be done without a URL by using this for `target`:
```
.target(Target.EmptyTarget.create(JsonRpcClient.class, "JsonRpcClient"));
```

This would be preferred in cases where an application needs to talk to multiple ledger instances.

PR #837, opened 2026-09-17
feat: allow a custom Feign Client to be supplied to XrplClient

nsmithau:
## Summary

- `XrplClient` always sends requests through Feign's default `Client`, which is backed by `java.net.HttpURLConnection`. There is no way to configure connection pooling, per-host concurrency limits, TLS, proxies, interceptors or metrics; the only knobs are connect and read timeouts.
- Adds `JsonRpcClient.construct(HttpUrl, feign.Client, Options)` and a matching public `XrplClient(HttpUrl, Client, Options)` constructor so callers can supply any Feign `Client`, for example `feign-okhttp` with an OkHttp `ConnectionPool`.
- The existing `construct(HttpUrl, Options)` now delegates to the new overload with `new Client.Default(null, null)`, which is exactly what `Feign.builder()` instantiates when no client is set, so current callers see no behaviour change. No existing signatures are modified and no dependencies are added.
- Documents the new constructor in the README under Usage, including that consumers must add a Feign client module themselves at the same Feign version `xrpl4j-client` depends on.

## Related issues

- Partially enables #269 (client-side load balancing) and #334 (one client talking to multiple ledgers): a caller-supplied `Client` can route requests to more than one host without waiting on DNS TTLs.
- #607 notes the HTTP layer is provisional and floats Retrofit as an alternative to Feign. This change is additive on a class already marked `@Beta`, so it can be dropped or replaced if the transport changes, and in the meantime it gives users a supported alternative to the workaround in #603 of writing a separate client.
- The new overload inherits the scheme behaviour tracked in #731 unchanged; validation, if added, belongs in the shared path both overloads now go through.

## Test plan

- [x] `mvn -pl xrpl4j-client test` with checkstyle enabled: 58 tests, 0 failures, 0 checkstyle violations
- [x] New `JsonRpcClientTest#testConstructWithCustomFeignClientRoutesRequestsThroughIt` mocks a `feign.Client`, constructs through the new overload, and verifies the request is executed by the mock and its response is returned
- [x] New `JsonRpcClientTest#testConstructWithNullFeignClientThrows` covers the null guard
- [x] Javadoc generation for `xrpl4j-client` succeeds
- [x] Verified end-to-end outside the repo against `s1.ripple.com` using `feign-okhttp` with a shared `ConnectionPool`: three sequential `server_info` calls reused a single pooled connection


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -162,6 +163,22 @@ public XrplClient(
...
+   * depends on.
+   *
+   * @param rippledUrl The {@link HttpUrl} of the node to connect to.
+   * @param client     The Feign {@link Client} used to execute HTTP requests.
+   * @param options    Feign request {@link Options} (connect timeout, read timeout, follow redirects).
+   */
+  public XrplClient(HttpUrl rippledUrl, Client client, Options options) {
+    this(JsonRpcClient.construct(rippledUrl, client, options));
+  }
+
```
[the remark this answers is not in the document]
nsmithau: Addressed in 90fc8e56. `XrplClientTest#constructorWithCustomFeignClientRoutesRequestsThroughIt` now constructs `XrplClient` through the new `(HttpUrl, Client, Options)` constructor with a mocked Feign `Client`, calls `channelVerify`, asserts on the deserialised `signature_verified` value from the mocked response, and verifies the mock executed the request.

I confirmed the test catches the regression you describe: with the constructor temporarily changed to call `JsonRpcClient.construct(rippledUrl, options)` and discard the client, the test fails with a connection-refused error against `localhost:1`.

On the docstring-coverage warning: the two functions it counts as undocumented are test methods. Test methods in this repository do not carry Javadoc, so I have left them consistent with the existing suite.

README.md
```diff
@@ -60,6 +60,25 @@ found in the BOM to your `pom.xml`. For example:
 </dependencies>
 ```
 
+### Using a pooled HTTP client
```
sappenin: Can you move this into the README for the xrpl-client? See `/xrplfj-client/README.md`. If anything, in this README, it should probably just have a small pointer to go read about how to use the client inside of the client module.

README.md
```diff
@@ -60,6 +60,25 @@ found in the BOM to your `pom.xml`. For example:
...
+
+By default `XrplClient` sends requests with Feign's built-in `HttpURLConnection` client, which
+has no connection-pool configuration. To use a pooled client, pass any Feign `Client` to the
+`XrplClient(HttpUrl, Client, Options)` constructor. For example, add `feign-okhttp` at the same
+Feign version that `xrpl4j-client` uses, then:
+
+```java
+OkHttpClient okHttpClient = new OkHttpClient.Builder()
```
sappenin: We should fully-qualify some of these so there's no ambiguity during a copy/paste (e.g., `okhttp3.OkHttpClient` vs. `feign.okhttp.OkHttpClient`).

Probably replace this whole segment with something like this:

```
okhttp3.OkHttpClient okHttpClient = new okhttp3.OkHttpClient.Builder()
  .connectionPool(new okhttp3.ConnectionPool(32, 30, java.util.concurrent.TimeUnit.SECONDS))
  .build();

XrplClient xrplClient = new XrplClient(
  okhttp3.HttpUrl.get("https://s1.ripple.com:51234/"),
  new feign.okhttp.OkHttpClient(okHttpClient),
  new feign.Request.Options(
    5, java.util.concurrent.TimeUnit.SECONDS,
    30, java.util.concurrent.TimeUnit.SECONDS,
    true
  )
);
```

sappenin:
@nsmithau thanks for submitting this PR. I had a few minor nits and reorganizations, so instead of making comments here, I just created a new PR that started from your changes. That's in #838 (should have everything you proposed here). I'll close this one in-favor of #838.

Issue #269, opened 2022-06-02
Enable client-side load-balancing in XrplClient.java


sappenin:
The idea here is that the current s1.ripple.com setup uses a DNS TTL of 60s, so any networking stack will peg the XRPL4j client to a given IP address for at least that long (or whatever the DNS TTL happens to be). If the node that the client stack is pegged to (i.e., the IP Address) is down, or not accesible, then the xrpl4j client will lose access.

One alternative here, especially in private-ledger scenarios (e.g.) where the xrpl4j client know which IP addresses to be talking to when making rippled calls, would be to allow the client to be constructed with one or more URLs and/or IPs so that if the client is unable to communicate with the rippled host, it can attempt to use a different host without waiting for DNS timeouts that the client cannot control.

PR #838, opened 2026-09-17
feat: allow a custom Feign Client to be supplied to XrplClient

sappenin:
## Summary

Builds on #837 (feat: allow a custom Feign Client to be supplied to XrplClient), with review feedback incorporated:

- Adds `JsonRpcClient.construct(HttpUrl, Client, Options)` and a matching public `XrplClient(HttpUrl, Client, Options)` constructor so callers can supply any Feign `Client`, for example `feign-okhttp` with an OkHttp `ConnectionPool`. (from #837)
- Deprecates `JsonRpcClient.construct(HttpUrl, Options)` and `XrplClient(HttpUrl, Duration, Duration)` in favor of the `Client`/`Options`-accepting overloads, which expose the full set of Feign configuration rather than just timeouts.
- Makes `XrplClient(JsonRpcClient)` public (was package-private, `@VisibleForTesting`-only). This is the most general extension point on the class — unlike the other constructors, it isn't tied to Feign at all, so a caller can supply a `JsonRpcClient` built with a custom `Client`, or an entirely different `JsonRpcClient` implementation not backed by Feign (relevant given #607 floats Retrofit as an alternative HTTP layer).
- Documents that supplying a custom `Client` hands TLS, proxy, and redirect handling over to that client's own configuration rather than the JVM defaults used by `Client.Default`.
- Moves the "Using a pooled HTTP client" README section into `xrpl4j-client/README.md`, where `XrplClient` is actually introduced (it previously landed before `XrplClient` is ever mentioned in the root README), and fixes the code sample's `OkHttpClient`/`ConnectionPool` naming collision between `okhttp3` and `feign.okhttp`.
- Strengthens the new unit tests to assert the supplied `Options` instance actually reaches `Client.execute(...)`, not just that the mock is invoked, and adds a regression test guarding `XrplClient(JsonRpcClient)`'s public visibility.

## Related issues

- Partially enables #269 (client-side load balancing) and #334 (one client talking to multiple ledgers).
- #607 notes the HTTP layer is provisional and floats Retrofit as an alternative to Feign; making `XrplClient(JsonRpcClient)` public means consumers aren't limited to Feign-specific extension points if that ever changes.
- Inherits the scheme behaviour tracked in #731 unchanged.

## Test plan

- [x] `mvn -pl xrpl4j-client -am test`: 60 tests, 0 failures, 0 errors
- [x] `mvn -pl xrpl4j-client -am checkstyle:check`: 0 violations
- [x] `JsonRpcClientTest#testConstructWithCustomFeignClientRoutesRequestsThroughIt` / `#testConstructWithNullFeignClientThrows` cover the new overload and its guard clause
- [x] `XrplClientTest#constructorWithCustomFeignClientRoutesRequestsThroughIt` covers an end-to-end `channelVerify` call routed through a mocked `feign.Client`, and asserts the supplied `Options` instance is the one passed to `Client.execute(...)`
- [x] `XrplClientTest#jsonRpcClientConstructorIsPublic` guards against `XrplClient(JsonRpcClient)` being narrowed back to package-private


xrpl4j-client/src/main/java/org/xrpl/xrpl4j/client/XrplClient.java
```diff
@@ -147,7 +146,13 @@ public XrplClient(final HttpUrl rippledUrl) {
    * @param rippledUrl     The {@link HttpUrl} of the node to connect to.
    * @param connectTimeout A {@link Duration} indicating the client's connect timeout.
    * @param readTimeout    A {@link Duration} indicating the client's read timeout.
+   *
+   * @deprecated Prefer {@link #XrplClient(HttpUrl, Client, Options)}, which exposes the full set of Feign
+   *   {@link Options} rather than just connect/read timeouts, and also allows a custom {@link Client} to be supplied.
+   *   For an {@link Options}-only equivalent of this constructor, use
+   *   {@code new XrplClient(JsonRpcClient.construct(rippledUrl, options))}.
```
cybele-ripple: Looks like `JsonRpcClient.construct(rippledUrl, options)` is itself deprecated by this same PR, so this points users from one deprecated API straight to another.
Maybe point to the following instead: 
```
new XrplClient(rippledUrl, new Client.Default(null, null), options)
```

sappenin: I decided to just remove that line entirely because`Client.Default(null,null)` is how Feign 13.8 does it, but it may not be the default forever. Plus, the first suggestion is preferable in any event.

See 534b345318c51308dd823ffef93e05531c5ce9b2 for the fix.

cybele-ripple (approved):

